PR #32, opened 2020-08-27
CI: add code-coverage and make the 'lint' stage faster

youngjoon-lee:
See http://ci.medibloc.org/job/panacea-core/job/PR-32/. You can find the `Code Coverage` menu on the left side-bar.

Also, in `Dockerfile`, download the pre-built `golangci-lint`, instead of `go get .../golangci-lint` everytime.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@
 .idea/
 *.iml
 build/
+
+# artifacts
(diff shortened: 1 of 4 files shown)


cl9200 (approved):
Good job

PR #34, opened 2020-09-01
Add signature verification for CreateDID

youngjoon-lee:
Based on #29 

Previously `MsgCreateDID` didn't have any signature and verification logic. It can cause that someone who doesn't know the DID private key can create a DID document.
To solve it, `MsgCreateDID` should have the signature, similar as `MsgUpdateDID`.


cl9200 (approved):
리뷰 완료하였습니다

PR #35, opened 2020-09-01
Deactivate DID instead of deleting it

youngjoon-lee:
Based on #34 

According to the [DID spec](https://www.w3.org/TR/did-core/#did-syntax):
```
A DID is expected to be persistent and immutable. That is, a DID is bound exclusively and permanently to its one and only subject. Even after a DID is deactivated, it is intended that it never be repurposed.
```
DID shouldn't be deleted from Keeper. It should be just deactivated, so that anyone cannot recreate it.

This topic also has been discussed in [#29](https://github.com/medibloc/panacea-core/pull/29#issuecomment-684121150).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,18 +28,6 @@ jobs:
     - name: Test
       run: make test
 
-    - name: Convert coverage.out into coverage.lcov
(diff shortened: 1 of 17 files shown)


cl9200 (approved):
리뷰 완료하였습니다

PR #37, opened 2020-09-02
Remove an unnecessary constant for DID keystore

youngjoon-lee:
In `x/did/client/keystore/`, I removed a constant `derivedKeyLen`, because it duplicates `scryptDKLen` which has the same meaning.
Also, when loading the key from JSON, it's better to use the `dklen` in the JSON, instead of using the constant.


cl9200 (approved):

youngjoon-lee:
Closing this PR because it would be enhanced by https://github.com/medibloc/panacea-core/issues/38

PR #54, opened 2020-09-22
Merge tag 'v1.2.5-internal'

youngjoon-lee:
In prior to upgrade Cosmos SDK to the latest, I would like to sync the `master` with the `v1.2.5-internal` which is deployed for mainnet and testnet.

It seems all commits in the `v1.2.5-internal` are for updating `go.mod`. (tendermint and medibloc/cosmos-sdk)

image


youngjoon-lee:
It seems the `cosmos/cosmos-sdk` which is used by `medibloc/cosmos-sdk` has a bug.
When running `panacead ...`, I got an error when `--trust-node` is false, but I couldn't specify it due to: 
```
(⎈ |N/A:N/A)~/r/panacea-core ❯❯❯ panacead gentx --name validator                                                                                             ✘ 1 tags/v1.2.5-internal ✱ ◼
Must specify these options: --node  when --trust-node is false
(⎈ |N/A:N/A)~/r/panacea-core ❯❯❯ panacead gentx --name validator --trust-node                                                                                ✘ 1 tags/v1.2.5-internal ✱ ◼
ERROR: unknown flag: --trust-node
```

cl9200:
> It seems the `cosmos/cosmos-sdk` which is used by `medibloc/cosmos-sdk` has a bug.
> When running `panacead ...`, I got an error when `--trust-node` is false, but I couldn't specify it due to:
> 
> ```
> (⎈ |N/A:N/A)~/r/panacea-core ❯❯❯ panacead gentx --name validator                                                                                             ✘ 1 tags/v1.2.5-internal ✱ ◼
> Must specify these options: --node  when --trust-node is false
> (⎈ |N/A:N/A)~/r/panacea-core ❯❯❯ panacead gentx --name validator --trust-node                                                                                ✘ 1 tags/v1.2.5-internal ✱ ◼
> ERROR: unknown flag: --trust-node
> ```
> 
> I'm also considering upgrading `cosmos/cosmos-sdk` directly, without merging this PR.

Maybe it's an error caused by viper's IsSet command whose spec is changed.
https://github.com/spf13/viper/releases/tag/v1.6.0

There was a similar error before.
https://github.com/medibloc/cosmos-sdk/commit/a464176648a06f7346f56456f1256f8a63967667

youngjoon-lee:
@cl9200 
Yeah. 
It's also related to https://github.com/cosmos/cosmos-sdk/pull/6021 fixed in `cosmos/cosmos-sdk` [v0.37.11](https://github.com/cosmos/cosmos-sdk/blob/master/CHANGELOG.md#v03711---2020-04-22).

So, let's upgrade the `cosmos/cosmos-sdk`.

The `gaiad` is using `cosmos/cosmos-sdk` [v0.37.13](https://github.com/cosmos/gaia/blob/main/go.mod#L7). How about following that version?

cl9200:
@youngjoon-lee 
I think v0.37.13 would be good.

cl9200 (approved):
:+1:

PR #67, opened 2020-10-16
Use 'documents' as the DID genesis JSON key, instead of 'Documents'

youngjoon-lee:
Previously, the `genesis.json` looked like:
```json
{
  "app_state": {
    "did": {
      "Documents": {
      }
  }
}
```

Let's use `documents` instead of `Documents`.
```json
```json
{
  "app_state": {
    "did": {
      "documents": {
          "did:panacea:9iofqBjmEkL4XRiCguNxun3GVKn2Ghoqud39uS25S1aY": {
          }
      }
  }
}
```

### DIFF
--- a/x/did/types/genesis.go
+++ b/x/did/types/genesis.go
@@ -1,7 +1,7 @@
 package types
 
 type GenesisState struct {
-	Documents map[string]DIDDocumentWithSeq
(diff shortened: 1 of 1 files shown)


cl9200 (approved):
:+1:

PR #73, opened 2020-10-24
Fix: Make LCD use the latest height or the height specified in the URL query

### LINKED ISSUE #72: LCD keeps looking at the old height
Right after starting the LCD, it looks at the latest block height.
But, after serving for the 1st request, it keeps looking at the old height which was used for the 1st request.

From my debugging, I just found that the `cliCtx.Height` isn't changed since the 1st request was served.
https://github.com/medibloc/panacea-core/blob/dfbaa62a90b931b8001a6fee2ae8ee2d7b1241d0/x/did/client/rest/query.go#L42

youngjoon-lee:
Close #72 

This version was already deployed on the Testnet for Medipass (only `panaceacli`, not `panacead`).
After merging this PR, I'm gonna release v1.3.2 and deploy it.

### DIFF
--- a/x/aol/client/rest/query.go
+++ b/x/aol/client/rest/query.go
@@ -59,6 +59,11 @@ func listTopicHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
 			return
 		}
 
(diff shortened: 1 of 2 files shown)


cl9200 (approved):
:+1:

PR #75, opened 2020-10-30
Fix parsing keeper keys of `x/aol` and `x/did`

### LINKED ISSUE #70: Fix parsing keeper keys of `x/aol` and `x/did`
~~The actual data (such as panacea accounts) can contain `0x00`s.~~
Currently, when parsing keeper keys, the `bytes.Split(..., seperator)` is being used.
The `separtor` is vary, such as `ownerAddress`, `topic`, and so on.
The `ownerAddress` or `topic` can contain any binary characters. That is, they can make parsing wrong. 
It happened when upgrading the testnet from v1.2.5-internal to v1.3.1.

Fortunately, it's simple to fix it.
We parse keeper keys, only when we export a genesis.
Also, whenever we parse keeper keys, we know the key prefix and we always extract the rest of the key (the last element).
So, let's truncate the prefix part from the key. That's what we wanna get.

This fix must be applied to both `v1.2.6-internal` and `v1.3.1`, so that the genesis can be exported correctly when upgrading the mainnet.

Consider using the data size instead of the delimiter, so that there is no potential risk.

youngjoon-lee:
Close #70

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,14 @@
 
 ## Unreleased
 
+### Features
+
 TBD
+
+### Bug fixes
+
+- [\#73](https://github.com/medibloc/panacea-core/pull/73) Make LCD use the latest height or the height specified in the URL query
+- [\#74](https://github.com/medibloc/panacea-core/pull/74) Fix parsing keeper keys of `x/aol` and `x/did`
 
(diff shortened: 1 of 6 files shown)


cl9200 (approved):
:+1:

PR #80, opened 2020-11-07
Add the `x/token` module for issuing new tokens

### LINKED ISSUE #79: A feature for issuing new tokens
This would be also useful for the BMGF project.

The `BaseAccount` in the `cosmos-sdk` already can hold `[]Coin` (one per currency).
So, I think we can create a new module `x/currency` (TBD) which can create new currencies (new denominators). 
I guess we can reuse the existing `tx send` feature for new currencies.

Note that new currencies will not be used for staking. Only MED is a staking coin. @cl9200

youngjoon-lee:
Close #79 

Code + Tests + Docs are all done. Such a big PR, but I think you're familiar with the `x/*` directory structure :)
I recommend to review the Gitbook changes first, so that you can get how to use this module.
And then, the core file is `types/token.go`.


client/lcd/swagger-ui/swagger.yaml
```diff
@@ -1993,6 +1993,72 @@ paths:
...
+    get:
+      summary: Query all tokens
+      description: Query all tokens
+      produces:
+        - application/json
+      tags:
+        - Token
+      responses:
+        200:
+          description: "A DID document in the format: https://www.w3.org/TR/did-core"
```
cl9200, line 2030: description from DID API  :)

cl9200 (approved):
:+1:

client/lcd/swagger-ui/swagger.yaml
youngjoon-lee: Danke :) 23649560e68633cb54e18a830a7f5bd167d5609f

PR #82, opened 2020-12-02
For 1.2.7-internal: Fix parsing keeper keys of `x/aol`

youngjoon-lee:
Similar as the PR #75 , but based on the `v1.2.6-internal`, because this fix is essential for upgrading from `v1.2.x` to `v1.3.x`.

This fix would be released as `v1.2.7-internal`.

### DIFF
--- a/x/aol/keeper.go
+++ b/x/aol/keeper.go
@@ -1,8 +1,6 @@
 package aol
 
 import (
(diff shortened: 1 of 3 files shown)


cl9200 (approved):
:+1:

PR #83, opened 2020-12-02
For v1.3.3: Use cosmos/cosmos-sdk v0.37.14

youngjoon-lee:
This would be released as `v1.3.3`.

Unlike `v1.3.3-internal`, this uses the `cosmos/cosmos-sdk` (instead of `medibloc/cosmos-sdk`).

For more details, https://github.com/medibloc/cosmos-sdk/issues/9


cl9200 (approved):
:+1:

PR #85, opened 2020-12-02
For 1.2.7: Fix parsing keeper keys of `x/aol`

youngjoon-lee:
Similar as the PR #82  , but based on the `v1.2.6` instead of `v1.2.6-internal`.
This fix would be released as `v1.2.7`.


cl9200 (approved):
:+1:

PR #86, opened 2020-12-02
Fix Lint CI

youngjoon-lee:
Previously, The `golangci-lint` github action has a problem. So, I've run it as a separate github action job. 
But, it has been failed from today for some reason.

I just bumped its version from v1 to v2, and moved it to the origin github action job. Now it works.


cl9200 (approved):
:+1:

PR #100, opened 2021-04-27
x/did: Fix checking if VerificationRelationships are valid

youngjoon-lee:
Previously, the role of `VerificationMethodByID(id)` was:
- Find a `authentication` from the DID Document
- Find a corresponding `verificationMethod` which has the same ID as the `authentication`.

It was used for the `DIDDocument.Valid()`. But, if the DID Document has `assertionMethod`s, not `authentication`s, the `VerificationMethodByID(id)` returns 'not found'.
```json
{
  "verificationMethod": [{
    "id": "key1",
    "publicKeyBase58": "asdgkadjspoiejflkadskjfa",
  }],
  "assertionMethod": [
    "key1"
  ]
}
```

So, the function should be split for various use cases.
- `VerificationMethodByID(id)`
  - Find a `verificationMethod` by its ID from the DID Document
- `VerificationMethodFrom(verificationRelationships, ID)`
  - Find a `verificationRelationship` from the DID Document
  - Find a corresponding `verificationMethod` which has the same ID as the `verificationRelationship`.

The 1st one is used in the `DIDDocument.Valid()`.
The 2nd one is used in the `verifyDIDOwnership()` which traverses the `authentication` list.


cl9200 (approved):
LGTM

PR #109, opened 2021-05-14
docs: Add 'panacea-opentestnet' URL to the Incentivized Testnet guide

youngjoon-lee:
will be announced to participants of the Incentivized Testnet.


ssoy85 (approved):
👍

PR #113, opened 2021-05-17
starport: Change Makefile to build only `panacead`

youngjoon-lee:
From Stargate, we will have only one binary: `panacead`.

CI should work from this PR.


cl9200 (approved):

gyuguen (approved):
👍

PR #115, opened 2021-05-18
Add burn coin modules

gyuguen:
A module has been added that deletes the coins when sent to the specified address.
- burn address : `panacea100000000000000000000000000000000nqmafp`

### DIFF
--- a/x/burn/internal/keeper/burn.go
+++ b/x/burn/internal/keeper/burn.go
@@ -18,7 +18,6 @@ func (k Keeper) BurnCoins(ctx sdk.Context, acc string) sdk.Error {
 	}
 
 	amt := bunAccount.GetCoins()
-
 	if amt.Empty() {
 		return nil
 	}
--- a/x/burn/internal/keeper/burn.go
+++ b/x/burn/internal/keeper/burn.go
@@ -7,10 +7,10 @@ import (
 	accountExported "github.com/cosmos/cosmos-sdk/x/auth/exported"
 )
 
-func (k Keeper) BurnCoins(ctx sdk.Context, acc string) sdk.Error {
+func (k Keeper) BurnCoins(ctx sdk.Context, acc string) error {
 	burnAccount, err := getAccount(ctx, k, acc)
 	if err != nil {
-		return sdk.ErrInvalidAddress(err.Error())
+		return err
 	}
 
 	if burnAccount == nil {
@@ -26,12 +26,12 @@ func (k Keeper) BurnCoins(ctx sdk.Context, acc string) sdk.Error {
 
 	_, err = k.bankKeeper.SubtractCoins(ctx, burnAccount.GetAddress(), amt)
 	if err != nil {
-		return sdk.ErrInvalidCoins(err.Error())
+		return err
 	}
 
 	ctx.Logger().Info("Success burn coin to burnAccount.", fmt.Sprintf("address: %s, coins: %s", acc, amt))
 
-	burnCoinsToSupply(ctx, k, amt)
+	burnCoinsFromSupply(ctx, k, amt)
 
 	return nil
 }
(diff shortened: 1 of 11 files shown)


x/burn/internal/keeper/burn.go
```diff
@@ -0,0 +1,57 @@
...
+		return sdk.ErrInvalidAddress(err.Error())
+	}
+
+	if bunAccount == nil {
+		return nil
+	}
+
+	amt := bunAccount.GetCoins()
+
+	if amt.Empty() {
```
youngjoon-lee, line 22: In Go, we didn't used to put any empty line before the subsequent `if` statement.
```suggestion
	amt := bunAccount.GetCoins()
	if amt.Empty() {
```

x/burn/internal/keeper/burn.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	accountExported "github.com/cosmos/cosmos-sdk/x/auth/exported"
+)
+
+func (k Keeper) BurnCoins(ctx sdk.Context, acc string) sdk.Error {
+	bunAccount, err := getAccount(ctx, k, acc)
```
youngjoon-lee, line 11: `bunAccount` looks like a typo :)

x/burn/internal/keeper/burn.go
```diff
@@ -0,0 +1,56 @@
...
+	burnAcc, err := sdk.AccAddressFromBech32(acc)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return k.accountKeeper.GetAccount(ctx, burnAcc), nil
+}
+
+func burnCoinsToSupply(ctx sdk.Context, k Keeper, amt sdk.Coins) {
```
youngjoon-lee, line 50: `...FromSupply` looks more clear :)
https://github.com/cosmos/cosmos-sdk/blob/fd60408b493365ad840f9366b1dd1a5d25909ffe/x/supply/internal/types/supply.go#L47

x/burn/internal/keeper/burn.go
```diff
@@ -0,0 +1,56 @@
...
+	if amt.Empty() {
+		return nil
+	}
+
+	ctx.Logger().Info("find burn coin.", fmt.Sprintf("address: %s, coins: %s", bunAccount.GetAddress(), bunAccount.GetCoins()))
+
+	_, err = k.bankKeeper.SubtractCoins(ctx, bunAccount.GetAddress(), amt)
+
+	if err != nil {
+		return sdk.ErrInvalidCoins(err.Error())
```
youngjoon-lee, line 29: Because `bankKeeper.SubstractCoins()` already returns `sdk.Error`, I think we can just return it as it is.
```suggestion
		return err
```

x/burn/module.go
```diff
@@ -0,0 +1,118 @@
...
+
+	"github.com/medibloc/panacea-core/x/burn/internal/types"
+)
+
+type AppModuleBasic struct{}
+
+type AppModule struct {
+	AppModuleBasic
+	keeper       keeper.Keeper
+	supplyKeeper supply.Keeper
```
youngjoon-lee, line 27: It seems this `supplyKeeper` is not needed, because the `burn.Keeper` already contains a `supplyKeeper`.

x/burn/module.go
```diff
@@ -0,0 +1,118 @@
...
+	return nil
+}
+
+func (AppModule) BeginBlock(_ sdk.Context, _ abci.RequestBeginBlock) {}
+
+func (am AppModule) EndBlock(ctx sdk.Context, block abci.RequestEndBlock) []abci.ValidatorUpdate {
+	err := am.keeper.BurnCoins(ctx, types.BurnAddress)
+
+	if err != nil {
+		panic("Failed burn. msg : " + err.Error())
```
youngjoon-lee, line 113: I'm not sure if it's okay to panic in the `EndBlock`. Let me google it.

gyuguen: @youngjoon-lee 
I think... `panic` doesn't seem to be a good way.

gyuguen: Delete "panic" and modify to write error log.

youngjoon-lee (approved):
LGTM 👍

x/burn/module.go
gyuguen: @youngjoon-lee 
That's right. I deleted it!

x/burn/internal/keeper/burn.go
gyuguen: @youngjoon-lee 
OK. Change sdk.Error to native error

x/burn/internal/keeper/burn.go
gyuguen: OK. I changed method name!
Thanks.

PR #123, opened 2021-05-27
feat: stargate: Add more cosmos/base proto dependencies

youngjoon-lee:
So far, we've used the only `cosmos/base/query/v1beta1/pagniation.proto`. But, we need to use other `cosmos/base/**/*.proto` files, such as `cosmos/base/v1beta1/coin.proto`.

Thus, let's fetch all of `cosmos/base/**/*.proto` files as dependencies. Later, we can minimize them.


gyuguen (approved):
👍

PR #124, opened 2021-05-28
feat: stargate: Add `token` module

youngjoon-lee:
Please don't review scaffoldings. Please review the only following commits:
- 4dc691c: Modify proto files
- d27d881: Implement logics
- 9392a8c: Remove swagger-ui-gen from dependencies of `make install`

This PR doesn't have any unit tests yet, so that we can share the basic strategy how to modify scaffolded codes.
I tested it via `panacead tx` and `panacead query` manually. I will commit unit tests at the next PR soon.

TODO: Consider the genesis compatibility between the previous version and Stargate. In the previous version, the genesis contains `map[string]Token`, but it contains `[]*Token` in the Stargate. I'll resolve it at the next PR.


gyuguen (approved):
👍

PR #126, opened 2021-05-28
feat: stargate: Add the default genesis overriding

youngjoon-lee:
When running `panacead init ...`, the Cosmos generates a genesis with default params that are defined in each `x/*` modules (`stake` coin denom, and so on...).
So, it's annoying to modify the generated `genesis.json` every time after running `panacead init ...`.

Thus, let's override it.

Actually, we've had this overriding logic so far: https://github.com/medibloc/panacea-core/blob/985271f795c2ff0bf1f6b63152d97fabb41dbfa2/init/init.go

In this PR, I made it work in Stargate by referring to https://github.com/public-awesome/stargaze/blob/b92bf9847559b9c7f4ac08576d056d3d00efe12c/cmd/starsd/cmd/init.go.

### DIFF
--- a/cmd/panacead/cmd/init.go
+++ b/cmd/panacead/cmd/init.go
@@ -76,6 +76,11 @@ func InitCmd(mbm module.BasicManager, defaultNodeHome string) *cobra.Command {
 	return initCmd
 }
 
+const (
+	blockTimeSec = 1
+	unbondingPeriod = 60 * 60 * 24 * 7 * 3 * time.Second // three weeks
+)
+
 // overrideGenesis overrides some parameters in the genesis doc to the panacea-specific values.
 func overrideGenesis(cdc codec.JSONMarshaler, genDoc *types.GenesisDoc) (json.RawMessage, error) {
 	appState := make(map[string]json.RawMessage)
@@ -87,7 +92,7 @@ func overrideGenesis(cdc codec.JSONMarshaler, genDoc *types.GenesisDoc) (json.Ra
 	if err := cdc.UnmarshalJSON(appState[stakingtypes.ModuleName], &stakingGenState); err != nil {
 		return nil, err
 	}
-	stakingGenState.Params.UnbondingTime = time.Second * 60 * 60 * 24 * 7 * 3 // three weeks
+	stakingGenState.Params.UnbondingTime = unbondingPeriod
 	stakingGenState.Params.MaxValidators = 50
 	stakingGenState.Params.BondDenom = assets.MicroMedDenom
 	appState[stakingtypes.ModuleName] = cdc.MustMarshalJSON(&stakingGenState)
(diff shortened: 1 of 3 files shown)


cmd/panacead/cmd/init.go
```diff
@@ -0,0 +1,142 @@
...
+
+	var slashingGenState slashingtypes.GenesisState
+	if err := cdc.UnmarshalJSON(appState[slashingtypes.ModuleName], &slashingGenState); err != nil {
+		return nil, err
+	}
+	slashingGenState.Params.SignedBlocksWindow = 10000
+	slashingGenState.Params.MinSignedPerWindow = sdk.NewDecWithPrec(5, 2)
+	slashingGenState.Params.SlashFractionDoubleSign = sdk.NewDecWithPrec(5, 2) // 5%
+	slashingGenState.Params.SlashFractionDowntime = sdk.NewDecWithPrec(1, 4)   // 0.01%
+	appState[slashingtypes.ModuleName] = cdc.MustMarshalJSON(&slashingGenState)
```
gyuguen, line 139: @youngjoon-lee 
I think the previous setting was a three-week period.
Is it the same setting?
https://github.com/medibloc/panacea-core/blob/985271f795c2ff0bf1f6b63152d97fabb41dbfa2/init/init.go#L125

gyuguen (approved):
👍

cmd/panacead/cmd/init.go
youngjoon-lee: Oh, good point! I guess the `MaxEvidenceAge` parameter has moved from `x/slashing` to a new module `x/evidence`: https://github.com/cosmos/cosmos-sdk/blob/master/CHANGELOG.md#state-machine-breaking-3.
image
I forgot to manage it. Let me fix it.

youngjoon-lee: e18e8e6
Please review this additional commit.

As a background,
`MaxEvidenceAge` moved from `x/slahsing` to `x/evidence` at [Cosmos SDK v0.38.0](https://github.com/cosmos/cosmos-sdk/blob/master/CHANGELOG.md#state-machine-breaking-3) as mentioned above. Then, it moved to Tendmermint consensus params at [Cosmos SDK v0.40.0](https://github.com/cosmos/cosmos-sdk/blob/master/CHANGELOG.md#state-machine-breaking-2) as below:
![image](https://user-images.githubusercontent.com/5462944/120158008-9a6c3900-c22e-11eb-8da9-4b97cdb417ab.png)

As a result, the `genesis.json` should look like:
```json
{
  "consensus_params": {                                                                                                                                                                                                                                     
    "evidence": {                                                                                                                       
      "max_age_num_blocks": "1814400",                                                                                                  
      "max_age_duration": "1814400000000000",
      ...                                                                                           
    },
    ...                                                                                     
  }, 
  "app_state": {
    "slashing": {
      "params": {
         "signed_blocks_window": "10000",                                                                                                
         "min_signed_per_window": "0.050000000000000000",                                                                                
         "downtime_jail_duration": "600s",                                                                                               
         "slash_fraction_double_sign": "0.050000000000000000",                                                                           
         "slash_fraction_downtime": "0.000100000000000000"
      }
    },
    ...
  }
}
```

For more details about Tendermint consensus params, see https://docs.tendermint.com/master/tendermint-core/using-tendermint.html#fields.

gyuguen (approved):
👍

PR #127, opened 2021-05-31
fix: stargate: Make the `token` genesis backward-compatible

youngjoon-lee:
Because the previous Panacea exports a `genesis.json` like below, the new Panacea should be able to read/parse it properly:
```json
{
  "app_state": {
    "token": {                                                                                                                          
      "tokens": {                                                                                                                       
        "LOV-205": {                                                                                                                    
          "mintable": false,                                                                                                            
          "name": "love",                                                                                                               
          "ownerAddress": "panacea1rym8acf8fjfkfc7aj6hv6kdew9kpscffuway0k",                                                             
          "symbol": "LOV-E88",                                                                                                          
          "totalSupply": {                                                                                                              
            "amount": "10000",                                                                                                          
            "denom": "ulove88"                                                                                                          
          }                                                                                                                             
        }                                                                                                                            
      }                                                                                                                                 
    }, 
  }
}
```
The `app.state.token.tokens` has been a Map. However, our Starport-scaffolded type is `[]*Token`.
So, I changed it to `map[string]*Token`.

TODO: unit-test

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,6 +15,7 @@ require (
 	github.com/spf13/cobra v1.1.1
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.7.1 // indirect
+	github.com/stretchr/testify v1.7.0 // indirect
 	github.com/tendermint/tendermint v0.34.9
 	github.com/tendermint/tm-db v0.6.4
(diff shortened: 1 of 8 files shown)


gyuguen (approved):
👍

PR #129, opened 2021-06-01
feat: stargate: Add `did` module

gyuguen:
This PR has a did module added.

* https://github.com/medibloc/panacea-core/commit/bfd0371057f8d99a404d529d62e09a82c6021cba : add types
* https://github.com/medibloc/panacea-core/commit/bfd0371057f8d99a404d529d62e09a82c6021cba : add keeper, util, genesis and etc 
* https://github.com/medibloc/panacea-core/commit/8e8da13b78eb365505c3df3043fd2b22d53e8361 : add client, query, server

I'll add a comment to the code.

link
- proto3 document : https://developers.google.com/protocol-buffers/docs/proto3#json
- gogoproto options : https://github.com/cockroachdb/gogoproto/blob/master/extensions.md


proto/burn/genesis.proto
```diff
@@ -0,0 +1,8 @@
+syntax = "proto3";
+package medibloc.panaceacore.burn;
+
+option go_package = "github.com/medibloc/panacea-core/x/burn/types";
+
+// GenesisState defines the burn module's genesis state.
+message GenesisState {
```
gyuguen, line 7: `burn module`'s GenesisState is required. But deleted before PR.
So I rollbacked `genesis.proto` file.

removed commit here : https://github.com/medibloc/panacea-core/pull/122/commits/5c344e4a91924261577e672d46526b7ab5b7ae9b

x/burn/expected/expected.go
```diff
@@ -0,0 +1,15 @@
+package expected
```
gyuguen, line 1: This is a PR that was applied to the previous version, so it has nothing to do with this PR.

x/did/client/cli/queryDID.go
```diff
@@ -0,0 +1,42 @@
...
+			res, err := queryClient.DIDDocumentWithSeq(context.Background(), params)
+
+			if err != nil {
+				return err
+			}
+			return clientCtx.PrintProto(res)
+		},
+	}
+
+	flags.AddQueryFlagsToCmd(cmd)
```
gyuguen, line 39: This code required to in `cmd` file.
If you delete it, failed to get values from settings.

x/did/client/cli/txDID.go
```diff
@@ -0,0 +1,331 @@
...
+
+// readBIP39ParamsFrom reads a mnemonic and a bip39 passphrase from the reader in the interactive mode.
+// It returns empty strings in the non-interactive mode, so that they can be auto-generated by crypto.GenSecp256k1PrivKey.
+func readBIP39ParamsFrom(interactive bool, reader *bufio.Reader) (string, string, error) {
+	if !interactive {
+		return "", "", nil
+	}
+
+	// mnemonic can be an empty string
+	mnemonic, err := input.GetString("Enter your BIP39 mnemonic, or hit enter to generate one:", reader)
```
gyuguen, line 178: In before version, `github.com/cosmos/cosmos-sdk/client/input` set alias. 
So it used to `"github.com/cosmos/cosmos-sdk/client`.
But this version not alias. Required import to `"github.com/cosmos/cosmos-sdk/client/input"`

x/did/client/cli/txDID.go
```diff
@@ -0,0 +1,331 @@
...
+	ks, err := didcrypto.NewKeyStore(keystoreBaseDir())
+	if err != nil {
+		return err
+	}
+	_, err = ks.Save(verificationMethodID, privKey[:], passwd)
+	return err
+}
+
+// Deprecated to https://github.com/cosmos/cosmos-sdk/pull/5904/commits/c16d93e90d6a698cead7c19b55fcede44587aa7f
+func getCheckPassword(reader *bufio.Reader) (string, error) {
```
gyuguen, line 224: This method deprecated in cosmos sdk.
So it was implemented out of necessity.
https://github.com/cosmos/cosmos-sdk/pull/5904/commits/c16d93e90d6a698cead7c19b55fcede44587aa7f

x/did/client/rest/rest.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+// RegisterRoutes registers did-related REST handlers to a router
+func RegisterRoutes(clientCtx client.Context, r *mux.Router) {
+	// this line is used by starport scaffolding # 2
+	registerQueryRoutes(clientCtx, r)
+	registerTxHandlers(clientCtx, r)
+
+	registerQueryRoutes(clientCtx, r)
+	registerTxHandlers(clientCtx, r)
```
gyuguen, line 21: It's duplicated code.
I'll delete yet.

x/did/internal/secp256k1util/secp256k1util.go
```diff
@@ -0,0 +1,40 @@
...
+	"fmt"
+
+	"github.com/btcsuite/btcutil/base58"
+	"github.com/tendermint/tendermint/crypto/secp256k1"
+)
+
+// PrivKeyFromBytes converts a byte slice into a Secp256k1 private key.
+// It returns an error when the length of the input is invalid.
+func PrivKeyFromBytes(bz []byte) (secp256k1.PrivKey, error) {
+	key := make([]byte, secp256k1.PrivKeySize)
```
gyuguen, line 13: Before code : `var key secp256k1.PubKeySecp256k1`
But secp256k1.PubKeySecp256k1 is changed to secp256k1.PubKey.
`secp256k1.PubKey` type is only byte array(`[]byte`)
So, Changed to check key length code

x/did/types/did.go
```diff
@@ -0,0 +1,436 @@
...
+	for _, ctx := range ctxs {
+		_, dup := set[ctx] // check the duplication
+		if dup || !ValidContext(ctx) {
+			return false
+		}
+		set[ctx] = struct{}{}
+	}
+	return true
+}
+
```
gyuguen, line 252: AS-IS 
- if `@context` one value then return string.
- if `@context` multi value then return string array.

TO-BE
- Only return string array.

Because Contexts type fixed to `[]string`. Not alias.
So it is difficult to implement `MarshalJSON, UnmarshalJSON` methods.

ref : https://github.com/medibloc/panacea-core/blob/2dfceffc647db15499c715e8833c5e379c04e028/x/did/types/did.go#L159

gyuguen: Come to think of it... 
The ₩@context ₩of the existing string type will be a problem. :thinking:

x/did/types/did.go
```diff
@@ -0,0 +1,436 @@
...
+func ParseDID(str string) (string, error) {
+	did := str
+	if !ValidDID(did) {
+		return "", Error(ErrInvalidDID, str)
+	}
+	return did, nil
+}
+
+func ValidDID(did string) bool {
```
youngjoon-lee, line 45: For consistency, I think the function name `ValidateDID` is better (which is verb-prefixed).
Previously, it was adjective which represent a property of `DID`. But now, I think verb is better :).

x/did/types/did.go
```diff
@@ -0,0 +1,436 @@
...
+
+func (doc DIDDocument) Empty() bool {
+	return EmptyDID(doc.ID)
+}
+
+// TODO Duplicate with protobuf
+/*func (doc DIDDocument) String() string {
+	bz, _ := json.Marshal(doc)
+	return string(bz)
+}*/
```
youngjoon-lee, line 202: 👍 Yeah, right. I think we can remove this comment. This `String()` is not logically important. So, we can use the proto-generated one.

youngjoon-lee, line 202: 👍 Yeah, right. I think we can remove this comment. This `String()` is not logically important. So, we can use the proto-generated one.

proto/did/did.proto
```diff
@@ -0,0 +1,47 @@
...
+  repeated VerificationRelationship capabilityInvocations = 8 [json_name = "capabilityInvocation"];
+  repeated VerificationRelationship capabilityDelegations = 9 [json_name = "capabilityDelegation"];
+  repeated Service services = 10 [json_name = "service"];;
+}
+
+message VerificationMethod {
+  string ID = 1;
+  string type = 2;
+  string controller = 3;
+  string pubKeyBase58 = 4;
```
youngjoon-lee, line 25: I'm not sure whether we need to put `[json_name = "publicKeyBase58]` or not.

youngjoon-lee: Or, I guess we can use the full name `publicKeyBase58` instead of `pubKeyBase58`. The previous code wasn't clean.

proto/did/did.proto
```diff
@@ -0,0 +1,47 @@
...
+message Authentication {
+  string veriMethodID = 1;
+  VerificationMethod dedicatedMethod = 2 [(gogoproto.wktpointer) = true];
+}
+
+message VerificationRelationship {
+  string verificationMethodID = 1;
+  VerificationMethod dedicatedVerificationMethod = 2;
+}
+
```
youngjoon-lee, line 37: It seems `Authentication` isn't used. Also, `Authentication` is `VerificationRelationship` as well.

x/did/genesis.go
```diff
@@ -0,0 +1,34 @@
...
+// InitGenesis initializes the capability module's state from a provided genesis
+// state.
+func InitGenesis(ctx sdk.Context, k keeper.Keeper, data types.GenesisState) {
+	for bz, doc := range data.Documents {
+		var key types.GenesisDIDDocumentKey
+		if err := key.Unmarshal(bz); err != nil {
+			panic(err)
+		}
+		k.SetDIDDocument(ctx, key.DID, *doc)
+	}
```
youngjoon-lee, line 18: What do you think to remove the `GenesisDIDDocumentKey` type? It has been defined previously because we had to handle the `DID` type which is not primitive (`string`).
```suggestion
	for did, doc := range data.Documents {
		k.SetDIDDocument(ctx, did, *doc)
	}
```

For now, we don't use type aliasing. So, we can just use string keys in the map as they are, IMHO. I also did it in the PR #124 .

youngjoon-lee: Also, we can check if the `did` is valid, before calling `SetDIDDocument()`.

x/did/types/errors.go
```diff
@@ -0,0 +1,30 @@
...
+// DONTCOVER
+
+import (
+	"fmt"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+// x/did module sentinel errors
+var (
+	ErrDIDExists = sdkerrors.Register(ModuleName, 101, "DID %v already exists")
```
youngjoon-lee, line 12:
```suggestion
	ErrDIDExists = sdkerrors.Register(ModuleName, 1, "DID %v already exists")
```
Let's use error codes starting from `1`. I'm not sure why it was starting from `101` previously, but it seems it's safe to use `1`~.

x/did/types/errors.go
```diff
@@ -0,0 +1,30 @@
...
+	ErrInvalidNetworkID = sdkerrors.Register(ModuleName, 110, "Invalid network ID: %s")
+	ErrInvalidDIDDocumentWithSeq = sdkerrors.Register(ModuleName, 111, "Invalid DIDDocumentWithSeq: %v")
+	ErrDIDDeactivated = sdkerrors.Register(ModuleName, 112, "DID was already deactivated: %v")
+	ErrVerificationMethodKeyTypeNotImplemented = sdkerrors.Register(ModuleName, 114, "Verification not implemented with key type: %v")
+
+)
+
+func Error(error *sdkerrors.Error, args ...interface{}) error {
+	return sdkerrors.New(error.Codespace(), error.ABCICode(), fmt.Sprintf(error.Error(), args))
+}
```
youngjoon-lee, line 30: I think it's better to use the error wrapping, instead of disassembling `sdkerrors.Error` for the string formatting.

For example, when you wanna wrap an `ErrInvalidDID` error with an additional message.
```go
sdkerrors.Wrap(ErrInvalidDID, "did:bitcoin:123125125")
```

Because Go doesn't have exceptions (+ err stacktrace), there have been a lot of discussions how to deal with errors nicely.
Finally, Go team decided to adopt the error wrapping concept: https://blog.golang.org/go1.13-errors#TOC_3. .
With this concept, we can check if the error `is`/`as` a certain error.

Furthermore,
Cosmos team wraps this concept one more time by defining the `sdkerror` package, which is what I don't like :). But, they recommend to use `sdkerrors.Wrap`, so that use can

youngjoon-lee: You can refer to https://github.com/medibloc/panacea-core/blob/d27d8818192eee72c192747a8a9e417346ae20ae/x/token/keeper/msg_server_token.go#L21

x/did/genesis.go
gyuguen: @youngjoon-lee 
I understood it well.
I'll fix it.

x/did/types/errors.go
gyuguen: @youngjoon-lee 
OK. I'll fixed it. 
Start code number 1

x/did/types/errors.go
gyuguen: @youngjoon-lee 

If it is wrapped, the error description is marked as two. 
In this case, `DID %v already exists` and `DID [did:asdfdsafsdafadsfdsafasdfdsf] already exists` are exposed together. 
I don't think it's a covering concept, but it can be seen as an error that occurs here. 

Please give me your opinion.

x/did/types/types.go
```diff
@@ -0,0 +1 @@
+package types
```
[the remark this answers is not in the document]
gyuguen: OK. I'll fixed it

x/did/types/did.go
youngjoon-lee: Oh. I got it. Let me think.

proto/did/did.proto
gyuguen: You're right!
I'll fixed it.

x/did/types/errors.go
youngjoon-lee: Right. The current Cosmos `sdkerrors` sucks ㅜㅠ. The only one choice that we can choose is:
```go
var (
	ErrInvalidName = sdkerrors.Register(ModuleName, 1, "invalid name")  // without %v or %s
)
```
```go
name := "dummy"
return sdkerros.Wrap(ErrInvalidName, name)
```
Then, the composited error message might be:
```
invalid name: dummy
```

proto/did/did.proto
gyuguen: Oh yes. It's not used.
I'll fixed it.

x/did/types/errors.go
youngjoon-lee: In general, the recommended way is:
```go
err := errors.New("mem crash")
err = fmt.Errorf("failed to calculate your age: %w", err)
fmt.Println(err)
```
Then, the err message might be:
```
failed to calculate your age: mem crash
```
which is readable.

But, Cosmos uses their own special way...

youngjoon-lee:
I left some comments, but other parts are really good! Thank you for submitting this PR, even though the `did` is the biggest module. 
Let me also think about the issue: https://github.com/medibloc/panacea-core/pull/129#discussion_r643667936

proto/did/did.proto
youngjoon-lee: I guess we can use https://developers.google.com/protocol-buffers/docs/proto3#oneof, but I have no idea what `gogoproto.wktpointer` is. Let me try.

x/did/types/errors.go
gyuguen: As we said before, modify it into a wrapping error.

proto/did/did.proto
```diff
@@ -6,28 +6,23 @@ option go_package = "github.com/medibloc/panacea-core/x/did/types";
 import "gogoproto/gogo.proto";
 
 message DIDDocument {
-  repeated string contexts = 1 [json_name = "@contexts"];;
-  string ID = 2 [json_name = "id"];
+  string contexts = 1 [json_name = "@context", (gogoproto.jsontag) = "@context", (gogoproto.customtype) = "Contexts"];
```
gyuguen, line 9: (gogoproto.jsontag) : Change json name. Not used jsonpb. Used json

gyuguen: (gogoproto.customtype) : Use custom go type. Required implements MarshalTo, UnmarshalTo(), Size()

x/did/client/rest/rest.go
```diff
@@ -14,12 +14,8 @@ const (
...
 func RegisterRoutes(clientCtx client.Context, r *mux.Router) {
 	// this line is used by starport scaffolding # 2
-	registerQueryRoutes(clientCtx, r)
-	registerTxHandlers(clientCtx, r)
-
-	registerQueryRoutes(clientCtx, r)
-	registerTxHandlers(clientCtx, r)
-
+	/*registerQueryRoutes(clientCtx, r)
+	registerTxHandlers(clientCtx, r)*/
```
gyuguen, line 18: Maybe legacy rest api.

x/did/keeper/msg_server_did.go
```diff
@@ -36,10 +36,10 @@ func (m msgServer) UpdateDID(goCtx context.Context, msg *types.MsgUpdateDID) (*t
 
 	docWithSeq := keeper.GetDIDDocument(ctx, msg.DID)
 	if docWithSeq.Empty() {
-		return nil, types.Error(types.ErrDIDNotFound, msg.DID)
+		return nil, types.ErrorWrapf(types.ErrDIDNotFound, "DID: %s", msg.DID)
```
gyuguen, line 39: Wrapping error.

x/did/types/did.go
```diff
@@ -36,13 +37,13 @@ func NewDID(pubKey []byte) string {
 
 func ParseDID(str string) (string, error) {
 	did := str
-	if !ValidDID(did) {
-		return "", Error(ErrInvalidDID, str)
+	if !ValidateDID(did) {
```
gyuguen, line 40: Change ValidXXX to ValidateXXX

x/did/types/did.go
```diff
@@ -234,27 +227,179 @@ func (doc DIDDocument) VerificationMethodFrom(relationships []*VerificationRelat
 	return VerificationMethod{}, false
 }
 
-func ValidContexts(ctxs []string) bool {
-	if len(ctxs) == 0 || ctxs[0] != ContextDIDV1 { // the 1st one must be ContextDIDV1
+type Contexts []Context
```
gyuguen, line 230: From here, contexts hand implementation of the part.

x/did/keeper/msg_server_did.go
youngjoon-lee: How about just using `sdkerrors.Wrapf()` directly? I think the `types.ErrorWrapf` has no point because it's exactly the same as `sdkerrors.Wraps()`.

x/did/client/rest/rest.go
youngjoon-lee: Yeah. It's okay to comment them out for now.

We may need to re-enable the legacy REST API. Medipass basically uses our Java SDK, but some insurance companies uses our REST API directly, IIRC.
If it's too difficult to enforce them to upgrade their logic, we may need to provide the legacy REST API with them for a while.
In that case, we also need to test whether the legacy API in Stargate is compatible with the previous one.

x/did/keeper/msg_server_did.go
gyuguen: Yes. It's better way
I'll fixed it

gyuguen: FIxed : https://github.com/medibloc/panacea-core/pull/129/commits/699999f03ceedb287b4d6c256946afa65f7699de

x/did/types/did.go
youngjoon-lee: Let me also test this manually. I'm not sure if this implementation works well because it's a bit complicated.

youngjoon-lee: I tested that it works well. But, what do you think about making it simpler like this: https://github.com/medibloc/panacea-core/commit/89cc7637c09319e51c700b2efe70b4ba9dd21501.
I referred to the following guides:
- https://github.com/gogo/protobuf/blob/master/custom_types.md
- https://github.com/gogo/protobuf/blob/master/test/t.go

gyuguen: @youngjoon-lee 
Thank you.
It has been modified more simply.

https://github.com/medibloc/panacea-core/pull/129/commits/1ccd1847596973556f4908f9e000a9d7b05795b6

gyuguen:
> I think the `JSONStringOrStrings` doesn't need to have `EmptyDID()` and `ValidDID()`, because it's not used only for DID. IMHO, it would be better to decouple the `JSONStringOrStrings` with DID concepts.

Yes. That's right.
I'm changed this commit!

https://github.com/medibloc/panacea-core/pull/129/commits/4a824c2ef20970da49dc40ae3f34ea70212764f8

youngjoon-lee (approved):
LGTM 👍 I appreciate that you've managed to migrate this big module.

PR #130, opened 2021-06-01
feat: stargate: Add unit tests for `token`-related types

youngjoon-lee:
Just added simple tests that check `Token` type validation logics.

TODO:
- Test keeper, handler, and so on (by following [Cosmos' idea](https://github.com/cosmos/cosmos-sdk/blob/d7c5bc5f35505e1d64e9b3521067871947d55bfa/x/bank/keeper/keeper_test.go#L67), I guess)

### DIFF
--- a/x/token/types/types_test.go
+++ b/x/token/types/types_test.go
@@ -1,0 +1,116 @@
+package types
+
+import (
+	"os"
+	"testing"
+
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestMain(m *testing.M) {
+	config := sdk.GetConfig()
+	config.SetBech32PrefixForAccount("panacea", "panaceapub")
+	config.Seal()
+
+	os.Exit(m.Run())
+}
+
+func TestNewSymbol(t *testing.T) {
+	symbol := NewSymbol("lov", "1234567890ABCDEFGHIJKLMNOPQRSTUXYZ")
+	require.Equal(t, "LOV-123", symbol)
+}
+
+func TestValidateName(t *testing.T) {
+	require.NoError(t, validateName("my name"))
+	require.ErrorIs(t, validateName(""), ErrInvalidName)
+
+	maxLenName := "12345678901234567890123456789012"
+	require.NoError(t, validateName(maxLenName))
+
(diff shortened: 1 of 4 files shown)


x/token/types/types_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestMain(m *testing.M) {
+	config := sdk.GetConfig()
+	config.SetBech32PrefixForAccount("panacea", "panaceapub")
+	config.Seal()
+
+	os.Exit(m.Run())
+}
```
gyuguen, line 20: @youngjoon-lee 
What kind of test is this?

youngjoon-lee: https://golang.org/pkg/testing/#hdr-Main
This is something like `setUp()` in JUnit.
The `m.Run()` executes all tests in that file. Before that, we can run some initialization logics.

In this file, I called `config.SetBech32PrefixForAccount("panacea", "panaceapub")` because it's needed for [sdk.AccAddressFromBech32](https://github.com/medibloc/panacea-core/pull/124/files#diff-b3d3e6a05fd122fd9200d46ef1f333b675e0b16cfa756dbb2f979f21b5d135ffR29).
Actually, the `SetBech32PrefixForAccount` is already being called in the `app/prefix.go`, but it cannot be imported to `x/**/*_test.go` (due to the circular import).

In near future, we need to have a common test initialization function.

gyuguen: OK. Thanks!

gyuguen (approved):
👍

x/token/types/types_test.go
```diff
@@ -0,0 +1,116 @@
+package types
```
gyuguen, line 1: @youngjoon-lee 
How about changing the package name to `package types_test`?
The variables declared in the test can be shared to the actual service, so it would be good to be separated.

gyuguen: Of course, Can not tested private method...

youngjoon-lee: Ya. Good point.
It's a good practice to define `*_test` packages for black-box tests, but then we cannot test private functions.
Here, we cannot use the `types_test` package because we're testing private `validate*` functions ㅜㅠ.

gyuguen: Oh, I understand.

gyuguen (approved):
👍

PR #131, opened 2021-06-03
feat: stargate: Integrate swagger doc with `panacead`

youngjoon-lee:
Now, `panacead` exposes `0.0.0.0:1317/swagger` (if `api.swagger` in the `app.toml` is `true`).

For that, `make update-swagger-docs` must be executed before running `make build`, because it generates a `statik.go` file which is used by `app/app.go`. 
The all-in-one command is `make all` (or `make build-all`).
In general, Panacea users need to run only `make install`, because Panacea contributors are responsible for committing Go files which are generated from the latest proto files.

Note that all files in the `client/docs/` are copied from the [cosmos-sdk](https://github.com/cosmos/cosmos-sdk/tree/v0.42.4/client/docs) repo. Don't needs to be reviewed seriously.


gyuguen (approved):
👍

PR #132, opened 2021-06-03
fix: stargate: Remove unnecessay `x/panacea`

youngjoon-lee:
`x/panacea` was auto-generated by `starport app ...`.


gyuguen (approved):
👍

youngjoon-lee:
CI keeps failing because the `proto/` directory was deleted.
CI will be passed after other `x/*` modules which have `proto/**/*.proto` files are merged to the `stargate` branch.

PR #134, opened 2021-06-07
chore: stargate: Remove unnecessary Vue codes

youngjoon-lee:
Starport has generated [Vue](https://vuejs.org/) codes for the web UI as below, but we don't need it.
https://medium.com/tendermint/launch-a-new-token-and-send-it-across-the-cosmos-starport-v0-12-is-here-d1578012ba09
image

Furthermore, Vue codes are based on "types" that were generated by Starport (that have fields that you don't want).
**So, it's annoying for us to maintain Vue codes as well for now.**

Let's delete the `vue/` directory.
We can rewrite them from the scratch, when we really need an web UI.


gyuguen (approved):
👍

PR #135, opened 2021-06-07
feat: stargate: Add `aol` module

youngjoon-lee:
I manually tested all operations.

You don't need to review commits separately. **I recommend you to review the whole commits at once** (not that much because Vue codes have been removed).

Unlike other custom modules, the `x/aol` has complicated data structures.
Thus, I've implemented a `compkey` package that can be used for keeper keys and genesis keys.

FYI, the genesis looks like:
```json
    "aol": {
      "owners": {
        "panacea18et7w2hhrlacv0y8r9rr80rcu5xhgk6xvlscx2": {
          "totalTopics": "1"
        }
      },
      "records": {
        "panacea18et7w2hhrlacv0y8r9rr80rcu5xhgk6xvlscx2/food/0": {
          "key": "azE=",
          "nanoTimestamp": "1623042738359010000",
          "value": "djE=",
          "writerAddress": "panacea18et7w2hhrlacv0y8r9rr80rcu5xhgk6xvlscx2"
        }
      },
      "topics": {
        "panacea18et7w2hhrlacv0y8r9rr80rcu5xhgk6xvlscx2/food": {
          "description": "",
          "totalRecords": "2",
          "totalWriters": "1"
        }
      },
      "writers": {
        "panacea18et7w2hhrlacv0y8r9rr80rcu5xhgk6xvlscx2/food/panacea18et7w2hhrlacv0y8r9rr80rcu5xhgk6xvlscx2": {
          "description": "",
          "moniker": "node1",
          "nanoTimestamp": "1623043288079266000"
        }
      }
    }
```

### TODO
- Add tests


x/aol/client/cli/txRecord.go
```diff
@@ -2,94 +2,30 @@ package cli
...
-}
-
-func CmdUpdateRecord() *cobra.Command {
-	cmd := &cobra.Command{
-		Use:   "update-record [id] [key] [value] [nanoTimestamp] [writerAddress]",
-		Short: "Update a record",
-		Args:  cobra.ExactArgs(5),
-		RunE: func(cmd *cobra.Command, args []string) error {
-			id, err := strconv.ParseUint(args[0], 10, 64)
+			feePayerAddress, err := cmd.Flags().GetString(flagFeePayer)
```
gyuguen, line 23: Who is `feePayerAddress`?
It looks like not previouslycode

youngjoon-lee: For the `add-record` tx, we've adopted a 'fee payer' concept who pay fees on behalf of the 'writer':
https://github.com/medibloc/panacea-core/blob/985271f795c2ff0bf1f6b63152d97fabb41dbfa2/x/aol/client/cli/tx.go#L152

x/aol/client/cli/txTopic.go
```diff
@@ -1,93 +1,36 @@
...
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			id, err := strconv.ParseUint(args[0], 10, 64)
+			clientCtx, err := client.GetClientTxContext(cmd)
 			if err != nil {
 				return err
 			}
 
-			clientCtx, err := client.GetClientTxContext(cmd)
+			description, err := cmd.Flags().GetString(flagDescription)
```
gyuguen, line 28: `description` is optional?

youngjoon-lee: yeah. it has been optional: https://github.com/medibloc/panacea-core/blob/985271f795c2ff0bf1f6b63152d97fabb41dbfa2/x/aol/client/cli/tx.go#L79

youngjoon-lee: I'll change codes to use `viper.GetString()`.

x/aol/client/cli/txRecord.go
gyuguen: I understand. Thanks

types/compkey/compkey.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	return bz
+}
+
+// encode encodes multiple byte slices into a byte slice.
+// If the size of some byte slices are not in uint8, it returns a non-nil error.
+func encode(values [][]byte) ([]byte, error) {
+	size := 0
+	for _, value := range values {
+		size += sizeUint8 + len(value)
```
gyuguen, line 64: I think `uint8` max size is 255. So max value size is 255.
How about used size type to `uint16` in preparation for the future?

youngjoon-lee: Right. I'm also thinking which one is better.
Currently, 255 is enough because `sdk.AccAddress` is 20-bytes and `topicName` is 70-bytes max.
In general, the shorter the key, the better the performance. So, it might be reasonable to limit the size of each part to 255, so that the key is not too long. 
But, as you said, it's a trade-off. Let me think more.

x/aol/keeper/grpc_query_topic.go
```diff
@@ -0,0 +1,71 @@
...
+	ctx := sdk.UnwrapSDKContext(c)
+
+	ownerAddr, err := sdk.AccAddressFromBech32(req.OwnerAddress)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid owner address")
+	}
+	compKeyPrefix, err := compkey.PartialEncode(&types.TopicCompositeKey{OwnerAddress: ownerAddr, TopicName: ""}, 1)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "fail to encode topic key")
+	}
```
gyuguen, line 52: I think it's possible to check what error occurred.
How about to write log or append to `err.Error()`?

youngjoon-lee: good point. I'll use `err.Error()`.

youngjoon-lee: 5c9556d38eb49e4be2bc9e7cd56a203cbf99f03c Done :)

x/aol/keeper/msg_server_record.go
```diff
@@ -0,0 +1,51 @@
...
+	if !k.HasTopic(ctx, topicKey) {
+		return nil, sdkerrors.Wrapf(types.ErrTopicNotFound, "topic <%s, %s>", msg.OwnerAddress, msg.TopicName)
+	}
+	writerKey := types.WriterCompositeKey{OwnerAddress: ownerAddr, TopicName: msg.TopicName, WriterAddress: writerAddr}
+	if !k.HasWriter(ctx, writerKey) {
+		return nil, sdkerrors.Wrapf(types.ErrWriterNotAuthorized, "writer <%s, %s, %s>", msg.OwnerAddress, msg.TopicName, msg.WriterAddress)
+	}
+
+	topic := k.GetTopic(ctx, topicKey)
+	offset := topic.NextRecordOffset()
```
gyuguen, line 34: I have 1 questions.

If a another node two record request, I think sequence is duplicated.

1. `node-1` : request add record (topic : id, record : jang)
2. `node-2` : request add record (topic : id, record : lee)
3. `node-1` : get topic (topic : id, totalRecord : 1)
4. `node-2` : get topic (topic : id, totalRecord : 1)
5. `node-1` : add record (topic : id, record : jang, offset : 2)
6. `node-2` : add record (topic : id, record : lee, offset : 2)
-- create block --

and get record `offset : 2` result two records.

Umm.. There's a potential risk factor. Is not problem?

youngjoon-lee: Yeah. That's the key of the Tendermint Consensus.

All state transitions are executed sequentially, and the sequence of state transitions is synced across all nodes. That's what Tendermint guarantees.

As you mentioned at `1` and `2`, each node can have a different sequence of transactions in their [mempool](https://docs.tendermint.com/master/tendermint-core/mempool.html) like below:
|node-1|node-2|
|---|---|
|`add-record("jang")`|`add-record("yj")`|
|`add-record("yj")`|`add-record("jang")`|
It can happen frequently in distributed systems, because some clients send requests to the node-1 and others send requests to the node-2. 

However, the final sequence is determined by a block proposer.
If the node-2 is a block proposer in the `Round X`, the node-2 assemble a new block as he want (according to his mempool).
```
- Block X
    - tx1: add-record("yj")
    - tx2: add-record("jang")
```
This block is broadcasted to the node-1. Then, the node-1 plays transactions in the order specified in the block. Here, the node-1 truncates his mempool (His mempool is not meaningful anymore because he isn't a block proposer).

By this consensus mechanism, all nodes will execute all state transitions in the same order.

types/compkey/compkey.go
gyuguen: I got it. Thanks

x/aol/keeper/msg_server_record.go
gyuguen: I understand. Thanks

gyuguen (approved):
👍

PR #136, opened 2021-06-08
feat: stargate: Add did module test

gyuguen:
Added test to DID module. 
We have created a test suite package for integration testing.

- Integration test helper package(testsuite) : https://github.com/medibloc/panacea-core/commit/f0ce44a5225046ba72ee99fc70ee00adbff0d8ea
- Apply testsuite to burn module test : https://github.com/medibloc/panacea-core/commit/e3057f0dfe7cc652090fcff90530c29938ab48f9
- Add `did` module test 
    - `client` : https://github.com/medibloc/panacea-core/commit/74af8e3426bacd07d38a8584c21679e7000059aa
    - `types` : https://github.com/medibloc/panacea-core/commit/1cc921a691758cdd4706e3409ac7970a1ee8a598
    - `keeper` : https://github.com/medibloc/panacea-core/commit/ca85e430ca4a1ce11638e48a62a30f4e34317213
    - `genesis/handler/module` : https://github.com/medibloc/panacea-core/commit/059a8ca4f5d1dc9fcdbb75d327053eb3997377e7


types/testsuite/suite.go
```diff
@@ -0,0 +1,128 @@
...
+	Ctx sdk.Context
+
+	AccountKeeper authkeeper.AccountKeeper
+	BankKeeper    bankkeeper.Keeper
+	BurnKeeper    burnkeeper.Keeper
+	DIDMsgServer  didtypes.MsgServer
+	DIDKeeper     didkeeper.Keeper
+}
+
+func (suite *TestSuite) SetupTest() {
```
gyuguen, line 40: It is performed each time the test is run.
Initial setup is in progress.

types/testsuite/suite.go
```diff
@@ -0,0 +1,128 @@
...
+	suite.BurnKeeper = *burnkeeper.NewKeeper(suite.BankKeeper)
+	suite.DIDKeeper = *didkeeper.NewKeeper(
+		cdc.Marshaler,
+		keyParams[didtypes.StoreKey],
+		memKeys[didtypes.MemStoreKey],
+	)
+	suite.DIDMsgServer = didkeeper.NewMsgServerImpl(suite.DIDKeeper)
+}
+
+func (suite *TestSuite) BeforeTest(suiteName, testName string) {
```
gyuguen, line 105: If there is something to be done for each test (e.g., initializing the file), you can implement it.

types/testsuite/suite.go
```diff
@@ -0,0 +1,128 @@
...
+	log.NewNopLogger().Info("Pass BeforeTest. suiteName: %s, testName: %s", suiteName, testName)
+}
+
+func (suite *TestSuite) AfterTest(suiteName, testName string) {
```
gyuguen, line 109: If there is anything to be done after each test (for example, deleting the test file), you can implement it.

types/testsuite/suite.go
```diff
@@ -0,0 +1,128 @@
...
+	log.NewNopLogger().Info("Pass AfterTest. suiteName: %s, testName: %s", suiteName, testName)
+}
+
+func newTestCodec() params.EncodingConfig {
+	cdc := codec.NewLegacyAmino()
+	interfaceRegistry := codectypes.NewInterfaceRegistry()
+	authtypes.RegisterInterfaces(interfaceRegistry)
+	banktypes.RegisterInterfaces(interfaceRegistry)
+	didtypes.RegisterInterfaces(interfaceRegistry)
```
gyuguen, line 118: If you have an interface, you have to declare it here.

x/did/keeper/msg_server_did_test.go
```diff
@@ -0,0 +1,71 @@
+package keeper
```
gyuguen, line 1: If the package declares the same as the actual logic, an error due to the circular reference will occur.
So `TestSuite` is not available.
The test recommends another package (`xxx_test`).

x/did/types/auth_test.go
```diff
@@ -0,0 +1,68 @@
+package types
```
gyuguen, line 1: This test can also be caught in a circular reference, so use the default 'suite'.

x/did/types/did.go
```diff
@@ -22,7 +22,7 @@ const (
 type JSONStringOrStrings []string
 
 func EmptyDIDs(strings []string) bool {
-	if strings == nil || len(strings) == 0 {
```
gyuguen, line 25: Arrays type empty check can be use method `len`. 
So I deleted check nil

gyuguen: (lint check failed...)

x/did/keeper/msg_server_did_test.go
youngjoon-lee: Ya. It's tricky to use the `keeper_test` package name here, because we're testing private stuff: `verifyDIDOwnership()`. 

Actually, this test file doesn't need to use the `testsuite`, since we don't need actual keepers in these tests. 

In the future, if we really need to use the `testsuite`, let's consider to define some interfaces as you said.

youngjoon-lee: Ah. Currently, the `x/did/handler_test.go` contains tests for `msgServer.CreateDID()` and so on. I think it's better to move those tests to this file: `msg_server_did_test.go`, because the `handler.go` is just an wrapper of `msg_server_did.go` from now on. Then, we need keepers for testing and we'll face the circular reference as you mentioned.
Let's me think.

x/did/types/auth_test.go
youngjoon-lee: Yeah. But, I think this test doesn't need to use the test suite actually, because it doesn't require any keepers.

x/did/keeper/msg_server_did_test.go
youngjoon-lee: IMHO, it would be okay to change `verifyDIDOwnership()` to `VerifyDIDOwnership()`, so that you can use the `keeper_test` package name.

It'd be simpler than defining some interfaces (that might be over-engineering).

youngjoon-lee (approved):
LGTM! Thank you for making the `testsuite` which is much better than mocking keepers.
I think this PR can be merged, but please consider about this [comment](https://github.com/medibloc/panacea-core/pull/136#discussion_r647116517).

x/did/keeper/msg_server_did_test.go
gyuguen: OK, I'll moved to `msg_server_did_test`

umm.. I think it would be good to proceed in two ways.

1. Integration tests use `testsuite` (package : `keeper_test`)
2. private test use `suite` (package : `keeper` )

gyuguen: Oh, I don't have to use the suite.

youngjoon-lee: Yeah. The test suite has nice concepts such as `BeforeTest` and `AfterTest`. But, if the test is super simple, I think it's okay to not use the test suite. The basic Go `testing` package would be enough.

gyuguen: OK.

PR #137, opened 2021-06-08
feat: stargate: Add verificationRelationship parsing tests to did

gyuguen:
Parsing for `VerificationRelationship` is already well applied.
Only supplement parsing test code.


youngjoon-lee (approved):
Aha. I got it that it works well because we've implemented MarshalJSON and UnmarshalJSON: https://github.com/medibloc/panacea-core/blob/5312f7c535181f4065bf243e275b0444652c8afe/x/did/types/did.go#L452

It would be perfect if we simplify the proto file using `oneof` or something else, so that the proto message contains only one field (it currently has two fields). 
But, let's improve it later. I'm not sure if it can be resolved in the proto file.

gyuguen:
> But, let's improve it later. I'm not sure if it's even possible.

OK. I'll check it out later.

PR #138, opened 2021-06-09
feat: stargate: Change did module rest api url and error code

gyuguen:
Change DID module rest api url
Change DID module error code to start with 2


youngjoon-lee (approved):
👍

PR #139, opened 2021-06-14
fix: stargate: Change `aol`/`token` REST API paths and error codes

youngjoon-lee:
similar as #138 , but for `aol` and `token` modules.


gyuguen (approved):
👍

Issue #108, opened 2021-05-14
Road to Stargate (Cosmos v0.42)


youngjoon-lee:
Target: Cosmos SDK [v0.42.4](https://github.com/cosmos/cosmos-sdk/releases/tag/v0.42.4)

Updating Panacea code
https://tutorials.cosmos.network/launchpad-to-stargate/tutorial/01-introduction.html
https://docs.cosmos.network/master/migrations/app_and_modules.html

Upgrading the Mainnet chain
https://docs.cosmos.network/master/migrations/chain-upgrade-guide-040.html

Node operators:
- Exchanges
- Validators (if they join before the Stargate upgrade)

Components that must be upgraded (order by priority):
- SDKs
  - Typescript SDK: https://github.com/medibloc/panacea-js
    - consider replacing it with https://github.com/cosmos/cosmjs (if possible)
  - Java SDK: https://github.com/medibloc/panacea-java
- Utilities
  - Explorer: https://github.com/medibloc/explorer (depending on panacea-js)
  - Web Wallet: https://github.com/medibloc/wallet (depending on panacea-js)
  - Key Server: https://github.com/medibloc/keyserver
- 3rd-party Utilities:
  - Ledger App: https://github.com/medibloc/ledger-panacea (not published yet to the Ledger Appstore)

Client services that must be upgraded:
- Medipass: Java SDK
  - [신한생명](https://www.notion.so/medibloc/630d2377bdf1483aa5d94d9c8efe0944?v=33c50dc92e864fed9a7a54ebf1a66f4e): REST API
- [Exchanges](https://www.notion.so/medibloc/db3056ae4be04ea1b3a13678f89679bc?v=082f73cae6bc4617b584657fdb31b415)
  - Upbit, Bittrex: REST API
  - Gate.io, BithumbSG: Typescript SDK (Javascript, actually)
  - Korbit, Coinrail: Java SDK

youngjoon-lee:
Consider using 1-byte key prefixes instead of [long prefixes](https://github.com/medibloc/panacea-core/blob/d243dbc4f76eac04df4f1d6af6f5da82f10281cc/x/aol/types/keys.go#L37) which were generated by Starport.

Actually, we've used 1-byte prefixes so far. Also, the latest Cosmos modules also use 1-byte prefixes.

PR #141, opened 2021-06-14
feat: Add a `x/wasm` module

youngjoon-lee:
# CAUTION!!!!
Merge this PR directly to the `master`, after the `stargate` branch is merged to the `master`.
It's because this PR is not the part of the issue #108.

# Changes
The Starport has a command: [`starport module import wasm`](https://github.com/tendermint/starport/pull/642), but unfortunately, it doesn't work well. Also, it's based on the old CosmWasm v0.14.

So, I imported CosmWasm v0.17 manually by referring to https://github.com/CosmWasm/wasmd/blob/v0.17.0/app/app.go

I tested it manually using the [Escrow](https://docs.cosmwasm.com/0.14/getting-started/compile-contract.html) example.


go.mod
```diff
@@ -15,8 +16,8 @@ require (
 	github.com/onsi/ginkgo v1.15.0 // indirect
 	github.com/onsi/gomega v1.10.5 // indirect
 	github.com/pborman/uuid v1.2.0
+	github.com/prometheus/client_golang v1.10.0
```
gyuguen, line 19: I remember `prometheus` is monitoring tools.
Is used here code?
or auto importing this?

youngjoon-lee: Yeah. The Stargate supports the [Telemetry](https://docs.cosmos.network/master/core/telemetry.html) feature which collects some metrics to Prometheus, even though I haven't read it seriously.

`x/wasm` modules have some telemetry-related options: https://github.com/CosmWasm/wasmd/blob/d2e35c249e39e90c5a85e03e0b75800431612b20/cmd/wasmd/root.go#L205

gyuguen: OK. I got it!

gyuguen (approved):
👍

PR #142, opened 2021-06-15
fix: stargate: Refine proto package names

youngjoon-lee:
Followed the scheme of Cosmos' proto package names: https://github.com/cosmos/cosmos-sdk/tree/v0.42.5/proto/cosmos.
```
panacea.{module}.{ver}.*
```

It's reasonable since we've decided to define REST paths as `panacea/{module}/{ver}/...`.


gyuguen (approved):
👍 
Thank you with changed `burn` and `did` module

PR #143, opened 2021-06-16
fix: stargate: Change proto path: `proto/*` to `proto/panacea/*`

youngjoon-lee:
Sorry for opening a PR similar as #142 

As [Cosmos](https://github.com/cosmos/cosmos-sdk/tree/master/proto), I just changed the directory structure:
```
mv proto/* proto/panacea/ 
```
It's so intuitive when importing these proto files from other projects.
```
import "panacea/aol/v2/..."
```


gyuguen (approved):
👍

PR #145, opened 2021-06-21
fix: stargate: Support multi-signers in CLI (for `tx aol add-record`)

youngjoon-lee:
How to sign with multi-signers is already described in the [Cosmos SDK doc](https://docs.cosmos.network/v0.42/run-node/txs.html#signing-a-transaction-2).
But, Cosmos SDK doesn't provide an all-in-one function for it.

Thus, I implemented it.

I tested it manually using `panacead tx aol add-record ...` for now, since it's not simple to write unit tests due to Keyring and Panacea daemon. But, I'll try to write tests somehow soon.

### DIFF
--- a/x/aol/client/cli/txRecord.go
+++ b/x/aol/client/cli/txRecord.go
@@ -1,7 +1,18 @@
 package cli
 
 import (
+	"bufio"
+	"fmt"
+	"os"
+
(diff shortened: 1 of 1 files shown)


gyuguen (approved):
👍 
Very nice!

PR #147, opened 2021-06-24
feat: stargate: Change did sign marshaller json to protobuf

gyuguen:
Change did sign marshaler json to protobuf.
Because we fade out to amino marshaler.


x/did/keeper/msg_server_did.go
```diff
@@ -65,8 +65,11 @@ func (m msgServer) DeactivateDID(goCtx context.Context, msg *types.MsgDeactivate
 		return nil, sdkerrors.Wrapf(types.ErrDIDDeactivated, "DID: %s", msg.DID)
 	}
 
-	signableDID := types.SignableDID(msg.DID)
-	newSeq, err := VerifyDIDOwnership(signableDID, docWithSeq.Seq, docWithSeq.Document, msg.VerificationMethodID, msg.Signature)
+	doc := types.DIDDocument{
```
gyuguen, line 68: Instance of using the did, use the DIDDocument

gyuguen: Because did is string. But not implements protobuf marshal

youngjoon-lee: Yeah. It makes sense even though the user must get a DID document before deactivating it.

youngjoon-lee (approved):
LGTM

x/did/keeper/msg_server_did.go
youngjoon-lee: Yeah. It makes sense.

PR #148, opened 2021-06-24
fix: stargate: Fixed aol topic bug when change total count

gyuguen:
There is a bug where the description is removed when add record and add/delete writer.
I fixed not removed description.


youngjoon-lee (approved):
Thank you!

PR #150, opened 2021-06-28
feat: stargate: Add aol, token test code

gyuguen:
Add tests for `aol` and `token` module


x/aol/keeper/owner_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+	// verify GetAllOwner
+	resultKeys, resultOwners := aolKeeper.GetAllOwners(ctx)
+	suite.Require().Equal(1, len(resultKeys))
+	suite.Require().Equal([]aoltypes.OwnerCompositeKey{key}, resultKeys)
+	suite.Require().Equal(1, len(resultOwners))
+	suite.Require().Equal([]aoltypes.Owner{owner}, resultOwners)
+}
+
+func (suite ownerTestSuite) TestMultiOneOwner() {
```
youngjoon-lee, line 60: If I understand it correctly, I think we can name this test as `TestMultiOwners`.

youngjoon-lee (approved):
Thank you so much for your effort!

x/aol/keeper/owner_test.go
gyuguen: That's right! I changed method name

PR #152, opened 2021-06-30
feat: stargate: Use 1-byte key prefix for keeper

### LINKED ISSUE #149: stargate: Consider 1-byte key prefix for keeper
Related from #108

For better performance, consider using 1-byte key prefixes instead of [long prefixes](https://github.com/medibloc/panacea-core/blob/d243dbc4f76eac04df4f1d6af6f5da82f10281cc/x/aol/types/keys.go#L37) which were generated by Starport.

Actually, we've used 1-byte prefixes so far: https://github.com/medibloc/panacea-core/blob/985271f795c2ff0bf1f6b63152d97fabb41dbfa2/x/aol/keeper/keeper_keys.go#L12.
Also, the latest Cosmos modules also use 1-byte prefixes: https://github.com/cosmos/cosmos-sdk/blob/62147290c13c6dd4c4505b33da4a4a8a90687cd0/x/slashing/types/keys.go#L32

youngjoon-lee:
Closes #149


x/did/keeper/did.go
```diff
@@ -7,15 +7,15 @@ import (
 )
 
 func (k Keeper) SetDIDDocument(ctx sdk.Context, did string, doc types.DIDDocumentWithSeq) {
-	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefix(types.DIDKey))
-	key := DIDDocumentKey(did)
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.DIDKeyPrefix)
+	key := []byte(did)
```
youngjoon-lee, line 11: The `keeper/keeper_keys.go` in the `x/did` is not necessary anymore. From Stargate, the definition of keeper key prefixes has been moved to the `types/keys.go` (`DIDKeyPrefix`). 

In the current implementation in the `stargate` branch, the keeper key is double-prefixed (by [DID-value-](https://github.com/medibloc/panacea-core/blob/f3d98c87dee37f4fb43b85f6b16116ced55fed1e/x/did/types/keys.go#L25) and [0x11](https://github.com/medibloc/panacea-core/blob/f3d98c87dee37f4fb43b85f6b16116ced55fed1e/x/did/keeper/keeper_keys.go#L10)). We've missed it.

gyuguen: Ah. We where using a duplicate prefix! 
I understand.

gyuguen (approved):
👍

PR #157, opened 2021-07-05
feat: Add git flow to make docker image

gyuguen:
When master branch changed and add new tag then build docker image and publish git package.
There were a few things to consider.
- Required of wasmvm module `libwasmvm.so` library 
- docker image build and publish([link](https://github.com/docker/build-push-action))

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -1,0 +1,59 @@
+name: Docker
+
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+on:
+  push:
+    branches: [ master ]
+    # Publish semver tags as releases.
+    tags: [ 'v*.*.*' ]
+
+env:
+  # Use docker.io for Docker Hub if empty
+  REGISTRY: ghcr.io
+  # github.repository as <account>/<repo>
+  IMAGE_NAME: ${{ github.repository }}
+
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      # Login against a Docker registry except on PR
+      # https://github.com/docker/login-action
+      - name: Log into registry ${{ env.REGISTRY }}
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@28218f9b04b4f3f62068d7b6ce6ca5b26e35336c
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      # Extract metadata (tags, labels) for Docker
+      # https://github.com/docker/metadata-action
+      - name: Extract Docker metadata
+        id: meta
+        uses: docker/metadata-action@98669ae865ea3cffbcbaa878cf57c20bbf1c6c38
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,59 @@
...
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      # Login against a Docker registry except on PR
+      # https://github.com/docker/login-action
+      - name: Log into registry ${{ env.REGISTRY }}
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@28218f9b04b4f3f62068d7b6ce6ca5b26e35336c
```
youngjoon-lee, line 37: Is there any reason why use a specific commit hash instead of the semver (`docker/login-action@v1`).

gyuguen: This is an automatically generated hash. 
I will test it after changing it to the latest version!

gyuguen: I updated current tag and the test result was successful in upload the git package.
Thank you review.

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+COPY . /src/panacea-core
+
+# Set working directory for the 'panacea-core' build
+WORKDIR /src/panacea-core
+
+# Install panacea-core
+RUN make clean && make build
+
+# Get 'libwasmvm.so' from wasmvm
+RUN git clone -b v0.14.0 https://github.com/CosmWasm/wasmvm.git /src/wasmvm
```
youngjoon-lee, line 21: Yeah. There's no other way to get `libwasmvm.so` because `CosmWasm/wasmvm` is indirectly referred by `panacea-core` (The `panacea-core` refers to `CosmWasm/wasmd` which refers to `CosmWasm/wasmvm`). 

Now, we must keep in mind that the version of `wasmvm` must be always updated whenever we update the `wasmd` (in the `go.mod`).

Btw, I just found that the `wasmvm` also provides a `libwasmvm_muslc.a` for Alpine Linux.
https://github.com/CosmWasm/wasmd/blob/d2e35c249e39e90c5a85e03e0b75800431612b20/Dockerfile#L18
It would be better to use that, but not for now because that's a static library (not a shared library). It means that we have to change our Makefile to build the `panacead` with the static library. So, let's do it later. For now, your strategy is good enough.

gyuguen, line 21: It is same version used by wasmd(v0.17.0)
If change the wasmd version, wasmvm version needs to be update too.

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,59 @@
...
+
+      # Build and push Docker image with Buildx (don't push on PR)
+      # https://github.com/docker/build-push-action
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v2.6.1
+        with:
+          context: .
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
```
youngjoon-lee, line 59: Aha. I got it.

To check if I understood it correctly, the docker image will be pushed: 
- When the `master` branch is updated, an image is pushed with the tag `master`.
- Or,
- When the `vX.Y.Z` git tag is created, an image is pushed with the tag `vX.Y.Z` and the `latest` docker tag is updated. 

Is it right?

gyuguen: That's right.
I test my packages.
스크린샷 2021-07-05 오전 10 14 58

youngjoon-lee (approved):
LGTM 👍

Dockerfile
gyuguen: OK!

PR #166, opened 2021-07-07
fix: Add `proto-lint` and Fix proto files for importing `genesis.json` properly

### LINKED ISSUE #154: Test the `panacead migrate` command
As described in the [cosmoshub upgrade guide](https://github.com/cosmos/gaia/blob/main/docs/migration/cosmoshub-3.md#upgrade-procedure), we probably need to run the `panacead migrate` command to migrate an exported genesis file from `panacea-2` to the genesis file for `panacea-3`.
The `panacead migrate` command may already have some migration logic implemented by cosmos-sdk because the genesis file format might have changed in the Stargate.

Thus, let's check if the current `panacead migrate` command in the `stargate` branch is enough for our upgrade scenario. If not, add more migration logic.

youngjoon-lee:
Closes #154 

Now, we can import a `genesis.json` successfully.

This PR has many changes, but not that complicated.
- Used snake-case for field names in proto files as discussed in the https://github.com/medibloc/panacea-core/issues/154#issuecomment-875446675
  - except `did.proto`, but I guess we can make it as snake-case as well. Let's do it as a next PR.
- Added `(gogoproto.customtype) = "VerificationRelationship"` which has been missed.
  - It has a silly issue, but not that critical: https://github.com/gogo/protobuf/issues/478
- Added `make proto-lint` which checks all `*.proto` files according to the rule in the `buf.yaml`.
  - inspired by cosmos-sdk
  - It returned the following error. So, I fixed all non-standard request/response names.
    > {"path":"proto/panacea/token/v2/query.proto","start_line":17,"start_column":53,"end_line":17,"end_column":74,"type":"RPC_RESPONSE_STANDARD_NAME","message":"RPC response type \"QueryAllTokenResponse\" should be named \"TokenAllResponse\" or \"QueryTokenAllResponse\"."}
  - Also, it returned another error. So, I added comments for all.
    > {"path":"proto/panacea/token/v2/query.proto","start_line":25,"start_column":9,"end_line":25,"end_column":27,"type":"COMMENT_FIELD","message":"Field \"symbol\" should have a non-empty comment for documentation."}
{"path":"proto/panacea/token/v2/query.proto","start_line":28,"start_column":1,"end_line":30,"end_column":2,"type":"COMMENT_MESSAGE","message":"Message \"QueryGetTokenResponse\" should have a non-empty comment for documentation."}


gyuguen:
So, are we basically implementing snake_case? (did is an exception)

gyuguen:
And these changes affect other sdk as well. If it becomes merge, I will also modify `panacea-java`.

youngjoon-lee:
> So, are we basically implementing snake_case? (did is an exception)

Yes. Now, we have all as snake-case, except `did.proto`.

> And these changes affect other sdk as well. If it becomes merge, I will also modify panacea-java.

True. It will cause a lot of SDK changes, but there's no choice. It'd be better to refine these proto files now before it's too late :) I'll also modify panacea-js as soon as this PR is merged.

gyuguen (approved):
👍 
Thank you for your efforts.

PR #169, opened 2021-07-08
fix: Use the protobuf `oneof` keyword for `VerificationRelationship`

youngjoon-lee:
`VerificationRelationship` is a good place to use the protobuf [oneof](https://developers.google.com/protocol-buffers/docs/proto3#oneof) keyword, because it contains either a string or a `VerificationMethod`.

The current definition is old-fashioned that I implemented in Panacea v1.3. 
The `oneof` is much more intuitive and much easier for client developers.

We don't need to worry about JSON <-> Protobuf, because we've already been relying on `gogoproto.customtype`.
Initially, I tried to remove the `gogoproto.customtype`, but the default behavior of `jsonpb` doesn't handle `oneof` stuff the way we want.
So, let's keep `gogoproto.customtype`. There's no choice in order to obey the W3C spec.


x/did/types/did.go
```diff
@@ -280,7 +280,7 @@ func (doc DIDDocument) validVerificationRelationships(relationships []Verificati
 		if !relationship.hasDedicatedMethod() {
 			// if the relationship isn't a dedicated verification method,
 			// the referenced verification method must be presented in the 'verificationMethod' property.
-			if _, ok := doc.VerificationMethodByID(relationship.VerificationMethodID); !ok {
+			if _, ok := doc.VerificationMethodByID(relationship.GetVerificationMethodId()); !ok {
```
gyuguen, line 283: I think `doc.VerificationMethodByID` should  change to `doc.VerificationMethodById`.
Just like the others. :)

gyuguen (approved):
👍 
I think it's a very nice way.

x/did/types/did.go
youngjoon-lee: Yeah, but Go recommends to use a consistent case for initialisms or acronyms: https://github.com/golang/go/wiki/CodeReviewComments#initialisms
Sadly, the protobuf generates them as camel-case stupidly. But, I think it's good for us to follow Go's rules for our source codes.

gyuguen: Aha. I got it.

PR #171, opened 2021-07-08
fix: Use snake-case for `did.proto` + proto lint fixes

### LINKED ISSUE #170: Use snake-case for field names in `did.proto`
This is a post-issue of the PR #166 .

The `did.proto` still has camel-case field names in order to obey the W3C DID spec.
But, the snake-case is the protobuf standard, and we already rely on the `json_name` tag. So, it wouldn't be a problem.

youngjoon-lee:
Closes #170


proto/panacea/did/v2/query.proto
```diff
@@ -11,8 +11,8 @@ option go_package = "github.com/medibloc/panacea-core/x/did/types";
 
 // Query defines the gRPC querier service.
 service Query {
-	// DIDDocumentWithSeq returns a DID Document with a sequence number.
-	rpc DIDDocumentWithSeq(QueryDIDRequest) returns (QueryDIDResponse) {
+	// DID returns a DID Document with a sequence number.
+	rpc DID(QueryDIDRequest) returns (QueryDIDResponse) {
```
youngjoon-lee, line 15: Due to 
> {"path":"proto/panacea/did/v2/query.proto","start_line":15,"start_column":58,"end_line":15,"end_column":74,"type":"RPC_RESPONSE_STANDARD_NAME","message":"RPC response type \"QueryDIDResponse\" should be named \"DIDDocumentWithSeqResponse\" or \"QueryDIDDocumentWithSeqResponse\"."}

Similar as https://github.com/medibloc/panacea-core/blob/d1fbdd948e963c587407d9c865514e71bbea7e7b/proto/panacea/aol/v2/query.proto#L17

gyuguen (approved):
👍 
Good :)

PR #184, opened 2021-07-12
chore: Remove a legacy statik.go

youngjoon-lee:
The `client/lcd/statik/statik.go` is not necessary anymore. The new swagger-related files are in `client/docs/` that is auto-generated by `make update-swagger-docs`.


.gitattributes
```diff
@@ -0,0 +1 @@
+client/docs/swagger-ui/* linguist-vendored
```
youngjoon-lee, line 1: The auto-generated Javascript files in the `client/docs/swagger-ui/` ruin the language statistic of GitHub.
image

gyuguen (approved):
👍

PR #185, opened 2021-07-12
docs: Add our custom `x/*` modules to Swagger UI

youngjoon-lee:
As described in the [protoc-swagger-gen.sh](https://github.com/medibloc/panacea-core/blob/266cf776e1aea78fc12ba8d1b2192abba7c6b6cd/scripts/protoc-swagger-gen.sh#L27), our custom `x/*` modules should be added to the `client/docs/config.json`.

image


client/docs/swagger-ui/swagger.yaml
```diff
@@ -29515,6 +29515,881 @@ paths:
                       format: byte
       tags:
         - Query
+  '/panacea/aol/v2/owners/{owner_address}/topics':
```
gyuguen, line 29518: Is auto generation?

youngjoon-lee: Yes. Everything in the `client/docs/` are auto-generated by [make proto-swagger-gen](https://github.com/medibloc/panacea-core/blob/83b3983ba35b7be47d296402ed29fdd3fb0a2269/Makefile#L51). That command invokes the [protoc-swagger-gen.sh](https://github.com/medibloc/panacea-core/blob/83b3983ba35b7be47d296402ed29fdd3fb0a2269/scripts/protoc-swagger-gen.sh#L28) script.

That script has two big steps:
1. Generating Swagger docs from all proto files
2. Combining them into one Swagger doc (using `swagger-combine`)

gyuguen: OK. Thank you!

gyuguen (approved):
👍

PR #188, opened 2021-07-13
docs: Release v2.0.0

youngjoon-lee:
Finally, let's freeze codes.

In addition, I've deleted an unnecessary `config.yml` file that was probably added by Starport (for Vue Web UI).


gyuguen (approved):
👍  Let's go

PR #189, opened 2021-07-14
chore: Run docker-publish Github Actions only after the CI is completed

youngjoon-lee:
The `Docker` Github Actions should be started only after the `ci` Actions is completed, because it doesn't make sense to publish a Docker image that doesn't pass the CI.

Reference: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run


gyuguen (approved):
It's good idea. 👍

PR #191, opened 2021-07-14
Revert "chore: Run docker-publish Github Actions only after the CI is completed"

youngjoon-lee:
Reverts medibloc/panacea-core#189


gyuguen (approved):
👍 
T.T

PR #194, opened 2021-07-20
fix: Remove legacy REST API of `x/token`

youngjoon-lee:
The `x/token` contains legacy REST APIs, because I've implemented it before we decided to not provide legacy REST APIs. We provides only gRPC gateway.


gyuguen (approved):
LGTM! 👍

PR #197, opened 2021-07-22
fix: Add major version `v2` to `go.mod`

### LINKED ISSUE #196: Add major version v2 to go.mod
https://blog.golang.org/v2-go-modules

If not, the following error occurs: 
> github.com/medibloc/panacea-core@v2.0.0: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2

youngjoon-lee:
Closes #196


gyuguen (approved):
LGTM. 👍 
If another user use `panacea-core`, an error must have occurred.

PR #202, opened 2021-07-27
fix: Discard `protodep`

### LINKED ISSUE #201: Discard `protodep`
The CI began to throw an error: https://github.com/medibloc/panacea-core/runs/3170573276
```
protodep up --use-https
[INFO] force update = false
[INFO] cleanup cache = false
[INFO] identity file = 
[INFO] use https = true
[INFO] Getting github.com/cosmos/cosmos-sdk 
[INFO] 84c33215658131d87daf3c629e909e12ed9370fa is not a tag, checking out by hash
[INFO] Getting github.com/cosmos/cosmos-sdk 
[INFO] 84c33215658131d87daf3c629e909e12ed9370fa is not a tag, checking out by hash
Error: checkout to 84c33215658131d87daf3c629e909e12ed9370fa is failed: worktree contains unstaged changes
```

I've checked the [cosmos/cosmo-sdk](https://github.com/cosmos/cosmos-sdk) git repo, but I couldn't find any issue. The commit `84c33215658131d87daf3c629e909e12ed9370fa` exists in that git repo.

I suspect that something was broken in the `protodep`. So, let's remove the `protodep`. We can probably use [buf](https://buf.build/) instead of `protodep`, but not that urgent.

youngjoon-lee:
Closes #201 

Instead of `protodep`, let's use `curl` for downloading directories from Github repos.
Also, let's push proto files to our git repo (as cosmos does).


Makefile
```diff
@@ -44,21 +44,25 @@ lint:
 ########################################
 ### Protobuf
 
-proto-gen: proto-update-deps
+proto-gen:
 	@echo "Generating *.pb.go files from *.proto files"
 	$(DOCKER) run --rm -v $(CURDIR):/workspace --workdir /workspace tendermintdev/sdk-proto-gen:v0.2 sh ./scripts/protocgen.sh
```
gyuguen, line 49: I'm not found `CURDIR` variable.
Is it missing? (or if it was intended. 😄 )

gyuguen: I've been curious about it for a long time. HaHa..

gyuguen: I think it is `CURDIR="$(pwd)"`

youngjoon-lee: CURDIR is one of GNU Make keywords. 

https://stackoverflow.com/questions/52437728/bash-what-is-the-difference-between-pwd-and-curdir

gyuguen: Thank you!

gyuguen (approved):
LGTM 👍

PR #220, opened 2021-08-20
docs: Update URL links in README

youngjoon-lee:
I removed URLs of the explorer and the wallet, because the explorer URL has been already posted in each 'network' repo.


gyuguen (approved):
LGTM! 👍

PR #222, opened 2021-09-02
docs: Add Tech Stack

youngjoon-lee:
I added a tech stack document which introduces briefly what Panacea has.


gyuguen (approved):
LGTM! 👍

PR #223, opened 2021-09-03
refactor: Change blocks_per_year in init.go and timeout_commit in Gitbook

youngjoon-lee:
…book


gyuguen (approved):
LGTM! 👍

youngjoon-lee:
Damn.. I just found that one more param should be changed by the governance. 

https://github.com/medibloc/panacea-core/blob/2b09fa09628aa42fd1387daf5a525307102a45b6/cmd/panacead/cmd/init.go#L148

It is not that critical, but let's keep it in mind. We can change it when we execute the SoftwareUpgradeProposal as well.

youngjoon-lee:
> LGTM!

Thanks. I'm gonna merge this PR after our first ParameterChangeProposal has been passed.

PR #239, opened 2021-09-23
docs: add how to use Cosmovisor to gitbook

gyuguen:
I wrote about how to install and apply the cosmovisor.


youngjoon-lee (approved):
Cool :+1:

PR #250, opened 2021-12-21
fix: Modify Dockerfile for Cosmwasm static linking

inchori:
Modify Dockerfile for Cosmwasm static linking.

The binary file for panacea `panacead' works well in Linux OS such as ubuntu or alpine.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,7 @@ RUN sha256sum /lib/libwasmvm_muslc.a | grep 0e62296b9f24cf3a05f8513f99cee536c708
 RUN make clean && LEDGER_ENABLED=true BUILD_TAGS=muslc make build
 
 # Final image
-FROM alpine:3.12
+FROM debian:buster-slim
 
 # Copy over binaries from the build-env
 COPY --from=build-env /src/panacea-core/build/panacead /usr/bin/panacead
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ FROM golang:1.16.5-alpine3.14 AS build-env
 RUN set -eux; apk add --no-cache ca-certificates build-base;
 
 RUN apk add git
-RUN apk add libusb-dev linux-headers
+RUN apk add linux-headers
 
 # Create directory
 RUN mkdir -p /src/panacea-core /src/wasmvm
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -14,19 +15,17 @@ COPY . /src/panacea-core
...
-
 # Get 'libwasmvm.so' from wasmvm
-RUN git clone -b v0.14.0 https://github.com/CosmWasm/wasmvm.git /src/wasmvm
+ADD https://github.com/CosmWasm/wasmvm/releases/download/v0.16.2/libwasmvm_muslc.a /lib/libwasmvm_muslc.a
+RUN sha256sum /lib/libwasmvm_muslc.a | grep 0e62296b9f24cf3a05f8513f99cee536c7087079855ea6ffb4f89b35eccdaa66
+
+RUN make clean && LEDGER_ENABLED=true BUILD_TAGS=muslc make build
 
 # Final image
-FROM debian:buster-slim
```
gyuguen, line 24: I think it's better to use debian:buster-slim here as before.
Because panacea binary is usally used in Ubuntu.

inchori: Thanks for your feedback!

I change final image from alpine to ubuntu in Dockerfile.

Thank you :)

Dockerfile
```diff
@@ -14,19 +15,17 @@ COPY . /src/panacea-core
 # Set working directory for the 'panacea-core' build
 WORKDIR /src/panacea-core
 
-# Install panacea-core
-RUN make clean && make build
-
 # Get 'libwasmvm.so' from wasmvm
-RUN git clone -b v0.14.0 https://github.com/CosmWasm/wasmvm.git /src/wasmvm
+ADD https://github.com/CosmWasm/wasmvm/releases/download/v0.16.2/libwasmvm_muslc.a /lib/libwasmvm_muslc.a
```
youngjoon-lee, line 19: I think it would be safer to use the wasmvm v0.14.0 (instead of v0.16.2).
The panacea-core imports wasmd v0.17.0 which uses wasmvm v0.14.0.

Dockerfile
```diff
@@ -14,19 +15,17 @@ COPY . /src/panacea-core
...
+RUN sha256sum /lib/libwasmvm_muslc.a | grep 0e62296b9f24cf3a05f8513f99cee536c7087079855ea6ffb4f89b35eccdaa66
+
+RUN make clean && LEDGER_ENABLED=true BUILD_TAGS=muslc make build
```
youngjoon-lee, line 22: Actually, we don't need to set the `LEDGER_ENABLED` env var :)

In the `Makefile`, we always enable the `ledger` build tag.
https://github.com/medibloc/panacea-core/blob/e46b2d6f72569e57141a7f1c806a1b82fa33da15/Makefile#L15

But, as you did, the `BUILD_TAGS=muslc` must be set.

inchori: Thanks for your feedback!

I change wasmvm version and checksum data.

Also I removed `LEDGER_ENABLED` env var

Thank you :)

gyuguen (approved):
LGTM 👍

Dockerfile
```diff
@@ -1,9 +1,10 @@
...
+FROM golang:1.16.5-alpine3.14 AS build-env
 
 # Install minimum necessary dependencies,
-ENV PACKAGES make git gcc
-RUN apt-get update -y
-RUN apt-get install -y $PACKAGES
+RUN set -eux; apk add --no-cache ca-certificates build-base;
+
+RUN apk add git
+RUN apk add linux-headers
```
youngjoon-lee, line 7:
```suggestion
RUN apk add libusb-dev linux-headers
```

Ah. As I know, the `libusb-dev` need to be installed, because we always enable the `ledger` tag in the Makefile.

gyuguen: It seems to be build without `livusb-dev`. Is this cause problems when the actual panacea run?

youngjoon-lee: I guess so. I haven't tested it by myself, but referred to https://github.com/CosmWasm/wasmd/blob/9a17505cf8b637942cad78e294aa1377d7f7b2f6/Dockerfile#L11

gyuguen: Aha, I got it. Thank you!

youngjoon-lee (approved):
congrat on your first PR 👍

PR #248, opened 2021-12-02
docs: add a guide to apply the validator's descrption

gyuguen:
I have created a document that can modify the validator's description.

PR #251, opened 2021-12-23
docs: merge split documents for `edit-validator`

youngjoon-lee:
Although we already added a `edit-validator` guide to the `join-the-network.md` (PR https://github.com/medibloc/panacea-core/pull/248), I just found that the `interaction-with-the-network-cli.md` has already had the `edit-validator` guide.

To avoid splitting information, I merged all `edit-validator` guides into the `interaction-with-the-network-cli.md` . Instead, I put only links in the `join-the-network.md`.


gyuguen (approved):
LGTM 👍

PR #252, opened 2022-01-11
feat: Add Create Deal for Data Market Module

inchori:
- Configure proto file and ```app/app.go``` for ```x/market``` module.
- Implement Create Deal feature for ```x/market``` module.
- Complete Test Code for ```keeper/deal_test.go```, ```types/deal_test.go```, and ```market/genesis_test.go```.
- Use Hash function of Cosmos SDK v0.43.0 for making Deal Address.

TODO
- Tx Cmd for Create Deal
- Validation for MsgCreateDeal
- Implement ExportGenesis function for ```x/market``` module.


Dockerfile
```diff
@@ -1,9 +1,10 @@
-FROM golang:1.16.5-buster AS build-env
```
gyuguen, line 1: 이 부분은 이전 개발 소스로 보여지는대 뭔가 새로 추가된게 있나요?

app/app.go
```diff
@@ -248,14 +252,15 @@ type App struct {
...
-	didKeeper   didkeeper.Keeper
-	burnKeeper  burnkeeper.Keeper
-	tokenKeeper tokenkeeper.Keeper
-	wasmKeeper  wasm.Keeper
+	aolKeeper    aolkeeper.Keeper
+	didKeeper    didkeeper.Keeper
+	burnKeeper   burnkeeper.Keeper
+	tokenKeeper  tokenkeeper.Keeper
+	wasmKeeper   wasm.Keeper
+	MarketKeeper marketkeeper.Keeper
```
gyuguen, line 260:
```suggestion
	marketKeeper marketkeeper.Keeper
```
다른 keeper와 마찬가지가지로 소문자로 가면 어떨까요

app/test_helper.go
```diff
@@ -0,0 +1,48 @@
...
+	return app
+}
+
+func FundAccount(bankKeeper bankkeeper.Keeper, ctx sdk.Context, addr sdk.AccAddress, amounts sdk.Coins) error {
+	if err := bankKeeper.MintCoins(ctx, minttypes.ModuleName, amounts); err != nil {
+		return err
+	}
+
+	return bankKeeper.SendCoinsFromModuleToAccount(ctx, minttypes.ModuleName, addr, amounts)
+}
```
gyuguen, line 43: 현재 사용하지는 않는거 같은대 나중에 사용하려고 하나요?

x/market/keeper/deal_test.go
```diff
@@ -0,0 +1,89 @@
...
+		WantDataCount:        10000,
+		TrustedDataValidator: []string{acc2.String()},
+		Owner:                acc1.String(),
+	}
+
+	owner, _ := sdk.AccAddressFromBech32(tempDeal.GetOwner())
+
+	dealId, _ := suite.MarketKeeper.CreateNewDeal(suite.Ctx, owner, tempDeal)
+	suite.Require().Equal(dealId, uint64(2))
+	deal, _ := suite.MarketKeeper.GetDeal(suite.Ctx, 2)
```
gyuguen, line 54:
```suggestion
	deal, _ := suite.MarketKeeper.GetDeal(suite.Ctx, dealId)
```
I think it not good to magicNumber.

x/market/keeper/deal_test.go
```diff
@@ -0,0 +1,89 @@
...
+		Budget:               &sdk.Coin{Denom: "umed", Amount: sdk.NewInt(10000000)},
+		WantDataCount:        10000,
+		TrustedDataValidator: []string{acc2.String()},
+		Owner:                acc1.String(),
+	}
+
+	owner, _ := sdk.AccAddressFromBech32(tempDeal.GetOwner())
+
+	dealId, _ := suite.MarketKeeper.CreateNewDeal(suite.Ctx, owner, tempDeal)
+	suite.Require().Equal(dealId, uint64(2))
```
gyuguen, line 53:
```suggestion
	suite.Require().Equal(dealId, expectedDealId)
```
I think we should let them know that the value 2 is the expected value.

x/market/types/deal_test.go
```diff
@@ -0,0 +1,21 @@
+package types_test
+
+import (
+	"fmt"
+	"github.com/medibloc/panacea-core/v2/x/market/types"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestNewDealAddress(t *testing.T) {
```
gyuguen, line 10: Is this test that the same two deal_ids return the same address?

x/market/keeper/deal.go
```diff
@@ -0,0 +1,115 @@
...
+		authtypes.NewBaseAccountWithAddress(
+			sdk.AccAddress(newDeal.GetDealAddress()),
+		),
+		newDeal.DealAddress),
+	)
+	k.accountKeeper.SetAccount(ctx, acc)
+
+	err = k.bankKeeper.SendCoins(ctx, owner, sdk.AccAddress(newDeal.GetDealAddress()), coins)
+	if err != nil {
+		return 0, nil
```
gyuguen, line 40:
```suggestion
		return 0, err
```
I think the second return will be an err.
And wrapping error message with `The owner's insufficient balance.`

x/market/keeper/deal_test.go
```diff
@@ -0,0 +1,89 @@
...
+	market.InitGenesis(suite.Ctx, suite.MarketKeeper, types.GenesisState{Deals: map[uint64]*types.Deal{
+		testDeal.GetDealId(): &testDeal,
+	}, NextDealNumber: 2})
+
+	suite.MarketKeeper.SetDeal(suite.Ctx, testDeal)
+
+	err := suite.BankKeeper.AddCoins(suite.Ctx, acc1, defaultFunds)
+	if err != nil {
+		panic(err)
+	}
```
gyuguen, line 38: Later, It seems necessary to test that the owner's balance is insufficient. So this code is move to `TestCreateNewDeal` method.

x/market/util/conv/doc.go
```diff
@@ -0,0 +1 @@
+package conv
```
gyuguen, line 1: What is this file?

gyuguen (changes requested):
Thank you for your hard work. I left some revies.

x/market/keeper/deal.go
```diff
@@ -0,0 +1,115 @@
...
+
+	acc = k.accountKeeper.NewAccount(ctx, authtypes.NewModuleAccount(
+		authtypes.NewBaseAccountWithAddress(
+			sdk.AccAddress(newDeal.GetDealAddress()),
+		),
+		newDeal.DealAddress),
+	)
+	k.accountKeeper.SetAccount(ctx, acc)
+
+	err = k.bankKeeper.SendCoins(ctx, owner, sdk.AccAddress(newDeal.GetDealAddress()), coins)
```
gyuguen, line 38:
```suggestion
	err = k.bankKeeper.SendCoins(ctx, owner, sdk.AccAddressFromBech32(newDeal.GetDealAddress()), coins)
```

I think you used to [here](https://github.com/cosmos/cosmos-sdk/blob/v0.44.5/types/address.go#L158).

This will work normally from v0.43.0 and later versions, so let's think about implementing it in a different way.

app/test_helper.go
```diff
@@ -0,0 +1,48 @@
...
+			abci.RequestInitChain{
+				Validators:      []abci.ValidatorUpdate{},
+				ConsensusParams: simapp.DefaultConsensusParams,
+				AppStateBytes:   stateBytes,
+			})
+	}
+	return app
+}
+
+func FundAccount(bankKeeper bankkeeper.Keeper, ctx sdk.Context, addr sdk.AccAddress, amounts sdk.Coins) error {
```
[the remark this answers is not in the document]
inchori: I used this method for helping ``TestCreateNewDeal``, but I replaced this method with ``AddCoins`` methods. I removed the method. Thanks for your feedback.

x/market/util/conv/doc.go
inchori: The ``doc.go`` file is default file in [Cosmos SDK 0.43](https://github.com/cosmos/cosmos-sdk/blob/3f8596c1955e40ef30e4abcd06f2237d132db3a9/internal/conv/doc.go), but I think it will be not use in our project so I will remove this file. Thank you for your feedback.

x/market/client/cli/flags.go
```diff
@@ -0,0 +1,45 @@
...
+	WantDataCount         uint64   `json:"want_data_count"`
+	TrustedDataValidators []string `json:"trusted_data_validators"`
+}
+
+type XCreateDealInputs createDealInputs
+
+type XCreateDealInputsExceptions struct {
+	XCreateDealInputs
+	Other *string
+}
```
gyuguen, line 25: I wonder why X is attached to prefix.
Is this auto-generated?

x/market/keeper/deal.go
```diff
@@ -0,0 +1,120 @@
...
+	newDeal := &types.Deal{
+		DealId:               dealId,
+		DealAddress:          dealAddress.String(),
+		DataSchema:           deal.GetDataSchema(),
+		Budget:               deal.GetBudget(),
+		TrustedDataValidator: deal.GetTrustedDataValidator(),
+		WantDataCount:        deal.GetWantDataCount(),
+		CompleteDataCount:    0,
+		Owner:                deal.GetOwner(),
+		Status:               "ACTIVE",
```
gyuguen, line 63: I think it would be better set `const`

```go
const (
  ACTIVE      = "ACTIVE"
  DEACTIVE    = "DEACTIVE"
  COMPLETED    = "COMPLETED"
)
```

x/market/client/cli/tx.go
```diff
@@ -0,0 +1,71 @@
...
+		return nil, err
+	}
+
+	err = deal.UnmarshalJSON(contents)
+	if err != nil {
+		return nil, err
+	}
+
+	return deal, nil
+}
```
hansol-medi, line 71: How about to move these functions to `x/market/client/cli/txDeal.go` like `did` module?

x/market/client/cli/tx.go
```diff
@@ -0,0 +1,71 @@
...
+
+func parseCreateDealFlags(fs *flag.FlagSet) (*createDealInputs, error) {
+	deal := &createDealInputs{}
+	dealFile, _ := fs.GetString(FlagDealFile)
+
+	if dealFile == "" {
+		return nil, fmt.Errorf("need deal json file using --%s flag", FlagDealFile)
+	}
+
+	contents, err := ioutil.ReadFile(dealFile)
```
hansol-medi, line 60: Does it mean the `CreateDealInput` is received as a file?
If then, we'd better receive it as args, not file.

proto/panacea/market/v2/deal.proto
```diff
@@ -0,0 +1,21 @@
...
+// Deal defines a deal.
+// TODO: Could remove complete_data_count field
+message Deal {
+  uint64 deal_id = 1;
+  string deal_address = 2;
+  repeated string data_schema = 3;
+  cosmos.base.v1beta1.Coin budget = 4;
+  repeated string trusted_data_validator = 5;
+  uint64 want_data_count = 6;
+  uint64 complete_data_count = 7;
```
hansol-medi, line 18: These two variable names are a bit confusing.
How about `target_num_data` and `filled_num_data`?

hansol-medi (commented):
Thank you for your works!
I left some comments.

x/market/keeper/deal.go
inchori: Yes, I agree about this. I'll status field to ```const```.

x/market/types/deal.go
```diff
@@ -0,0 +1,51 @@
...
+	if len(bz) == 0 {
+		return sdkerrors.Wrap(sdkerrors.ErrUnknownAddress, "address cannot be empty")
+	}
+
+	if len(bz) > address.MaxAddrLen {
+		return sdkerrors.Wrapf(sdkerrors.ErrUnknownAddress, "address max length is %d, got %d", address.MaxAddrLen, len(bz))
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
inchori: The Cosmos SDK v0.42 ```AccAddressFromBech32``` is different from Comsos SDK 0.43 ```AccAddressFromBech32```. The max length of address which is constant variable ```AddrLen``` in v0.42 is 20, but the max length of address in v0.43 is 255. When ```dealAddress``` use v0.42 ```AccAddressFromBech32```, it makes error. So I need to use v0.43 ```AccAddressFromBech32``` for make ```dealAddress```. I'll write comment on this function.

x/market/client/cli/tx.go
inchori: Yes, I've heard that when we create deal, we will use json file.

x/market/client/cli/tx.go
inchori: Yes, I agree with your feed back. Thank you!

x/market/client/cli/tx.go
gyuguen: I think it's no proble to receive data in a file. 
Because the [update of did](https://github.com/medibloc/panacea-core/blob/master/x/did/client/cli/txDID.go#L100) and [pool creation of osmosis](https://github.com/osmosis-labs/osmosis/blob/main/x/gamm/client/cli/tx.go#L46) is already received as a json file.

However,some tx uses both json file and argument.

I think it would be good to apply in both ways.(json file and argument) 😀

x/market/keeper/deal.go
```diff
@@ -0,0 +1,120 @@
...
+	k.accountKeeper.SetAccount(ctx, acc)
+
+	err = k.bankKeeper.SendCoins(ctx, owner, dealAddress, coins)
+	if err != nil {
+		return 0, sdkerrors.Wrapf(types.ErrNotEnoughBalance, "The owner's balance is not enough to make deal")
+	}
+	return newDeal.GetDealId(), nil
+}
+
+func (k Keeper) newDeal(dealId uint64, deal types.Deal) (types.Deal, error) {
```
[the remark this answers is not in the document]
gyuguen: Oh that's good. This function is only used within the keeper.

x/market/client/cli/flags.go
inchori: Oh, The prefix X is not auto-generated. I refer to [Osmosis's parse.go](https://github.com/osmosis-labs/osmosis/blob/main/x/gamm/client/cli/parse.go) and I just use X prefix.

proto/panacea/market/v2/deal.proto
youngjoon-lee: Agree. We can have better names. Also, as Gyuguen said, I'm curious what the `complete_data_count` is for. Let me review other codes first.

x/market/client/cli/flags.go
```diff
@@ -0,0 +1,45 @@
...
+const (
+	FlagDealFile = "deal-json-file"
+)
+
+type createDealInputs struct {
+	DataSchema            []string `json:"data_schema"`
+	Budget                string   `json:"budget"`
+	WantDataCount         uint64   `json:"want_data_count"`
+	TrustedDataValidators []string `json:"trusted_data_validators"`
+}
```
gyuguen, line 18: How about using `MsgCreateDeal` in `proto/panacea/market/v2/tx.proto`.

If Budget is a problem, it seems good to change a spec.

And you can refer to the `json_name` option of  `did/did.proto`

x/market/client/cli/txDeal.go
```diff
@@ -0,0 +1,39 @@
...
+
+			return tx.GenerateOrBroadcastTxWithFactory(clientCtx, txf, msg)
+		},
+	}
+
+	cmd.Flags().AddFlagSet(FlagSetCreateDeal())
+	flags.AddTxFlagsToCmd(cmd)
+
+	_ = cmd.MarkFlagRequired(FlagDealFile)
+
```
gyuguen, line 37:
```suggestion
	cmd.Flags().String(FlagDealFile, "" ,"Deal json file path")
	_ = cmd.MarkFlagRequired(FlagDealFile)

	flags.AddTxFlagsToCmd(cmd)

```

How about fix it like this?
I think we don't need a `market/client/cli/flags.go` files.

x/market/client/cli/flags.go
```diff
@@ -0,0 +1,45 @@
...
+	if err := decoder.Decode(&createDealEX); err != nil {
+		return nil
+	}
+
+	*input = createDealInputs(createDealEX.XCreateDealInputs)
+	return nil
+}
+
+func FlagSetCreateDeal() *flag.FlagSet {
+	fs := flag.NewFlagSet("", flag.ContinueOnError)
```
gyuguen, line 41: What does this settings mean?
I wonder what changes without this setting.

x/market/keeper/deal.go
```diff
@@ -0,0 +1,120 @@
...
+		return 0, err
+	}
+
+	var coins sdk.Coins
+	coins = append(coins, *deal.GetBudget())
+
+	acc := k.accountKeeper.GetAccount(ctx, sdk.AccAddress(newDeal.GetDealAddress()))
+	if acc != nil {
+		return 0, sdkerrors.Wrapf(types.ErrDealAlreadyExist, "deal %d already exist", dealId)
+	}
```
youngjoon-lee, line 26: It seems that this logic is only for checking if the deal module account already exist. Did I understand it correctly?

Then, I think we need to use `AccDealAddressFromBech32(newDeal.GetDeadAddress())` (or something similar).

x/market/types/deal.go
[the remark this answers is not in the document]
hansol-medi: I see. The dealAddress is generated by hash function, which is implemented in v0.43.x, with its dealId.
The resulting address may exceed 20 bytes. 
Is this the way osmosis did for create pool?

youngjoon-lee (commented):
Thank you for investigating how to use module accounts. I guessed that the most challenging part is the module account.
I left some comments, but didn't review all codes yet. I will revisit this PR soon.

x/market/types/deal.go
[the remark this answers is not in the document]
inchori: Yes, Osmosis also use this function for making pool address.

x/market/keeper/deal.go
youngjoon-lee: I'm not sure since I'm not familiar with module accounts. Please check if the code above works well as you indended.

x/market/types/deal.go
```diff
@@ -0,0 +1,51 @@
...
+		return nil, err
+	}
+
+	return sdk.AccAddress(bz), nil
+}
+
+func VerifyDealAddressFormat(bz []byte) error {
+	verifier := sdk.GetConfig().GetAddressVerifier()
+	if verifier != nil {
+		return nil
```
youngjoon-lee, line 39:
```suggestion
		return verifier(bz)
```
Shouldn't we call the `verifier` function? as https://github.com/cosmos/cosmos-sdk/blob/33dbf6a795e68c744a2c312b76d41614b3563c01/types/address.go#L143

x/market/types/deal.go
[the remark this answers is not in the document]
youngjoon-lee: I just understood why this logic needs to be copied from cosmos-sdk. As we discussed, this workaround is what Osmosis did: https://github.com/osmosis-labs/osmosis/pull/169. Thanks for digging into this issue.

Please mark these functions or files clearly (e.g. `v044_temp`), so that we can remove them after upgrading cosmos-sdk to v0.44.

proto/panacea/market/v2/deal.proto
youngjoon-lee: It seems that `complete_data_count` is not used anywhere. If there is no special reason to have this field, we can remove it.

If you think this field should be preserved for future use, please let me know.

x/market/types/deal.go
inchori: I agree with your review. I'll replace ```nil``` with ```verfiier(bz)```.

x/market/client/cli/flags.go
youngjoon-lee: IMHO, we don't need `XCreateDealInputsExceptions`. I just looked at Osmosis source codes, and it seems that the `Other` field is for copying properties of an other existing pool in order to create a new pool. In our case, we aren't gonna need this `Other` option.

x/market/client/cli/flags.go
youngjoon-lee: I guess this is the general pattern to create an unnamed flag set using https://github.com/spf13/pflag (if we don't need any name for the flag set).

youngjoon-lee: Anyway, in my opinion, we don't need to follow the Osmosis' way. 

For consistency, I'd suggest to define flags one by one using: `cmd.Flags().String(...)`: https://github.com/medibloc/panacea-core/blob/388585acad2058b5de0c98ef615f8ed41eb286b6/x/aol/client/cli/txTopic.go#L42

When upgrading cosmos-sdk to v0.44, we can change all modules to use `FlagSet`.

x/market/client/cli/flags.go
youngjoon-lee: IMHO, it's ok to have this separated `createDealInputs` struct, because this struct is slightly different from `MsgCreateDeal`. I don't think it's good to modify the `tx.proto` for adding some json_name tags.

Also, `MsgCreateDeal` can be more complex in the future, as [MsgCreatePool](https://github.com/osmosis-labs/osmosis/blob/ba4dc51cc8e5c45e96fa14fd4b5ca9321ca92a78/x/gamm/types/tx.pb.go#L34) in the Osmosis.

x/market/client/cli/flags.go
gyuguen: Oh, This create a flagset without any name. 
I understand. 😀

youngjoon-lee:
Reviewed all files. Please ping me after resolving (or rejecting) feedbacks above. Then, I'll revisit.

x/market/client/cli/flags.go
gyuguen: Aha, It's ok.
Let's leave this struct as it is.

x/market/client/cli/flags.go
inchori: I agree with that we need to follow Osmosis' way and not yet to use ```FlagSet```. Thanks for your feedback.

x/market/client/cli/txDeal.go
inchori: I think this review is related to below issue. So I will resolve this conversation. Thank you for your feedback 😀

x/market/keeper/deal.go
inchori: Yes, This is my mistake. I didn't change this code yet. I'll change it to ```AccDealAddressFromBech32(newDeal.GetDealAddress())```. Thank you for your feedback.

x/market/client/cli/flags.go
```diff
@@ -0,0 +1,32 @@
...
+	dec := json.NewDecoder(bytes.NewReader(data))
+	dec.DisallowUnknownFields()
+
+	if err := dec.Decode(&createDeal); err != nil {
+		return nil
+	}
+
+	*input = createDealInputs(createDeal)
+	return nil
+}
```
gyuguen, line 32: How about moving all of thes sources to `x/market/client/cli/txDeal.go`?

These don't seem to be related to `flag`

inchori: Thank your for your review. I think it's good idea but about rename this file to ```parse.go```?

x/market/keeper/deal_test.go
```diff
@@ -0,0 +1,113 @@
...
+	return types.Deal{
+		DealId:                1,
+		DealAddress:           types.NewDealAddress(1).String(),
+		DataSchema:            []string{acc1.String()},
+		Budget:                &sdk.Coin{Denom: "umed", Amount: sdk.NewInt(1000000000)},
+		TargetNumData:         10000,
+		FilledNumData:         0,
+		TrustedDataValidators: []string{acc2.String()},
+		Owner:                 acc1.String(),
+		Status:                "ACTIVE",
```
gyuguen, line 111: Please change to const.

```suggestion
		Status:                ACTIVE,
```

proto/panacea/market/v2/deal.proto
```diff
@@ -0,0 +1,21 @@
...
+// Deal defines a deal.
+// TODO: Could remove complete_data_count field
+message Deal {
+  uint64 deal_id = 1;
+  string deal_address = 2;
+  repeated string data_schema = 3;
+  cosmos.base.v1beta1.Coin budget = 4;
+  repeated string trusted_data_validators = 5;
+  uint64 target_num_data = 6;
+  uint64 filled_num_data = 7;
```
youngjoon-lee, line 18: Aha. Now I understand what this field means. I guess you're gonna update this value whenever you get a SellData tx. 
Then, I think the `cur num data` and `max num data` would be clearer. wdyt?
it's little awkward that the deal is 'filled'. 

btw. i'm not sure yet how we'll improve this deal concept in near future. But, from my gut feeling, it would be enough to have only budget and pricePerData. Then, the deal becomes deactivated when the balance of the deal account runs out.
But, don't think too much of this now. let's see when we implement SellData tx

inchori: I agree with your review that rename the field to ```cur_num_data``` and ```max_num_data```.  Also we need to think this when we implement SellData feature. And It's good idea that we just have ```budget``` field and ```pricePerData``` field! Thank you for your review.

hansol-medi: Good! `max_num_data` and `cur_num_data` are clearer.

gyuguen (approved):
LGTM 👍 
I checked all the reviews
Please check youngjoon-lee commants.

x/market/types/deal.go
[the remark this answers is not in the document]
inchori: This review is quite important topic for create-deal feature. So I will not resolve this conversation.

youngjoon-lee (approved):
Thank you! WAGMI!!

hansol-medi (approved):
👍👍👍

PR #254, opened 2022-01-17
fix: Add validation for MsgCreateDeal

inchori:
- Rename ```target_num_data``` and ```filled_num_data``` to ```max_num_data``` and ```cur_num_data```.
- Validation for MsgCreateDeal in ```tx.proto``` file and ```ValidateBasic()``` function in ```message_deal.go``` file.
- Remove ```NextDealNumber``` field in ```ExportGenesis``` function.

TODO
- Implement GetDealsList for ```ExportGenesis```.
- Test Code for client command-line ```txDeal.go```.


x/market/genesis.go
```diff
@@ -22,12 +22,11 @@ func ExportGenesis(ctx sdk.Context, k keeper.Keeper) *types.GenesisState {
...
 	dealsMap := make(map[uint64]*types.Deal)
 
-	//TODO: Implement GetDeals and Mapping to dealsMap
+	//TODO: Implement GetDealsList and Mapping to dealsMap
 	//for _, deal := k.GetDeals(ctx) {
 	//}
 
 	return &types.GenesisState{
-		Deals:          dealsMap,
-		NextDealNumber: k.GetNextDealNumberAndIncrement(ctx),
```
youngjoon-lee, line 31: I think we shouldn't remove this line, because the [types.GenesisState](https://github.com/medibloc/panacea-core/blob/38c4e44b07b8b476168ebd739fd361c153613725/x/market/types/genesis.pb.go#L29) still has the `NextDealNumber` field. Also, the [InitGenesis()](https://github.com/medibloc/panacea-core/blob/392b35dae1c4c7a9ba07653bf7cca223d775ef71/x/market/genesis.go#L12) uses that field.

IMHO, we can replace `NextDealNumber: k.GetNextDealNumberAndIncrement(ctx),` with `NextDealNumber: k.GetNextDealNumber(ctx),`

https://github.com/medibloc/panacea-core/pull/252#discussion_r783588591

x/market/genesis.go
```diff
@@ -22,12 +22,11 @@ func ExportGenesis(ctx sdk.Context, k keeper.Keeper) *types.GenesisState {
...
+		Deals: dealsMap,
```
youngjoon-lee, line 30: I think we shouldn't remove this line, because the [types.GenesisState](https://github.com/medibloc/panacea-core/blob/38c4e44b07b8b476168ebd739fd361c153613725/x/market/types/genesis.pb.go#L29) still has the `NextDealNumber` field. Also, the [InitGenesis()](https://github.com/medibloc/panacea-core/blob/392b35dae1c4c7a9ba07653bf7cca223d775ef71/x/market/genesis.go#L12) uses that field.

IMHO, we can replace `NextDealNumber: k.GetNextDealNumberAndIncrement(ctx),` with `NextDealNumber: k.GetNextDealNumber(ctx),`

https://github.com/medibloc/panacea-core/pull/252#discussion_r783588591

gyuguen (approved):
LGTM 👍

x/market/genesis.go
inchori: Thanks for your review! I agree with that replacing ```NextDealNumber: k.GetNextDealNumberAndIncrement(ctx),``` with ```NextDealNumber: k.GetNextDealNumber(ctx),```.

youngjoon-lee (approved):
LGMT!

PR #255, opened 2022-01-18
feat: Add Sell Data for Data Market Module

inchori:
What I done
- Configure ```tx.proto``` file and ```msg_server_deal.go```.
- Write ```SellOwnData``` function in ```keeper/deal.go``` .
- Write Test Code for ```SellOwnData``` function.

TODO
- Verification for DataValidator Signature.
- Replace ```max_num_data``` field and ```cur_num_data``` field with ```pricePerData```.
- Command-Line for ```SellData``` for Tx.

This branch will be base branch for ```SellData``` feature.


x/market/keeper/deal.go
```diff
@@ -125,3 +125,76 @@ func (k Keeper) SetDeal(ctx sdk.Context, deal types.Deal) {
 	bz := k.cdc.MustMarshalBinaryLengthPrefixed(&deal)
 	store.Set(dealKey, bz)
 }
+
+func (k Keeper) SellOwnData(ctx sdk.Context, seller sdk.AccAddress, cert types.DataValidationCertificate) (sdk.Coin, error) {
```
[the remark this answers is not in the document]
gyuguen: I think the same.
It would be better to check if there is the same datahash in this deal.

x/market/types/keys.go
```diff
@@ -25,8 +25,32 @@ var (
...
+// combineKeys function defines combines deal_id with data_hash.
+func combineKeys(keys ...[]byte) []byte {
+	combined := []byte{}
+	for i, key := range keys {
+		combined = append(combined, key...)
+		if i < len(keys)-1 { // not last item
+			combined = append(combined, KeyIndexSeparator...)
+		}
+	}
+	return combined
```
gyuguen, line 55:
```suggestion
	return bytes.Join(keys, KeyIndexSeparator)
```

I think you can change it to this simple. 😀

x/market/types/message_deal.go
```diff
@@ -71,3 +71,43 @@ func (msg *MsgCreateDeal) GetSigners() []sdk.AccAddress {
...
+func (msg *MsgSellData) Route() string {
+	return RouterKey
+}
+
+func (msg *MsgSellData) Type() string {
+	return "SellData"
+}
+
+// ValidateBasic is validation for MsgCreateDeal.
+func (msg *MsgSellData) ValidateBasic() error {
```
gyuguen, line 93: Why don't we add the necessary field verification?

gyuguen (commented):
Thank you. I put a few comments

x/market/keeper/deal.go
```diff
@@ -125,3 +125,76 @@ func (k Keeper) SetDeal(ctx sdk.Context, deal types.Deal) {
...
+		}
+	}
+	if !flag {
+		return sdk.Coin{}, fmt.Errorf("data validator is invalid address")
+	}
+
+	// Is deal's balance insufficient
+	dealBalance := k.bankKeeper.GetBalance(ctx, dealAddress, "umed")
+	sellerBalance := k.bankKeeper.GetBalance(ctx, seller, "umed")
+	if dealBalance.IsLT(sellerBalance) {
```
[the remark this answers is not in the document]
inchori: Yes, you're right. I need to fix this. Thanks for your feedback!

x/market/keeper/deal.go
```diff
@@ -125,3 +125,76 @@ func (k Keeper) SetDeal(ctx sdk.Context, deal types.Deal) {
...
+	//TODO: Change the field(max_num_data) to price_per_data
+	totalAmount := findDeal.GetBudget().Amount.Uint64()
+	countOfData := findDeal.GetMaxNumData()
+
+	amountOfReward := totalAmount / countOfData
+
+	pricePerData := sdk.NewCoin("umed", sdk.NewIntFromUint64(amountOfReward))
+
+	var coins sdk.Coins
+	coins = append(coins, pricePerData)
```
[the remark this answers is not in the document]
inchori: I think it's seems to be clear to read. Thanks for your feedback!

x/market/keeper/deal.go
```diff
@@ -125,3 +125,76 @@ func (k Keeper) SetDeal(ctx sdk.Context, deal types.Deal) {
...
+	flag := false
+	for _, v := range trustedValidators {
+		if validator == v {
+			flag = true
+			break
+		}
+	}
+	if !flag {
+		return sdk.Coin{}, fmt.Errorf("data validator is invalid address")
+	}
```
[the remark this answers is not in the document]
inchori: I agree with your review. It seems to be improve the code's readability. Thank you!!

x/market/types/message_deal.go
inchori: Sorry for that. I will add ```ValidateBasic``` for verifying field. I'll put the ```TODO``` comment on this function.

x/market/keeper/deal.go
```diff
@@ -125,3 +125,76 @@ func (k Keeper) SetDeal(ctx sdk.Context, deal types.Deal) {
...
+	dataHash := cert.UnsignedCert.DataHash
+	dealKeyData := types.GetKeyPrefixData(dealId, dataHash)
+	storedData := k.cdc.MustMarshalBinaryLengthPrefixed(cert.GetUnsignedCert())
+	store.Set(dealKeyData, storedData)
+}
+
+func (k Keeper) SetCurNumData(deal types.Deal) {
+	curNumData := deal.GetCurNumData() + 1
+	deal.CurNumData = curNumData
+}
```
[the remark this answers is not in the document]
inchori: Yes, I thinks so too. It's unnecessary.

x/market/keeper/deal.go
```diff
@@ -125,3 +125,76 @@ func (k Keeper) SetDeal(ctx sdk.Context, deal types.Deal) {
...
+	if err != nil {
+		return sdk.Coin{}, sdkerrors.Wrapf(types.ErrNotEnoughBalance, "The deal's balance is not enough to make deal")
+	}
+
+	k.SetCurNumData(findDeal)
+	k.SetData(ctx, findDeal.GetDealId(), cert)
+	return pricePerData, nil
+}
+
+func (k Keeper) SetData(ctx sdk.Context, dealId uint64, cert types.DataValidationCertificate) {
```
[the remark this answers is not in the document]
inchori: Thank you!! I think you're right. We don't store the ```data``` itself, we just store ```DataCert```!. I'll rename it.

x/market/types/message_deal.go
gyuguen: Please take out the comment that you're sorry. 😀
You don't have to be sorry :)

x/market/types/message_deal.go
```diff
@@ -71,3 +71,43 @@ func (msg *MsgCreateDeal) GetSigners() []sdk.AccAddress {
...
+	bz := ModuleCdc.MustMarshalJSON(msg)
+	return sdk.MustSortJSON(bz)
+}
+
+func (msg *MsgSellData) GetSigners() []sdk.AccAddress {
+	creator, err := sdk.AccAddressFromBech32(msg.Seller)
+	if err != nil {
+		panic(err)
+	}
+	return []sdk.AccAddress{creator}
```
gyuguen, line 112:
```suggestion
	signer, err := sdk.AccAddressFromBech32(msg.Cert.UnsignedCert.RequesterAddress
	if err != nil {
		panic(err)
	}
	return []sdk.AccAddress{signer}
```
I think signer and requesterAddress will be the same.
So if you bring a signer, how about using it like this?

types/testsuite/suite.go
```diff
@@ -151,6 +152,7 @@ func newTestCodec() params.EncodingConfig {
 	banktypes.RegisterInterfaces(interfaceRegistry)
 	didtypes.RegisterInterfaces(interfaceRegistry)
 	markettypes.RegisterInterfaces(interfaceRegistry)
+	cryptocodec.RegisterInterfaces(interfaceRegistry)
```
gyuguen, line 155: Pubkey must be registered as an interface.

x/market/keeper/deal_test.go
```diff
@@ -137,17 +137,22 @@ func (suite *dealTestSuite) TestSellOwnData() {
...
-//}
+func (suite *dealTestSuite) TestVerify() {
+	cert := makeTestCert()
+
+	validatorAddr, err := sdk.AccAddressFromBech32(cert.UnsignedCert.GetDataValidatorAddress())
+	suite.Require().NoError(err)
+	suite.Require().Equal(newAddr, validatorAddr)
+
+	account := suite.AccountKeeper.NewAccountWithAddress(suite.Ctx, validatorAddr)
+	err = account.SetPubKey(privKey.PubKey())
```
gyuguen, line 148: Some of the codes are validated with DataValidator's Pubkey.

gyuguen: Also, secp256k1 should use cosmos-sdk, not tendermint.
Because `account.SetPubKey` received only cosmos-sdk's secp256k1

youngjoon-lee (commented):
Nicely splitted PR and good comments from everyone.

x/market/keeper/deal_test.go
inchori: Thank you for solving issue #259! Register ```cryptocodec``` interface in ```suite.go``` solves this issue as @hansol-medi said.

x/market/keeper/deal.go
```diff
@@ -125,3 +125,76 @@ func (k Keeper) SetDeal(ctx sdk.Context, deal types.Deal) {
...
+	store := ctx.KVStore(k.storeKey)
+	dataHash := cert.UnsignedCert.DataHash
+	dealKeyData := types.GetKeyPrefixData(dealId, dataHash)
+	storedData := k.cdc.MustMarshalBinaryLengthPrefixed(cert.GetUnsignedCert())
```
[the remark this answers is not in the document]
inchori: Thank you for your feedback! I agree with your review. I'll change it.

x/market/types/message_deal.go
inchori: I agree with your review. The ```seller``` is duplicated with ```RequesterAddress```. Thank you! I'll change it.

inchori: In addition, the ```GetSigners``` method is auto-generated, but I think that it would be better to change.

inchori:
I resolve some reviews.
If #260 approved, I will merge #260 to this branch.

TODO
- ```ValidatBasic()``` for ```MsgSellData```.
- Test ```SetCurNumData``` function in ```keeper/deal.go```.

x/market/types/keys.go
```diff
@@ -25,8 +28,25 @@ var (
 
 	// KeyPrefixDeals defines key to store deals
 	KeyPrefixDeals = []byte{0x02}
+
+	// KeyPrefixDataCertificateStore defines key to store data
```
hansol-medi, line 32: Please add 'certificate' at the end of this comment as I said before. Thanks!

x/market/types/message_deal.go
```diff
@@ -71,3 +71,44 @@ func (msg *MsgCreateDeal) GetSigners() []sdk.AccAddress {
...
+// ValidateBasic is validation for MsgSellData.
+// TODO: Basic Validation for MsgSellData.
+func (msg *MsgSellData) ValidateBasic() error {
+	_, err := sdk.AccAddressFromBech32(msg.Seller)
+	if err != nil {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid creator address (%s)", err)
+	}
+
+	return nil
+}
```
hansol-medi, line 101: Shouldn't we check here if the seller and the requester (in data certificate) are the same?

x/market/types/message_deal.go
```diff
@@ -71,3 +71,44 @@ func (msg *MsgCreateDeal) GetSigners() []sdk.AccAddress {
...
+
+func (msg *MsgSellData) GetSignBytes() []byte {
+	bz := ModuleCdc.MustMarshalJSON(msg)
+	return sdk.MustSortJSON(bz)
+}
+
+func (msg *MsgSellData) GetSigners() []sdk.AccAddress {
+	signer, err := sdk.AccAddressFromBech32(msg.Cert.UnsignedCert.GetRequesterAddress())
```
hansol-medi, line 109: There is nothing wrong here, but I think the signer should be obtained from `msg.Seller`.
Strictly speaking, the `requesterAddress` is set by data validator in process of data validation, so it is not the signer of this transaction.
To sell data successfully, the seller and requesterAddress (in data certificate) should be the same, but I think this check should be done in validation of tx.

If I miss something in this review (including the next [pr](https://github.com/medibloc/panacea-core/pull/260), please let me know. Thank you!

x/market/keeper/deal_test.go
```diff
@@ -104,6 +106,36 @@ func (suite *dealTestSuite) TestGetBalanceOfDeal() {
...
+//TODO: Get Num of DataCertificate stored in KV store.
+func (suite *dealTestSuite) TestSellOwnData() {
+	err := suite.BankKeeper.AddCoins(suite.Ctx, acc1, defaultFunds)
+	suite.Require().NoError(err)
+
+	err = suite.BankKeeper.AddCoins(suite.Ctx, acc3, zeroFunds)
+	suite.Require().NoError(err)
+
+	tempDeal := types.Deal{
+		DataSchema:            []string{acc1.String()},
```
hansol-medi, line 118: For future works, we have to use valid data schema for this.
Also, there need some fail-cases for selling data (e.g. invalid schema, inactive deal, and so on).
If then, let's make some TODO for this! I want @youngjoon-lee and @gyuguen to comment for this!

hansol-medi (commented):
Thanks @inchori for your works. I left a few comments more for this pr. If you have any question of thoughts, feel free to mention me!

x/market/keeper/deal.go
```diff
@@ -125,3 +125,87 @@ func (k Keeper) SetDeal(ctx sdk.Context, deal types.Deal) {
...
+	if err != nil {
+		return sdk.Coin{}, sdkerrors.Wrapf(types.ErrNotEnoughBalance, "The deal's balance is not enough to make deal")
+	}
+
+	SetCurNumData(findDeal)
+	k.SetDataCertificate(ctx, findDeal.GetDealId(), cert)
+	return pricePerData, nil
+}
+
+func (k Keeper) isDataCertDuplicate(ctx sdk.Context, cert types.DataValidationCertificate) (sdk.Coin, error) {
```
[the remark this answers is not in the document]
inchori: Thanks for your feedback! As I know when we sell data, the message response has ```reward``` response. So I return ```sdk.Coin``` value type in this function.

x/market/types/message_deal.go
inchori: Yes I didn't fix yet. I will fix and push it soon!

x/market/keeper/deal.go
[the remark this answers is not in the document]
hansol-medi: Right! `SellOwnData` function returns `reward`, but it doesn't mean the `isDataCertDuplicate` func should return `sdk.Coin` too. It just check the data certificate is duplicated or not. As you can see, this function returns `sdk.Coin{}` regardless of the result.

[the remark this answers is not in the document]
inchori: Ah yes, You're right. 😂

x/market/keeper/deal.go
```diff
@@ -125,3 +125,87 @@ func (k Keeper) SetDeal(ctx sdk.Context, deal types.Deal) {
...
+	store := ctx.KVStore(k.storeKey)
+	dataCertKey := types.CombineKeys(sdk.Uint64ToBigEndian(cert.UnsignedCert.GetDealId()), []byte(cert.UnsignedCert.GetDataHash()))
```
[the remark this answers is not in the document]
inchori: Thank you I fix it to ```GetKeyPrefixDataCertificate```.

x/market/keeper/deal_test.go
inchori: Yes, I agree with your review. I didn't write test code for fail cases yet. I will also write test code for fail cases!. Thanks you!

x/market/types/message_deal.go
youngjoon-lee: Good point. I also think we should use `msg.Seller`.

inchori: Okay Thank you! I'll change it too.

inchori:
@youngjoon-lee @hansol-medi Thanks for yours feedback! Could you review or approve #260 this branch first please? I will merge it to this branch and keep working!

x/market/keeper/deal_test.go
youngjoon-lee: i agree that we need to have tests for various corner cases. But, if you feel this PR is getting bloated, you can just leave TODO comments and do it as a next PR.

x/market/keeper/deal.go
```diff
@@ -125,3 +125,87 @@ func (k Keeper) SetDeal(ctx sdk.Context, deal types.Deal) {
...
+	if err != nil {
+		return coin, err
+	}
+
+	findDeal, err := k.GetDeal(ctx, cert.UnsignedCert.GetDealId())
+	if err != nil {
+		return sdk.Coin{}, err
+	}
+
+	if findDeal.GetStatus() != ACTIVE {
```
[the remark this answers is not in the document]
inchori: Yes, I need to add when the deal is ```COMPLETED``` too. Thank you!

youngjoon-lee (approved):
Great job. I think this PR can be merged as long as some comments above are resolved.

inchori:
TODO
- Failure cases test code for not only ```CreateDeal```, but also ```SellData``` in next PR.
- Change field name ```cur_num_data``` and ```max_num_data``` to ```price_per_data``` in ```deactivate-deal``` feature.

x/market/keeper/deal.go
```diff
@@ -125,3 +125,108 @@ func (k Keeper) SetDeal(ctx sdk.Context, deal types.Deal) {
...
+	}
+
+	findDeal, err := k.GetDeal(ctx, cert.UnsignedCert.GetDealId())
+	if err != nil {
+		return sdk.Coin{}, err
+	}
+
+	if findDeal.GetStatus() == INACTIVE || findDeal.GetStatus() == COMPLETED {
+		return sdk.Coin{}, sdkerrors.Wrap(err, "deal is not activated")
+	}
```
hansol-medi, line 142: What I meant was to separate these two cases.
I thought it would be better to response fail message which informs the current state of deal, `INACTIVE` or `COMPLETED`
If you think no need to distinguish these two cases, the previous code you wrote `if findDeal.GetStatus() != ACTIVE` would be enough.

x/market/types/message_deal.go
```diff
@@ -71,3 +71,85 @@ func (msg *MsgCreateDeal) GetSigners() []sdk.AccAddress {
...
+	if unsignedCert.RequesterAddress == "" {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "empty requester address")
+	}
+	_, err = sdk.AccAddressFromBech32(unsignedCert.RequesterAddress)
+	if err != nil {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid requester address (%s)", err)
+	}
+
+	return nil
+}
```
hansol-medi, line 142: We need to check if the seller and requester address (in unsigned certificate) are the same.
It may be missed and I think this check would be done here `ValidateBasic`. wdyt?

hansol-medi (commented):
Many reviews have been resolved. 
I think you might miss few points and I commented.

x/market/keeper/deal.go
inchori: Ah I understand it. I will change the errors message.

x/market/types/message_deal.go
inchori: Yes, you're right. I need to add it.

x/market/keeper/deal.go
```diff
@@ -125,3 +125,108 @@ func (k Keeper) SetDeal(ctx sdk.Context, deal types.Deal) {
...
+func SetCurNumData(deal *types.Deal) {
+	curNumData := deal.GetCurNumData() + 1
+	deal.CurNumData = curNumData
+}
+
+func (k Keeper) Verify(ctx sdk.Context, validatorAddr sdk.AccAddress, cert types.DataValidationCertificate) (bool, error) {
+	validatorAcc := k.accountKeeper.GetAccount(ctx, validatorAddr)
+	if validatorAcc == nil {
+		return false, sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid validator address")
+	}
```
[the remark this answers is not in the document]
gyuguen: Yes. To be precise, if there isw no balance, the account itself does not exist on-chain, and no publicKey is created unless tx is performed.
This code will only response with an error if the account does not exist.
In my option, it would be good if the following code was added.
```go
if validatorAcc.GetPubKey() == nil {
  return false, sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "the publicKey does not exist in the validator account")
}
```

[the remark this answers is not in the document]
inchori: Yes, I agree with gyuguen's review. I think I need to add the validation code for the public key existed.

x/market/keeper/deal_test.go
gyuguen: I agree too. 
Test code is important not only for successful test cases, but also for failing cases.
I told @inchori that it would be good to test the success cases and some important failure cases first.

x/market/keeper/deal.go
```diff
@@ -125,3 +125,108 @@ func (k Keeper) SetDeal(ctx sdk.Context, deal types.Deal) {
...
+	err = k.isTrustedValidator(cert, findDeal)
+	if err != nil {
+		return sdk.Coin{}, err
+	}
+
+	//TODO: Change the field(max_num_data) to price_per_data
+	totalAmount := findDeal.GetBudget().Amount.Uint64()
+	countOfData := findDeal.GetMaxNumData()
+
+	pricePerData := sdk.NewCoin("umed", sdk.NewIntFromUint64(totalAmount/countOfData))
```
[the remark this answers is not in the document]
inchori: Oh I didn't knew that there has constant value for "umed". I will change the other codes too.

proto/panacea/market/v2/tx.proto
```diff
@@ -27,3 +29,29 @@ message MsgCreateDeal {
...
+}
+
+// UnsignedDataValidationCertificate defines unsigned data validation certificate.
+message UnsignedDataValidationCertificate {
+  uint64 deal_id = 1;
+  string data_hash = 2;
+  string encrypted_data_url = 3;
+  string data_validator_address = 4;
+  string requester_address = 5;
+}
```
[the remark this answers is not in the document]
inchori: I have a question. If ```signature``` is Base64 encoded, is the data type of ```signature``` will be ```string```?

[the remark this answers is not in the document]
gyuguen: Yes, that's right. Base64 encoding increases the length, but replaces the byte array with a human-readable string.
You can take a look at an example [here](https://github.com/medibloc/panacea-data-market-validator/pull/21/files#diff-c70ba4f8c966bc19cae21720247c4a3e51f8b1d1fa6fae21136dbc981ced8debR11).

note: https://en.wikipedia.org/wiki/Base64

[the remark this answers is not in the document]
inchori: In my opinion, I think that change ```[]byte``` to ```string``` would be better!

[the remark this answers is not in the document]
hansol-medi: It seems that type of signature is defined in bytes in other `.proto` file of cosmos-sdk.
For consistence and ease to verify signature, I think bytes would be better.
In case of `data_hash` and `encrypted_data_url`, I think we can use either `string` or `bytes` for our convenience.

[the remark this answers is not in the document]
inchori: I will apply the conclusion of this discussion in next PR after resolve other reviews.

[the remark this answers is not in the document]
gyuguen: @inchori @hansol-medi 
Thank you for your comments. What I am concerned about is that byte arrays are not easy to express in json format.
When the seller has refined data in json format, if the signature is in the byte array format, it will probably have a long array of numbers.
```json
{
    "signature":[10, 9, 11, 20, 18]
}
```
If we base64 encode it, it will show up as a human-readable string
```json
{
    "signature": "asdfeewqcv..."
}
```

In general, in json, I don't use a byte array, but I think that a base64 encoded string isused.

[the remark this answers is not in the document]
gyuguen: Ah, I'm thinking again, proto is not a place to specify json. 😅
However, the client needs to think differently.(Where the json file is handled directly)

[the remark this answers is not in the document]
inchori: So if the ```signature``` field is Base64 encoded, the data type of ```signature``` will be changed ```parse.go``` file .

hansol-medi (approved):
LGTM!

x/market/client/cli/txDeal.go
```diff
@@ -119,17 +120,32 @@ func NewSellDataMsg(clientCtx client.Context, txf tx.Factory, fs *flag.FlagSet)
 		return txf, nil, fmt.Errorf("failed to parse data certificate file: %w", err)
 	}
 
+	encryptedDataUrlBytes, err := base64.URLEncoding.DecodeString(sellData.Cert.UnsignedCert.EncryptedDataUrlBase64)
```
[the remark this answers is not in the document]
gyuguen: Oh, I missed this part in the previous PR.
I think `base64.UrlEncoding` should be written in the URL or file name. 
The field `EncryptedDataUrlBase64`  has a URL in its name, but it is actually encrypted String data, so it seems better to use `base64.StdEncoding`, astandard Base64 encoding.

youngjoon-lee (approved):
Good!

x/market/client/cli/txDeal.go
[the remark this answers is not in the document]
inchori: I didn't know there have a difference between ```StdEncoding``` and ```UrlEncoding```. Thank you!

app/config.go
```diff
@@ -0,0 +1,61 @@
+package app
```
gyuguen, line 1: I think it would be better to put this commit in the next pr.
This is because new features have already been approvaed by other reviewers before they are uploaded.

PR #261, opened 2022-01-27
fix: Redefine connection method and data type of some fields

inchori:
Redefine connection method and data type of some fields as @gyuguen said.

**JSON**
- Change the fields ```data_hash```, ```encrypted_data_url```, and ```signature```in ```UnsignedDataValidationCertificate``` to ```string``` in ```parse.go```.
- Add the suffix ```_base64``` to the fields.

**Proto**
- Change the fields ```data_hash```, ```encrypted_data_url```, in ```UnsignedDataValidationCertificate``` to ```string``` in ```tx.proto```.


hansol-medi (approved):
👍

gyuguen (approved):
LGTM 👍

PR #267, opened 2022-02-09
fix: Change parsing proto to json

inchori:
First of all, I just changed parsing method only in ```createDeal``` using ```jsonpb```.
If my work is correct, I will change method in ```sellData```.

### DIFF
--- a/x/market/client/cli/txDeal.go
+++ b/x/market/client/cli/txDeal.go
@@ -2,13 +2,13 @@ package cli
 
 import (
 	"bytes"
-	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"github.com/cosmos/cosmos-sdk/client"
 	"github.com/cosmos/cosmos-sdk/client/flags"
 	"github.com/cosmos/cosmos-sdk/client/tx"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/gogo/protobuf/jsonpb"
 	"github.com/medibloc/panacea-core/v2/x/market/types"
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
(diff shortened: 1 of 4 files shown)


x/market/client/cli/txDeal.go
```diff
@@ -100,48 +106,23 @@ func CmdDeactivateDeal() *cobra.Command {
...
-	}
-
-	msg := types.NewMsgCreateDeal(
-		deal.DataSchema,
-		&budget,
-		deal.MaxNumData,
-		deal.TrustedDataValidators,
-		clientCtx.GetFromAddress().String(),
-	)
+func NewBuildCreateDealMsg(txf tx.Factory, fs *flag.FlagSet) (tx.Factory, types.MsgCreateDeal, error) {
```
gyuguen, line 109: Is there a reason to use it like this because it seems like there is no need to return `tx.Factory`?

inchori: Oh yes, in this function the ```tx.Factory``` seems to be not needed. 
By the way, there have some problem in unmarshaling the create deal json file. So I drafted this PR.

x/market/client/cli/txDeal_test.go
```diff
@@ -1,26 +1,16 @@
...
+func TestParseSellDataFlags(t *testing.T) {
+	flags := CmdSellData().Flags()
+	err := flags.Set(DataVerificationCertificateFile, "./testdata/data_certificate_file.json")
 
-// TODO: Test Client Command-Line MsgSellData
-func (suite *txTestSuite) TestCmdSellData() {
+	require.NoError(t, err)
 
+	_, err = parseSellDataFlags(flags)
+	require.NoError(t, err)
```
hansol-medi, line 15: Why don't we assert that parsing results are the same to the expected values?

x/market/client/cli/txDeal.go
```diff
@@ -150,57 +150,41 @@ func NewSellDataMsg(clientCtx client.Context, txf tx.Factory, fs *flag.FlagSet)
...
-		return txf, nil, err
+		DealId:               sellData.UnsignedCert.DealId,
+		DataHash:             sellData.UnsignedCert.DataHash,
+		EncryptedDataUrl:     sellData.UnsignedCert.EncryptedDataUrl,
+		DataValidatorAddress: sellData.UnsignedCert.DataValidatorAddress,
+		RequesterAddress:     sellData.UnsignedCert.RequesterAddress,
 	}
 
 	signed := types.DataValidationCertificate{
 		UnsignedCert: &unSigned,
```
gyuguen, line 162:
```suggestion
		UnsignedCert: sellData.UnsignedCert,
```
I think you can use this as it is.

x/market/client/cli/txDeal.go
```diff
@@ -150,57 +150,41 @@ func NewSellDataMsg(clientCtx client.Context, txf tx.Factory, fs *flag.FlagSet)
...
-		Signature:    signatureBytes,
+		Signature:    sellData.Signature,
 	}
 
 	msg := types.NewMsgSellData(signed, clientCtx.GetFromAddress().String())
```
gyuguen, line 166:
```suggestion
	msg := types.NewMsgSellData(sellData, clientCtx.GetFromAddress().String())
```

I think you can use sellData s it is above.
And you might want to change the name of this.

x/market/client/cli/txDeal.go
gyuguen: Ah, Please ignore this comment.

x/market/client/cli/txDeal.go
inchori: It seems to better. I'll change it.

x/market/client/cli/txDeal_test.go
inchori: Oh yes, I'll add it.

youngjoon-lee (approved):
lgtm

gyuguen (approved):
LGTM 👍

PR #270, opened 2022-02-16
fix: Set Deal Status Completed

inchori:
Set deal status `COMPLETED` when deal's `cur_num_data` and `max_num_data` is equal.

### DIFF
--- a/x/market/keeper/deal_test.go
+++ b/x/market/keeper/deal_test.go
@@ -1,6 +1,7 @@
 package keeper_test
 
 import (
+	"strconv"
 	"testing"
 
 	"github.com/cosmos/cosmos-sdk/crypto/keys/secp256k1"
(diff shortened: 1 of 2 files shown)


x/market/keeper/deal_test.go
```diff
@@ -516,5 +552,29 @@ func makeTestCert() types.DataValidationCertificate {
...
 	}
+}
+
+func makeTestDataCert2() types.DataValidationCertificate {
+	uCert := types.UnsignedDataValidationCertificate{
+		DealId:               2,
+		DataHash:             []byte("1a312c1223x2fs3"),
+		EncryptedDataUrl:     []byte("https://panacea.org/a/123.json"),
+		DataValidatorAddress: newAddr.String(),
+		RequesterAddress:     acc2.String(),
```
gyuguen, line 563: It seems to me that only this address is a different value.
How about it takes an address as a parameter?

inchori: Ah you means the function likes ```func makeTestDataCert(address sdk.AccAddress) { ... }``` ?

gyuguen: That's right.

hansol-medi (approved):
LGTM.

The fail case of sell data to `COMPLETED` deal is treated in `TestDealIsNotActive()`.
In `TestDealIsNotActive()`, the status of deal is set directly, not by actual sell of data.
For better test of `COMPLETED` of deal, I think it's better to sell max num of data to the deal (to make deal status `COMPLETED`) and try sell more data with assertion of error.
This scenario can actually check if the deal set `COMPLETED` when max num of data sold.

x/market/keeper/deal_test.go
inchori: Not only different for address, but also for ```dataHash```. So I will takes a ```dataHash``` and ```address``` as a parameter.

gyuguen: Thank you :)

gyuguen (approved):
LGTM 👍

youngjoon-lee (commented):
LGTM

PR #282, opened 2022-03-17
refactor: rename module name

hansol-medi:
I split this issue for convenience in review.

In this PR, I just renamed module name from `market` to `datadeal`.


app/app.go
```diff
@@ -252,15 +252,15 @@ type App struct {
...
-	burnKeeper   burnkeeper.Keeper
-	tokenKeeper  tokenkeeper.Keeper
-	wasmKeeper   wasm.Keeper
-	marketKeeper marketkeeper.Keeper
+	aolKeeper      aolkeeper.Keeper
+	didKeeper      didkeeper.Keeper
+	burnKeeper     burnkeeper.Keeper
+	tokenKeeper    tokenkeeper.Keeper
+	wasmKeeper     wasm.Keeper
+	datadealKeeper datadealkeeper.Keeper
```
gyuguen, line 260: How about changing the field names declared in the struct as follows according to camel expression?
```suggestion
	dataDealKeeper datadealkeeper.Keeper
```

types/testsuite/suite.go
```diff
@@ -38,16 +38,16 @@ type TestSuite struct {
...
+	AccountKeeper     authkeeper.AccountKeeper
+	AolKeeper         aolkeeper.Keeper
+	AolMsgServer      aoltypes.MsgServer
+	BankKeeper        bankkeeper.Keeper
+	BurnKeeper        burnkeeper.Keeper
+	DIDMsgServer      didtypes.MsgServer
+	DIDKeeper         didkeeper.Keeper
+	TokenKeeper       tokenkeeper.Keeper
+	DatadealKeeper    datadealkeeper.Keeper
+	DatadealMsgServer datadealtypes.MsgServer
```
gyuguen, line 50: How about modifying it with camel expression here as well?
```suggestion
	DataDealKeeper    datadealkeeper.Keeper
	DataDealMsgServer datadealtypes.MsgServer
```

gyuguen (approved):
Thank you.
I left a few reviews.

app/app.go
hansol-medi: I thought that `datadeal` has been used as a single word intentionally.
Do you think it is better to declared with camel expression?

gyuguen: I don't think that's wrong either.

Well, our URL notation looks like `data-deal`, `data-pool`.(https://github.com/medibloc/panacea-ecosystem-arch/blob/main/data-marketplace/data-pool-model/NFT-model.md#downloading-data)

Looking at the notation above, I think it would be better to write the two words separately.

hansol-medi: I got it. Thanks for kind comments.
Then, it would be better to use separate words in module name as well, `data-deal`. wdyt?

gyuguen: Is the module name you are talking about this link?
https://github.com/medibloc/panacea-core/tree/ft/273/rename-module/x/datadeal

hansol-medi: Yes right. 
For now, however, I think we use camel expression for variables and `x/datadeal` for module name as we discussed before.

app/app.go
```diff
@@ -98,12 +98,12 @@ import (
 	"github.com/medibloc/panacea-core/v2/x/burn"
 	burnkeeper "github.com/medibloc/panacea-core/v2/x/burn/keeper"
 	burntypes "github.com/medibloc/panacea-core/v2/x/burn/types"
+	"github.com/medibloc/panacea-core/v2/x/datadeal"
+	dataDealkeeper "github.com/medibloc/panacea-core/v2/x/datadeal/keeper"
+	dataDealtypes "github.com/medibloc/panacea-core/v2/x/datadeal/types"
```
gyuguen, line 103: How about changing it to all lowercase?
Although it is an alias, it is a package, and all other aliases are used in lowercase letters.

```suggestion
	datadealkeeper "github.com/medibloc/panacea-core/v2/x/datadeal/keeper"
	datadealtypes "github.com/medibloc/panacea-core/v2/x/datadeal/types"
```

hansol-medi: I see.
I'll change all aliases with lowercase.

gyuguen: Good. Thank you :)

inchori (approved):
LGTM 👍

youngjoon-lee (approved):
lgtm

PR #283, opened 2022-03-21
feat: skeleton for `x/datapool` and proto files

hansol-medi:
- Skeleton codes for `x/datapool` have been made via starport.
- proto files and generated pb.go files for Txs
  - RegisterDataValidator
  - CreatePool
  - SellData
  - BuyDataAccessNFT
  - RedeemDataAccessNFT
- For query, a simple query for getting pool was added.
  - GetPool


inchori (approved):
LGTM 👍

proto/panacea/datapool/v2/pool.proto
```diff
@@ -0,0 +1,52 @@
...
+
+// Pool defines a data pool
+message Pool {
+  uint64                    pool_id = 1;
+  uint64                    round = 2;
+  repeated string           data_schema = 3;
+  uint64                    target_num_data = 4; // target number of data
+  uint64                    cur_num_data = 5;    // current number of data
+  uint64                    max_nft_supply = 6;  // max supply of data access NFT of the pool
+  uint64                    issued_nft = 7;      // current issued NFT
```
youngjoon-lee, line 18: is this `issued_nft` means the curation NFT ID?

youngjoon-lee (approved):
thank you

proto/panacea/datapool/v2/pool.proto
hansol-medi: The latter. It would be better to use `num_issued_nft` to make clearer. Thanks.

gyuguen (approved):
LGTM 👍

PR #285, opened 2022-03-23
feat: Register Data Validator Transaction

inchori:
Implement Register Data Validator Transaction
- Implement `RegisterDataValidator` method in `keeper/pool.go`.
- Configure `DataValidator` message in `tx.proto` for store data validator.

In next PR, I'll implement Cmd for Register Data Validator transaction.


proto/panacea/datapool/v2/pool.proto
```diff
@@ -49,4 +49,9 @@ message DataAccessNFTRedeemReceipt {
...
   string redeemer = 4;      // 'panacea1' address of redeemer
   uint64 block_height = 5;  // The block height at which NFT redeemed
-}
\ No newline at end of file
+}
+
+// DataValidator defines a detail of data validator.
+message DataValidator {
+  string endpoint = 1;
+}
```
inchori, line 57: I just put a `endpoint` field in `DataValidator` message. Refer to this [docs](https://github.com/medibloc/panacea-ecosystem-arch/blob/main/data-marketplace/data-pool-model/NFT-model.md#registering-a-data-validator-information), the some fields can be added.

x/datapool/types/message_pool.go
```diff
@@ -0,0 +1,168 @@
...
+
+func (msg *MsgRegisterDataValidator) GetSigners() []sdk.AccAddress {
+	dataValidator, err := sdk.AccAddressFromBech32(msg.Address)
+	if err != nil {
+		panic(err)
+	}
+	return []sdk.AccAddress{dataValidator}
+}
+
+var _ sdk.Msg = &MsgCreatePool{}
```
inchori, line 46: Because I need to write test code, I implemented all of our transaction such as `MsgCreatePool`, `MsgSellData`, `MsgBuyDataAccessNFT` in `message_pool.go`.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 TBD
 
+## [v2.0.2](https://github.com/medibloc/panacea-core/releases/tag/v2.0.2) - 2022-03-08
```
gyuguen, line 7: Is there a reason why the same version was added?

proto/panacea/datapool/v2/pool.proto
gyuguen: My guess is that the DataValidator should contain an address as well.
How about adding an address like this?
```
message DataValidator {
  string address = 1;
  string endpoint = 2;
}
```

Perhaps additional different values will be added during actual development.

proto/panacea/datapool/v2/tx.proto
```diff
@@ -31,7 +31,7 @@ service Msg {
 // MsgRegisterDataValidator defines the Msg/RegisterDataValidator request type.
 message MsgRegisterDataValidator {
   string address = 1;       // The address of data validator to be registered
-  string endpoint = 2;      // https://my-validator.org
+  DataValidator validator_detail = 2;      // https://my-validator.org
```
gyuguen, line 34: I think it would be good to come here only as an endpoint as before.

x/datapool/keeper/pool.go
```diff
@@ -0,0 +1,49 @@
...
+	store := ctx.KVStore(k.storeKey)
+	dataValidatorKey := types.GetKeyPrefixDataValidator(dataValidatorAddress)
+	if store.Has(dataValidatorKey) {
+		return sdkerrors.Wrapf(types.ErrDataValidatorAlreadyExist, "the data validator %s is already exists", dataValidatorAddress)
+	}
+
+	balance := k.bankKeeper.GetBalance(ctx, dataValidatorAddress, assets.MicroMedDenom)
+	if balance.IsZero() {
+		return sdkerrors.ErrInsufficientFunds
+	}
```
gyuguen, line 20: My guess is that there will be cases where MED is 0, although it is a valid validator address.
How about checking that the account exists and that the PublicKey exists in the account?

x/datapool/types/message_pool.go
```diff
@@ -0,0 +1,168 @@
...
+
+func (msg *MsgRegisterDataValidator) Route() string {
+	return RouterKey
+}
+
+func (msg *MsgRegisterDataValidator) Type() string {
+	return "RegisterDataValidator"
+}
+
+func (msg *MsgRegisterDataValidator) ValidateBasic() error {
```
gyuguen, line 25: It would be good to include an empty value check for `endpoints`.

x/datapool/types/message_pool.go
```diff
@@ -0,0 +1,168 @@
...
+
+func (msg *MsgCreatePool) Route() string {
+	return RouterKey
+}
+
+func (msg *MsgCreatePool) Type() string {
+	return "CreatePool"
+}
+
+func (msg *MsgCreatePool) ValidateBasic() error {
```
gyuguen, line 56: It would be good to check the required values here as well.

CHANGELOG.md
inchori: Hmm.. It's may be wrong commit. I'll delete it.

x/datapool/keeper/pool.go
inchori: Ah yes. We talked about it. I'll change it not only for this, but also for test code.

x/datapool/types/message_pool.go
inchori: I'll modify this function in next PR.

x/datapool/types/message_pool.go
hansol-medi: I'll add detailed validation for `MsgCreatePool` in #275

gyuguen: Aha, I understand it.

proto/panacea/datapool/v2/pool.proto
inchori: Yes, this might be better.

proto/panacea/datapool/v2/tx.proto
inchori: I just put only`DataValidator` field in `MsgRegisterDataValidator`. [551dd34](https://github.com/medibloc/panacea-core/pull/285/commits/551dd341b2af599174a28c493e717a8506f8c1d7)

gyuguen (approved):
LGTM 👍

hansol-medi (approved):
LGTM.

After merging this branch, I will rebase it for #275. Thanks!

PR #287, opened 2022-03-24
feat: MsgCreatePool Tx (w/o curator NFT)

hansol-medi:
Implemented `MsgCreatePool` transaction excepting curator NFT part. 
To mint curator NFT using CosmWasm contract when creating pool, deployment and instantiation of contract should precede. Since this issue is not resolved yet, minting NFT is not included in this PR. 

My first approach was that the module(`x/datapool`) would be a minter for curator NFT. That means all the curator NFT would be minted by module account. For this, the contract should be deployed and instantiated before all creation of pool so I tried that in `InitGenesis()` using [`wasmkeeper.create()`](https://github.com/CosmWasm/wasmd/blob/16ea5b031591a305442447a8972062f51f4e6cd5/x/wasm/keeper/keeper.go#L159). `wasmd tx store {wasmCode}` also use the wasmkeeper.create(). However, I have not figured it out yet. When storing contract, there was a compatibility issue of calling wasmVM. 

Before diving into this approach more, I want your thoughts for minting curator NFT.
If this approach is wrong or you have better idea, please leave comments or let's discuss about it. 

### remaining issues
- minting curator NFT when creating pool
- test for `MsgCreatePool`
- stuffs related to genesis


proto/panacea/datapool/v2/pool.proto
```diff
@@ -9,22 +9,14 @@ import "google/protobuf/duration.proto";
...
-  cosmos.base.v1beta1.Coin  deposit = 11;
-  google.protobuf.Duration  download_period = 12 [
-    (gogoproto.stdduration) = true
-  ];
-  string                    curator = 13;
-  string                    status = 14;
+  uint64     pool_id = 1;
+  string     pool_address = 2;
+  uint64     round = 3;
+  PoolParams pool_params = 4;
```
gyuguen, line 15: I wonder what the criteria for separating Pool and PoolParams are.
As expected, it seems that the input value and the state are divided.

x/datapool/client/cli/txPool.go
```diff
@@ -35,3 +42,75 @@ func CmdRegisterDataValidator() *cobra.Command {
...
+	contents, err := ioutil.ReadFile(file)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read file")
+	}
+
+	dec := json.NewDecoder(bytes.NewReader(contents))
+
+	if err := dec.Decode(&poolParamsInput); err != nil {
+		return nil, err
+	}
```
gyuguen, line 86: How about modifying it like this?
```suggestion
	err = json.Unmarshal(contents, &poolParamsInput)
	if err != nil {
		return nil, err
	}
```

x/datapool/types/pool.go
```diff
@@ -0,0 +1,72 @@
...
+		return verifier(bz)
+	}
+
+	if len(bz) == 0 {
+		return sdkerrors.Wrap(sdkerrors.ErrUnknownAddress, "address cannot be empty")
+	}
+
+	if len(bz) > address.MaxAddrLen {
+		return sdkerrors.Wrapf(sdkerrors.ErrUnknownAddress, "address max length is %d, got %d", address.MaxAddrLen, len(bz))
+	}
```
gyuguen, line 69: I think this part should be verified first.

gyuguen (commented):
Thank you for your hard work.
I left a few comments. 😀

x/datapool/types/pool.go
```diff
@@ -0,0 +1,72 @@
+package types
+
+import (
+	"fmt"
+	"strings"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/medibloc/panacea-core/v2/x/datadeal/v044_temp/address"
```
inchori, line 9: How about we refactor this package to under `x/` directory?

inchori (commented):
I left a few comment. Thank you

proto/panacea/datapool/v2/pool.proto
hansol-medi: Exactly. I divided the input values as a separate struct. 
Just pass inputs to `keeper.CreatePool` and make new pool in `keeper.CreatePool`.

x/datapool/client/cli/txPool.go
hansol-medi: Thanks. Modified in [6a4a3a8](https://github.com/medibloc/panacea-core/pull/287/commits/6a4a3a8d4e4f36ac9fb386eac6b09f47cf48a81b)

proto/panacea/datapool/v2/pool.proto
```diff
@@ -9,22 +9,14 @@ import "google/protobuf/duration.proto";
...
+  uint64     cur_num_data = 5;    // current number of data
+  uint64     num_issued_nft = 6;  // current issued NFT
```
youngjoon-lee, line 17:
```suggestion
  uint64     num_issued_nfts = 6;  // current issued NFTs
```

proto/panacea/datapool/v2/pool.proto
gyuguen: Yeah, It seems to good.

x/datapool/types/pool.go
```diff
@@ -0,0 +1,72 @@
...
+		Round:        1,
+		PoolParams:   &poolParams,
+		CurNumData:   0,
+		NumIssuedNft: 0,
+		Curator:      curator.String(),
+		Status:       PENDING,
+	}
+}
+
+func NewPoolAddress(poolId uint64) sdk.AccAddress {
```
youngjoon-lee, line 32:
```suggestion
func newPoolAddress(poolId uint64) sdk.AccAddress {
```
pls make this function private if it's not used outside of this package.

x/datapool/types/pool.go
```diff
@@ -0,0 +1,72 @@
...
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/medibloc/panacea-core/v2/x/datadeal/v044_temp/address"
+)
+
+const (
+	PENDING = "PENDING"
+	ACTIVE  = "ACTIVE"
+)
+
+func NewPool(poolId uint64, curator sdk.AccAddress, poolParams PoolParams) *Pool {
```
youngjoon-lee, line 17:
```suggestion
func NewPool(poolID uint64, curator sdk.AccAddress, poolParams PoolParams) *Pool {
```
https://github.com/golang/go/wiki/CodeReviewComments#initialisms

x/datapool/keeper/pool.go
```diff
@@ -63,3 +68,99 @@ func (k Keeper) SetDataValidator(ctx sdk.Context, dataValidator types.DataValida
...
+		if !k.IsRegisteredDataValidator(ctx, accAddr) {
+			return 0, sdkerrors.Wrapf(types.ErrNotRegisteredDataValidator, "the data validator %s is not registered", dataValidator)
+		}
+	}
+
+	// curator send deposit to pool for creation of pool
+	var deposit sdk.Coins
+	deposit = append(deposit, *newPool.PoolParams.Deposit)
+
+	err = k.bankKeeper.SendCoins(ctx, curator, poolAddress, deposit)
```
youngjoon-lee, line 116:
```suggestion
        deposit := sdk.NewCoins(*newPool.PoolParams.Deposit)
	err = k.bankKeeper.SendCoins(ctx, curator, poolAddress, deposit)
```
You can simplify this.

x/datapool/types/message_pool.go
```diff
@@ -61,6 +79,49 @@ func (msg *MsgCreatePool) ValidateBasic() error {
...
+			return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid issuer address of %s", issuerAddr)
+		}
+	}
+
+	if poolParams.Deposit == nil {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "the deposit is empty")
+	}
+	if !poolParams.Deposit.IsValid() {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "the deposit is invalid")
+	}
```
youngjoon-lee, line 123: In next PRs, we should define some genesis params (in the `genesis.json`) for this module: https://github.com/cosmos/cosmos-sdk/blob/master/x/slashing/types/params.go.

x/datapool/types/pool.go
youngjoon-lee: Also, `VerifyPoolAddressFormat` below.

proto/panacea/datapool/v2/genesis.proto
```diff
@@ -6,4 +6,6 @@ option go_package = "github.com/medibloc/panacea-core/x/datapool/types";
 import "google/protobuf/any.proto";
 
 // GenesisState defines the datapool module's genesis state.
-message GenesisState {}
\ No newline at end of file
+message GenesisState {
+  uint64 next_pool_number = 1;
```
youngjoon-lee, line 10: In next PRs, we need to add `pools` here. Please add a TODO comment for now.

x/datapool/types/pool.go
hansol-medi: Likewise, I think NFT would be better than Nft.

proto/panacea/datapool/v2/pool.proto
hansol-medi: modified in [c2f5cf9](https://github.com/medibloc/panacea-core/pull/287/commits/c2f5cf906992109ff1cfd74749ebe2b0e6782b70)

x/datapool/types/pool.go
hansol-medi: modified in [c2f5cf9](https://github.com/medibloc/panacea-core/pull/287/commits/c2f5cf906992109ff1cfd74749ebe2b0e6782b70)

x/datapool/types/pool.go
hansol-medi: modified in [c2f5cf9](https://github.com/medibloc/panacea-core/pull/287/commits/c2f5cf906992109ff1cfd74749ebe2b0e6782b70) and [f953e5c](https://github.com/medibloc/panacea-core/pull/287/commits/f953e5ca1499bdebe492b75185bd47a369c82032)

x/datapool/keeper/pool.go
hansol-medi: Thanks. I modified in [c2f5cf9](https://github.com/medibloc/panacea-core/pull/287/commits/c2f5cf906992109ff1cfd74749ebe2b0e6782b70) as you guided.

x/datapool/types/message_pool.go
hansol-medi: I added TODO comment and will do it next PR. Thanks

proto/panacea/datapool/v2/genesis.proto
hansol-medi: Also added TODO comment in genesis.proto. Thanks!

x/datapool/keeper/pool.go
```diff
@@ -63,3 +68,98 @@ func (k Keeper) SetDataValidator(ctx sdk.Context, dataValidator types.DataValida
...
+	poolAddress, err := types.AccPoolAddressFromBech32(newPool.GetPoolAddress())
+	if err != nil {
+		return 0, sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid address of pool %s", newPool.GetPoolAddress())
+	}
+
+	// check if account with pool address exists or not
+	acc := k.accountKeeper.GetAccount(ctx, poolAddress)
+	if acc != nil {
+		return 0, sdkerrors.Wrapf(types.ErrPoolAlreadyExist, "pool %d already exist", poolID)
+	}
```
gyuguen, line 95: Did you check this because of duplicate poolIDs?

x/datapool/types/message_pool.go
```diff
@@ -48,6 +48,24 @@ func (msg *MsgRegisterDataValidator) GetSigners() []sdk.AccAddress {
...
+		TrustedDataValidators: poolParams.TrustedDataValidators,
+		TrustedDataIssuers:    poolParams.TrustedDataIssuers,
+		Deposit:               poolParams.Deposit,
+		DownloadPeriod:        poolParams.DownloadPeriod,
+	}
+
+	return &MsgCreatePool{
+		Curator:    curator,
+		PoolParams: params,
+	}
```
gyuguen, line 66: I think it can be changed like this.
Is there any reason to declare this new?
```suggestion
	return &MsgCreatePool{
		Curator:    curator,
		PoolParams: poolParams,
	}
```

x/datapool/keeper/pool.go
hansol-medi: You are right. To be precise, it checks if the account according to the address generated using poolID exists or not. Since new account under the address is generated when creating pool.

x/datapool/types/message_pool.go
hansol-medi: Oh, there is no reason for this. Thanks. I'll refactor it

x/datapool/types/pool.go
hansol-medi: May I ask what does it mean? This verification is used for creating pool address. In that process, this verify the address is valid or not. What's the point I missed!?

x/datapool/types/pool.go
hansol-medi: I don't have any idea for this. It is not a big deal I think, so we can move it. Could you comment your thought @gyuguen and @youngjoon-lee ?

x/datapool/types/pool.go
gyuguen: AHA. I mean when `bz` is `nil` when this logic is first executed, `nil` is entered when calling `verifier(bz)`, so it would be a problem.
However, if `nil` processing is done well in `verifier(bz)`, there semms to be no problem. 😀

x/datapool/types/pool.go
gyuguen: My guess is that `v044_temp` is not a module, so it shouldn't go under `x/`.
Ummm.... If I had to move this, it would be to move it to the root folder.

x/datapool/types/pool.go
hansol-medi: I got it thanks. I'll change it!

x/datapool/keeper/pool.go
hansol-medi: I got your point. Yes. If collision occurs when creating pool address(it is generated by hashing), it will fail to create pool and will continue forever. 

We use hash for generating pool address assuming there is no collision. So I think there is no need to check if the address exists or not under this assumption (If collision occurs, we need a workaround for hash collision as you said, and it is not the only problem of this part. e.g. datadeal). Anyway I'll remove this check since there is no meaning under this assumption.

gyuguen: Yes, That will be fine. 😀

x/datapool/types/message_pool.go
hansol-medi: modified in [5f69ad0](https://github.com/medibloc/panacea-core/pull/287/commits/5f69ad06762947b77e90421a4b2487f058772fa2)

x/datapool/keeper/pool.go
hansol-medi: removed in [5f69ad0](https://github.com/medibloc/panacea-core/pull/287/commits/5f69ad06762947b77e90421a4b2487f058772fa2)

x/datapool/types/pool.go
hansol-medi: modified in [5f69ad0](https://github.com/medibloc/panacea-core/pull/287/commits/5f69ad06762947b77e90421a4b2487f058772fa2)

youngjoon-lee (approved):
lgtm

gyuguen (approved):
LGTM 👍

inchori (approved):
LGTM 👍 Thanks for your work.

PR #288, opened 2022-03-28
feat: NFT contract features in `x/datapool` module

hansol-medi:
As mentioned in #287, wasm NFT contract is required to be deployed to mint curator NFT when creating pool.
For that, 2 Txs and 1 Query are implemented. And add feature of minting curator NFT in `MsgCreatePool`.

Tx
- `MsgRegisterNFTContract` : creates, instantiates contract and sets the address of the generated contract
- `MsgUpgradeNFTContract` : creates new contract and migrate to new one

Query
- `QueryContract` : returns registered contract address

The way this contract being managed (e.g. by code, governance, etc.) has not been determined yet. So there are no restrictions on `MsgRegisterNFTContract` and `MsgUpgradeNFTContract` for now. When we decide how to manage it, it should be added. 

FYI)
For migration of contract, [`migrate()`](https://docs.cosmwasm.com/docs/1.0/smart-contracts/migration#setting-up-a-contract-for-migrations) function seems to be required. Since there was no `migrate()` in `cw721_base` contract, I tested migration by adding the function to the contract.


x/datapool/handler.go
```diff
@@ -24,6 +24,12 @@ func NewHandler(k keeper.Keeper) sdk.Handler {
 		case *types.MsgCreatePool:
 			res, err := msgServer.CreatePool(sdk.WrapSDKContext(ctx), msg)
 			return sdk.WrapServiceResult(ctx, res, err)
+		case *types.MsgRegisterNFTContract:
```
gyuguen, line 27: Anyone can create or modify the ContractAddress, so the ContractAddress may change unintentionally.
I think this should be done through governance or upgradeHandler.

x/datapool/handler.go
```diff
@@ -24,6 +24,12 @@ func NewHandler(k keeper.Keeper) sdk.Handler {
...
+			res, err := msgServer.RegisterNFTContract(sdk.WrapSDKContext(ctx), msg)
+			return sdk.WrapServiceResult(ctx, res, err)
+		case *types.MsgUpgradeNFTContract:
```
gyuguen, line 30: I think this above issue is the same goes for upgrades.
Because if a malicious person deploys and migrates the forged contract, all NFTs can be hijacked.

gyuguen (commented):
Thank you!
I addes a few things to discuss in the comments.

x/datapool/handler.go
hansol-medi: I totally agree with you 💯 . 
For now, anyone can deploy and migrate contract (no auth required). We have to make a decision for this.
But, it is out of scope of this PR I think. We can handle it in the next PR.

x/datapool/keeper/pool.go
```diff
@@ -163,3 +187,92 @@ func (k Keeper) SetPool(ctx sdk.Context, pool *types.Pool) {
...
+	if err != nil {
+		return 0, sdkerrors.Wrapf(err, "failed to create contract")
+	}
+
+	return codeID, nil
+}
+
+// DeployAndRegisterContract creates, instantiate contract and store contract address
+func (k Keeper) DeployAndRegisterContract(ctx sdk.Context, wasmCode []byte) error {
+	moduleAddr := types.GetModuleAddress()
```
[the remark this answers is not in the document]
hansol-medi: Thank you. I didn't know this module account. It would be better to declare module address in `app.go` as you suggested. nice!

x/datapool/handler.go
gyuguen: Yes, Grate!

x/datapool/keeper/pool.go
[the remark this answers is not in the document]
hansol-medi: The module account address is generated by the same way of other module account address. ([e0fe83f](https://github.com/medibloc/panacea-core/pull/288/commits/e0fe83f6c1350b66852643878efdd2e1237d766a))
https://github.com/medibloc/panacea-core/blob/7595cf300067e0e2969643cbb5a15acbf7960912/app/app.go#L633

Also, add to `maccPerms` as you mentioned to prevent from sending funds to module account. I knew that the `maccPerms` is used for not only declaration of module accounts' permission but prevention of sending funds to module account.

proto/panacea/datapool/v2/query.proto
```diff
@@ -12,6 +12,11 @@ service Query {
   rpc Pool(QueryPoolRequest) returns (QueryPoolResponse) {
     option (google.api.http).get = "/panacea/datapool/v2/pools/{pool_id}";
   }
+
+  // Contract returns a contract address registered to x/datapool module
+  rpc Contract(QueryContractRequest) returns (QueryContractResponse) {
+    option (google.api.http).get = "/panacea/datapool/v2/contract";
```
youngjoon-lee, line 18: It would be better to name this as `.../nft-contract`

x/datapool/README.md
```diff
@@ -0,0 +1,34 @@
+## Deploy and Register NFT smart contract
+
+```shell
+TX_FLAG=(--chain-id {your chainID} --gas auto --gas-prices 5umed --gas-adjustment 1.3)
+MINTER=$(panacead keys show {your address or key of minter} -a)
+panacead tx datapool register-contract cw721_base.wasm --from $MINTER $TX_FLAG -y
```
youngjoon-lee, line 6: If I understand it correctly, the minter is the module account. Then, the name `MINTER` is not proper in this example. It would be reasonable to name it as `SENDER` or so.

x/datapool/client/cli/txContract.go
```diff
@@ -0,0 +1,94 @@
+package cli
+
+import (
+	"fmt"
+	"io/ioutil"
+
+	wasmUtils "github.com/CosmWasm/wasmd/x/wasm/client/utils"
```
youngjoon-lee, line 7:
```suggestion
	wasmutils "github.com/CosmWasm/wasmd/x/wasm/client/utils"
```
Let's name packages with only lowercase chars. https://go.dev/doc/effective_go#package-names

x/datapool/keeper/pool.go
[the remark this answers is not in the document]
youngjoon-lee: 👍

youngjoon-lee (approved):
All right. Thank you for your work.

[As gyuguen said](https://github.com/medibloc/panacea-core/pull/288/files#r837175012), it would be risky to expose transactions for creating/upgrading the NFT contract.
Creating the NFT contract needs to be performed by the chain upgrade handler, as gyuguen tested.
And, upgrading the NFT contract needs to be performed by governance. For that, I don't think we have to define a new gov proposal type. We can use the existing proposal type defined by CosmWasm: `panacead tx gov submit-proposal migrate-contract`. 

So, at the next PR, let's implement a chain upgrade handler which stores/instantiates the NFT contract. Then, it's okay for me to delete transactions in this PR.
And, let's test if we can upgrade the NFT contract by `panacead tx gov submit-proposal migrate-contract`. But, if `cw-nfts` doesn't have the `migrate()` function yet, I think there is nothing much that we can in our side. Probably, we can contribute to `cw-nfts` in the future.

x/datapool/keeper/pool.go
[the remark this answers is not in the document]
youngjoon-lee: 👍

gyuguen (approved):
LGTM 👍

proto/panacea/datapool/v2/query.proto
hansol-medi: renamed in [abecf64](https://github.com/medibloc/panacea-core/pull/288/commits/abecf64bd84313eb2a8de05373eb08f76fe860c8)

x/datapool/README.md
hansol-medi: Yes it would be more proper name. changed it in [abecf64](https://github.com/medibloc/panacea-core/pull/288/commits/abecf64bd84313eb2a8de05373eb08f76fe860c8).

x/datapool/client/cli/txContract.go
hansol-medi: modified in [abecf64](https://github.com/medibloc/panacea-core/pull/288/commits/abecf64bd84313eb2a8de05373eb08f76fe860c8)

proto/panacea/datapool/v2/query.proto
youngjoon-lee: Ah. Sorry. Please use underscore instead of hyphen: `.../nft_contract`.

youngjoon-lee: Also, I would put the prefix `NFT` to all `Contract` symbols such as function names, rpc names and so on.

hansol-medi: I got it. It would be clearer. Let me add `NFT` prefix to all `Contract`

hansol-medi: done in [d9ff16a](https://github.com/medibloc/panacea-core/pull/288/commits/d9ff16a1f4e2195a4613f30d8cc5437ae32ef292)

youngjoon-lee (approved):

PR #293, opened 2022-03-31
feat: Query for Registered Data Validator

inchori:
- Query for Registered Data Validator.
- This feature will be used in Off-chain Data Validator for checking duplication of validator.


gyuguen (approved):
LGTM 👍

youngjoon-lee (approved):
danke

PR #294, opened 2022-03-31
feat: add `MsgCreatePool` genesis params

hansol-medi:
- `paramsKeeper.Subspace` has been added to `x/datapool` keeper.
- Implement `InitGenesis()` and `ExportGenesis()`
- The deposit for creating pool is moved to module params. It is a fixed value and managed by gov.


gyuguen (approved):
I think it would be good to write test code when you have time.
👍

hansol-medi:
Yes I'm writing test code for MsgCreatePool. It will be the next PR!

inchori (approved):
LGTM 👍

hansol-medi:
Genesis parameter of `nft_contract_address` was missed, so it has been added in #295

app/app.go
```diff
@@ -452,6 +452,7 @@ func New(
 		appCodec,
 		keys[datapooltypes.StoreKey],
 		keys[datapooltypes.MemStoreKey],
+		app.GetSubspace(datapooltypes.ModuleName),
```
youngjoon-lee, line 455: I haven't used Subspace yet, but IIRC, parameters stored in the Subspace cannot be modified by gov proposals. If so, I think we should use Keeper instead of Subspace as genesis parameters of other modules such as `x/staking`: https://github.com/cosmos/cosmos-sdk/blob/v0.42.11/x/staking/keeper/params.go.

Please make sure if it's okay for use to use Subspace.

youngjoon-lee (changes requested):

app/app.go
youngjoon-lee: You can also test if the parameter can be updated by gov proposals in your local, by submitting [ParameterChange proposals](https://docs.cosmos.network/v0.44/modules/gov/01_concepts.html#proposal-submission).

youngjoon-lee: Ah. I'm just confused. Sorry. I will review other parts again.

youngjoon-lee (approved):
lgtm

app/app.go
hansol-medi: I'm testing if it is working with submitting ParameterChange proposals. (it seems not working for now and fixing)
As I understand the example you shared (`x/staking`), `k.paramstore` is type of `paramtypes.Subspace` of staking keeper. I will merge this PR after testing parameter change with ParameterChange proposal.

hansol-medi:
The test of changing `data_pool_deposit` parameter by gov has been completed. Example of gov proposal json was added to README.md ([aa43977](https://github.com/medibloc/panacea-core/pull/294/commits/aa43977e2a44953d41fe8b53e0e4d70265dd8e40))

There are two minor changes after your review
- use lowercase for params key (cosmos use lowercase for params key)
- change type of `data_pool_deposit` to `sdk.Coin` from `sdk.Coins`

PR #295, opened 2022-04-01
feat: add tests for `MsgCreatePool` Tx

hansol-medi:
- add `wasmkeeper` to test suite for deployment of NFT contract in test
- tests for `MsgCreatePool` Tx have been added
  - `pool_keeper_test`
  - `grpc_query_pool_test`
  - `genesis_test`
- add `nft_contract_address` to genesis params


proto/panacea/datapool/v2/pool.proto
```diff
@@ -54,7 +54,11 @@ message PoolParams {
   repeated string          data_schema = 1;             // JSON Schema URIs
   uint64                   target_num_data = 2;         // A target num of data to be collected
   uint64                   max_nft_supply = 3;          // A max num of NFTs that can be minted (= A total num of potential buyers)
-  cosmos.base.v1beta1.Coin nft_price = 4;               // A price of a NFT
```
gyuguen, line 57: Is there any reason to change to accept multiple coins?

gyuguen (approved):
LGTM 👍 
I left one question :)

proto/panacea/datapool/v2/pool.proto
hansol-medi: Actually, I don't remember why I changed the type of `nft_price` to sdk.Coins. So I do roll back it

x/datapool/genesis.go
```diff
@@ -12,6 +12,10 @@ func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState)
 	k.SetParams(ctx, genState.Params)
 	k.SetPoolNumber(ctx, genState.NextPoolNumber)
 
+	if genState.NftContractAddress != nil {
+		k.SetNFTContractAddress(ctx, genState.NftContractAddress)
+	}
```
youngjoon-lee, line 17: I think it's okay to remove this `if` statement, even though the `DefaultGenesis.NftContractAddress` is `nil`: https://github.com/medibloc/panacea-core/pull/295/files#diff-1b224410c503ee7f657889a5370b3ff80d2396445a85534d2e4a544bbe357aecR12.

youngjoon-lee: Ah, I understood why you added this `if` statement. 
https://github.com/medibloc/panacea-core/blob/722287ea7d0b093a1a3517a54310791f5bdab032/x/datapool/keeper/pool.go#L253

Okay then, let's go as it is.

youngjoon-lee (approved):
lgtm

inchori (approved):
LGTM 😃

PR #296, opened 2022-04-04
feat: Update Registered Data Validator

inchori:
- Update (or Edit) Registered Data Validator.
- Data Validator can change their endpoint URL as @youngjoon-lee [said](https://github.com/medibloc/panacea-data-market-validator/pull/65#:~:text=FYI%2C%20we%20also,the%20next%20PR.).
- Implement keeper and client.


x/datapool/types/message_pool.go
```diff
@@ -46,6 +46,47 @@ func (msg *MsgRegisterDataValidator) GetSigners() []sdk.AccAddress {
...
+}
+
+func (msg *MsgUpdateDataValidator) Type() string {
+	return "UpdateDataValidator"
+}
+
+func (msg *MsgUpdateDataValidator) ValidateBasic() error {
+	_, err := sdk.AccAddressFromBech32(msg.ValidatorDetail.Address)
+	if err != nil {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid data validator address (%s)", err)
```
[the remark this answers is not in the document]
gyuguen: Sorry.  It is my mistake.
You used it right. 😅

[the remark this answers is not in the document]
gyuguen: Sorry.  It is my mistake.
You used it right. 😅

gyuguen (approved):
LGTM 👍

proto/panacea/datapool/v2/tx.proto
```diff
@@ -43,6 +46,16 @@ message MsgRegisterDataValidator {
 message MsgRegisterDataValidatorResponse {
 }
 
+// MsgUpdateDataValidator defines the Msg/UpdateDataValidator request type.
+message MsgUpdateDataValidator {
+  DataValidator validator_detail = 1;
```
youngjoon-lee, line 51: It'd be better to specify which fields can be updated, as Cosmos does: https://github.com/cosmos/cosmos-sdk/blob/6f79bf341741ab1f6c68804f4f113753d70179fe/proto/cosmos/staking/v1beta1/tx.proto#L65

proto/panacea/datapool/v2/tx.proto
```diff
@@ -15,6 +15,9 @@ service Msg {
   // RegisterDataValidator defines a method for registration of data validator.
   rpc RegisterDataValidator(MsgRegisterDataValidator) returns (MsgRegisterDataValidatorResponse);
 
+  // UpdateDataValidator defines a method for updating of data validator.
+  rpc UpdateDataValidator(MsgUpdateDataValidator) returns (MsgUpdateDataValidatorResponse);
```
youngjoon-lee, line 19: Good to have this Tx. Thank you.

But, let's not execute this tx automatically in the `panacea-data-market-validator`. If operators accidentally change their `endpoint` value in their `config.toml`, and if tx is executed automatically at that time, they will pay lots of tx fees unintentionally. So, it would be let them execute this tx maunally when they really need.

Then, I have a new concern that letting dataval operators execute the CreateDataValidator manually as well. Then, the dataval process can just check whether it was registered to Panacea. If not, it can print an error and exit.
I didn't think seriously about this topic before, because the dataval accepts a mnemonic anyway to use it for signing DataValidationCerts. But, I'm thinking if executing any txs automatically is degrading its UX. What do you think? @gyuguen @hansol-medi @inchori

gyuguen: Are you saying that you run RegisterDataValidator and UpdateDataValidator manually?

If so, I agree.
I think it's better to run this manually via command.
For tx that may cause problems, it seems better to operate manually by the user.

hansol-medi: I also agree to operate those Txs manually. It may cause problems as gyuguen said, and we couldn't make it fully automatic for all cases as youngjoon concerned.

proto/panacea/datapool/v2/tx.proto
inchori: Thanks, I'll change the comment.

proto/panacea/datapool/v2/tx.proto
inchori: Thanks for discussion!. Register and Update of Data Validator Transactions will be executed manually based on above reason.

proto/panacea/datapool/v2/tx.proto
youngjoon-lee: Ah. I meant defining updatable fields explicitly in the `MsgUpdateDataValidator` message, instead of using the `DataValidator` message as it is:
```proto
message MsgUpdateDataValidator {
  string endpoint = 1;
}
```
or so

inchori: Ah Okay I understand. The `--from` flag can define the `datavalidator` address which wants to update. I'll change it.

youngjoon-lee: Ah. 1 sec.. For the `--from` flag, we may need the `sender` field or so. Let me check.

youngjoon-lee: Yeah. You need the `validator_address` in the `MsgUpdateDataValidator` for verifying the tx signature.

youngjoon-lee: Then, let's go with 
```
message MsgUpdateDataValidator {
  string data_validator_address = 1; // signer
  string endpoint = 2;
}
```

Even though it looks a duplicated definition with the `DataValidator` message, it may be safer for the future.

inchori: Yes, but the `validator_address` and `endpoint` were included in `DataValidator` type. So l thought that there have no changes or rename the field name(`validator_detail`) to `update_data_validator`.

inchori: Okay, I'll change it.

youngjoon-lee (approved):
lgtm

gyuguen (approved):
LGTM 👍

PR #300, opened 2022-04-07
fix: datapool genesis validate condition bug

hansol-medi:
If genesis state doesn't have NFT contract address, the validation of genesis state fails. So, check first whether genesis state has NFT contract address.


gyuguen (approved):
LGTM 👍

hansol-medi:
In the next PR, the parameter of registered contract address will be moved to genesis params.

PR #301, opened 2022-04-07
feat: added sell data transaction

### LINKED ISSUE #276: Implement the MsgSellData transaction
Develop accoding to the following process.
* Receive `DataValidatorCertification` data from seller.
* After basic validation, and use the `pool-Id + round + dataHash` key combination to save toe Store.
* When saving complete, it responses with a `1 DP-{pool-Id}` token

Data validation
* Verification of the requestor and seller information of the certificate
* Validate the signature of the certificate
* Validate duplicate data with DataHash
* Comparison of certificate and pool information

Review items
* Token generation uses `x/token`. 
* Used to `cw20` in Cosmwasm

gyuguen:
Closes #276 

SellData transaction consists of the following.
* Compare the tx executor with the requester specified in the certificate.
* Validate the certificate.
* Check if the certificate has already been sold.
* Compare the current pool status value with the certificate information.
* Return 1 shareToken.

What is under review
* Shall we use the `x/token` module for shareToken


x/datapool/keeper/pool_test.go
```diff
@@ -138,3 +145,308 @@ func (suite *poolTestSuite) TestGetPool() {
...
+	// Modify the target data of the pool
+	pool.PoolParams.TargetNumData = 1
+	suite.DataPoolKeeper.SetPool(suite.Ctx, pool)
+
+	cert, err := makeTestDataCertificate(suite.Cdc.Marshaler, poolID, round, dataHash)
+	suite.Require().NoError(err)
+
+	shareToken, err := suite.DataPoolKeeper.SellData(suite.Ctx, requesterAddr, *cert)
+	suite.Require().NoError(err)
+	suite.Require().Equal("DP/1", shareToken.Denom)
```
gyuguen, line 235: Our initial denom is `DP-{poolID}`
However, when verifying in Coin, the following regular expression is used.
`[a-zA-Z][a-zA-Z0-9/]{2,127}`
So I changed `-` to `/`

x/datapool/keeper/pool.go
```diff
@@ -284,3 +281,140 @@ func (k Keeper) MigrateNFTContract(ctx sdk.Context, newWasmCode []byte) error {
...
+		return nil, err
+	}
+
+	k.SetDataValidationCertificate(ctx, cert)
+
+	k.increaseCurNumAndUpdatePool(ctx, pool)
+
+	shareToken := types.GetAccumPoolShareToken(pool.PoolId, 1)
+	// TODO We need to send the token to the seller.
+	// We now need to consider using the current 'x/token' or looking for another way.(ex. cw20)
```
gyuguen, line 315: I'll try to make good use of bankKeeper first to do this. (In another PR)

x/datapool/client/cli/txPool.go
```diff
@@ -130,3 +130,50 @@ func newCreatePoolMsg(clientCtx client.Context, file string) (sdk.Msg, error) {
...
+	}
+
+	flags.AddTxFlagsToCmd(cmd)
+	return cmd
+}
+
+func readCertificateFromFile(file string) (*types.DataValidationCertificate, error) {
+	contents, err := ioutil.ReadFile(file)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read file")
```
hansol-medi, line 169:
```suggestion
		return nil, fmt.Errorf("failed to read file: %w", err)
```

x/datapool/keeper/pool_test.go
```diff
@@ -183,6 +191,309 @@ func (suite *poolTestSuite) TestGetPool() {
...
+	suite.Require().Equal(sdk.NewInt(1), shareToken.Amount)
+
+	// Check the current pool status
+	getPool, err := suite.DataPoolKeeper.GetPool(suite.Ctx, poolID)
+	suite.Require().NoError(err)
+	suite.Equal(uint64(1), getPool.CurNumData)
+	suite.Equal(types.PENDING, getPool.Status)
+}
+
+func (suite *poolTestSuite) TestSellData_change_status_activity() {
```
hansol-medi, line 260: We may have to use consistent name for test. Looking at tests in `x/did`, one separator of underscore(_) seems to used between test subject(TestSellData) and case of each test(change_status_activity). And use Carmelcase. Which is better? I'm also going to change the name of test I made

x/datapool/keeper/pool_test.go
```diff
@@ -183,6 +191,309 @@ func (suite *poolTestSuite) TestGetPool() {
...
+	unsignedCertBz, err := suite.Cdc.Marshaler.MarshalBinaryBare(cert.UnsignedCert)
+	suite.Require().NoError(err)
+
+	curatorSign, err := curatorPrivKey.Sign(unsignedCertBz)
+	suite.Require().NoError(err)
+	cert.Signature = curatorSign
+
+	shareToken, err := suite.DataPoolKeeper.SellData(suite.Ctx, requesterAddr, *cert)
+	suite.Require().Error(err)
+	fmt.Println(err.Error())
```
hansol-medi, line 324:
```suggestion
```

x/datapool/keeper/pool_test.go
```diff
@@ -183,6 +191,309 @@ func (suite *poolTestSuite) TestGetPool() {
...
+
+func (suite *poolTestSuite) TestSellData_change_status_activity() {
+	poolID := uint64(1)
+	round := uint64(1)
+	dataHash := []byte("dataHash")
+
+	validatorAccount := suite.AccountKeeper.NewAccountWithAddress(suite.Ctx, dataVal1)
+	err := validatorAccount.SetPubKey(dataValPubKey)
+	suite.Require().NoError(err)
+	suite.AccountKeeper.SetAccount(suite.Ctx, validatorAccount)
```
hansol-medi, line 268: It would be useful to make separate function to set validator account. Widely used in this test.

x/datapool/keeper/pool_test.go
```diff
@@ -183,6 +191,309 @@ func (suite *poolTestSuite) TestGetPool() {
...
+	}
+
+	return &types.DataValidationCertificate{
+		UnsignedCert: &unsignedCert,
+		Signature:    sign,
+	}, nil
+}
+
+func makeTestDataPool(poolID uint64) *types.Pool {
+	nftPrice := sdk.NewCoin(assets.MicroMedDenom, sdk.NewInt(1000000))
```
hansol-medi, line 483: You can use `NFTPrice` declared in const.

hansol-medi (approved):
It looks good to me 👍 . I left a few trivial comments.

x/datapool/client/cli/txPool.go
gyuguen: Thank you

x/datapool/keeper/pool_test.go
gyuguen: yes i agree
So how about using it with CamelCase?

`TestSellData_change_status_activity` -> `TestSellDataChangeStatusActivity`

x/datapool/keeper/pool_test.go
gyuguen: Okay. I will use this.

x/datapool/keeper/pool_test.go
gyuguen: You're right. I will seperate the function to add validators.

x/datapool/keeper/pool_test.go
hansol-medi: Good. it is good to me. I wonder others' thoughts.

inchori (approved):
LGTM

x/datapool/keeper/pool.go
```diff
@@ -211,9 +210,25 @@ func (k Keeper) CreatePool(ctx sdk.Context, curator sdk.AccAddress, poolParams t
 	// store pool
 	k.SetPool(ctx, newPool)
 
+	// mint tokens as many as targetNumData
+	k.setInitialSupply(ctx, poolID)
+	err = k.mintPoolShareToken(ctx, poolID, poolParams.TargetNumData)
+	if err != nil {
+		return 0, sdkerrors.Wrapf(err, "failed to mint share token")
+	}
```
youngjoon-lee, line 218: We should keep in mind these two functions (set supply and mint tokens) must be always called together. If someone changes the supply and forgets to mint tokens, the cosmos-sdk will raise an invariant error that the total supply of the token is not matched with the actual minted amount.
To prevent this, we may define one function that not only sets supply but also mints tokens.

youngjoon-lee (commented):
Perfect! Thank you for neat implementations.

x/datapool/keeper/pool.go
youngjoon-lee: Ah. I just saw a related comment at the next PR #302: https://github.com/medibloc/panacea-core/pull/302#discussion_r850167499.
IIRC, whenever we mint more tokens, we have to increase the total supply of that token. So, I think the `setInitialSupply` and the `mintPoolShareToken` should be always called together. I'm not sure since I've read this part of codes long time ago.

We can test this scenario by:
- minting more tokens without increasing its supply
- dumping a new `genesis.json` by `panacead export ...`
- validating the `genesis.json` by `panacead validate-genesis ...`.

youngjoon-lee: Ah. Please forget about comments above. I haven't read the `setInitialSupply()` seriously that sets the supply to 0, not to the `TargetNumData`. Sorry for confusions.

PR #302, opened 2022-04-07
feat: added ability mint shareToken

gyuguen:
This function uses `BankKeeper` directly without using the `x/token` module.
Because I know that `x/module` is deprecated.

When data is sold, `shareToken` is minted and sent to the seller.

### DIFF
--- a/x/datapool/keeper/pool.go
+++ b/x/datapool/keeper/pool.go
@@ -9,6 +9,7 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 	gogotypes "github.com/gogo/protobuf/types"
 	"github.com/medibloc/panacea-core/v2/x/datapool/types"
 )
@@ -209,7 +210,18 @@ func (k Keeper) CreatePool(ctx sdk.Context, curator sdk.AccAddress, poolParams t
 	// store pool
 	k.SetPool(ctx, newPool)
 
+	// Initialize shareToken supply
+	k.setInitialSupply(ctx, poolID)
+
 	return newPool.GetPoolId(), nil
+}
+
+func (k Keeper) setInitialSupply(ctx sdk.Context, poolID uint64) {
+	supply := banktypes.Supply{
+		Total: sdk.NewCoins(types.GetAccumPoolShareToken(poolID, 0)),
+	}
+
+	k.bankKeeper.SetSupply(ctx, &supply)
 }
 
 func (k Keeper) GetNextPoolNumberAndIncrement(ctx sdk.Context) uint64 {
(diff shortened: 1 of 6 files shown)


x/datapool/keeper/pool_test.go
```diff
@@ -230,17 +230,18 @@ func (suite *poolTestSuite) TestSetDataCertificate() {
...
 
-	pool := makeTestDataPool(poolID)
-	suite.DataPoolKeeper.SetPool(suite.Ctx, pool)
+	suite.TestCreatePool()
+
+	pool, err := suite.DataPoolKeeper.GetPool(suite.Ctx, 1)
+	suite.Require().NoError(err)
+	poolID := pool.PoolId
+	round := pool.Round
+	dataHash := []byte("dataHash")
```
gyuguen, line 244: It called `CreatePool` because it need to initialize the supply for token minting.

x/datapool/types/expected_keepers.go
```diff
@@ -4,11 +4,19 @@ import (
...
+	"github.com/cosmos/cosmos-sdk/x/bank/exported"
 )
 
 type BankKeeper interface {
 	GetBalance(ctx sdk.Context, addr sdk.AccAddress, denom string) sdk.Coin
 	SendCoins(ctx sdk.Context, fromAddr sdk.AccAddress, toAddr sdk.AccAddress, amt sdk.Coins) error
+	SendCoinsFromModuleToAccount(ctx sdk.Context, senderModule string, recipientAddr sdk.AccAddress, amt sdk.Coins) error
+
+	MintCoins(ctx sdk.Context, moduleName string, amt sdk.Coins) error
+	BurnCoins(ctx sdk.Context, name string, amt sdk.Coins) error
```
gyuguen, line 16: It doesn't need this feature right now. However, since the minted shareToken may be deleted later, so I added it first.
Additionally, you need to add the authtypes.Burner permission.

hansol-medi: When do we need `BurnCoins()`? maybe revoke pool? Do you have any idea about this?

hansol-medi: When do we need `BurnCoins()`? maybe revoke pool? Do you have any idea about this?

hansol-medi (approved):
It looks good to me to use bankkeeper to mint share token.

x/datapool/keeper/pool.go
```diff
@@ -175,9 +176,21 @@ func (k Keeper) CreatePool(ctx sdk.Context, curator sdk.AccAddress, poolParams t
 	// store pool
 	k.SetPool(ctx, newPool)
 
+	// Initialize shareToken supply
+	k.setInitialSupply(ctx, poolID)
```
hansol-medi, line 180: When creating pool, we may be able to set total supply with `poolParams.targetNumData`, and send data share token to seller when selling data.

gyuguen: Yes, that is also possible.
If that happens, we will have to mint the shareToken as much as `poolParams.targetNumData` when creating the pool.

It's been a bit of a discussion, if we decide to distribute a reward every time an NFT is sold, we need to exclude for the ShareToken owned by the pool.
If this decision is not accepted, it seems better to mint in advance, as @hansol-medi suggested.

hansol-medi: If we distribute the rewards from NFT sales to seller and curator before reaching `poolParams.targeNumData` (we need further discussion as you said), we may have to calculate the portion of seller to total number of seller, which is equivalent to `poolParams.targeNumData`. 

In other words, I meant that **mint and send ShareToken every seller sell data** or **mint all ShareToken when creating pool and send every seller sell data**.

x/datapool/types/expected_keepers.go
gyuguen: I haven't really thought about it in detail yet.
Maybe refund?

gyuguen:
Fixed mint of shareToken when createPool operation is performed. :)
https://github.com/medibloc/panacea-core/pull/302/commits/a1c02f7f6a8d0bd91b0b73f94ea6915d26522b66

x/datapool/keeper/pool.go
```diff
@@ -436,6 +454,19 @@ func (k Keeper) increaseCurNumAndUpdatePool(ctx sdk.Context, pool *types.Pool) {
 	k.SetPool(ctx, pool)
 }
 
+func (k Keeper) mintPoolShareToken(ctx sdk.Context, poolID, amount uint64) error {
+	k.setInitialSupply(ctx, poolID)
```
hansol-medi, line 458: How about we separate this `setInitialSupply()` out from `mintPoolShareToken()`? 
If next round goes, data pool may mint more `ShareTokens` as much as increased `targetNumData`. No more initialization of supply may be required in next round curation.

gyuguen: Oh good. I'll fixed it

PR #303, opened 2022-04-08
feat: mint curator NFT & instantate contract

### LINKED ISSUE #275: Implement the MsgCreatePool transaction

hansol-medi:
Closes #275 

The NFT contract is stored and instantiated by governance ([wasm gov](https://github.com/CosmWasm/wasmd/blob/64d39c3c11f435b3a1b7a98856f0a89c3b404150/x/wasm/client/proposal_handler.go#L11)). Accordingly, the Txs and query related NFT contract were deleted and genesis parameters of contract were added.

To create data pool, 3 governances should be passed. Detailed example is provided in README.md.
1. store NFT contract wasm code
2. instantiate contract (set module as admin)
3. change parameter: register contract info to `x/datapool` module

FYI) The wasm proposal is disabled now, so you have to enable the wasm proposal if you want to test above things. This setting is not included in this PR. 
https://github.com/medibloc/panacea-core/blob/e44d49b8a3a7f4b0de3521b60ac2b0c094ec35c4/app/app.go#L122


x/datapool/keeper/pool.go
```diff
@@ -169,12 +177,38 @@ func (k Keeper) CreatePool(ctx sdk.Context, curator sdk.AccAddress, poolParams t
...
+
+	instantiateMsg := types.NewInstantiateNFTMsg(poolName, symbol, newPoolAddr)
+	instantiateMsgBz, err := json.Marshal(instantiateMsg)
+	if err != nil {
+		return 0, sdkerrors.Wrapf(err, "failed to marshal instantiateMsg")
+	}
+
+	codeID := k.GetParams(ctx).DataPoolCodeId
+
+	poolAccAddr, err := sdk.AccAddressFromBech32(newPoolAddr)
```
gyuguen, line 199: I think it chaged `sdk` to `types`
```suggestion
	poolAccAddr, err := types.AccAddressFromBech32(newPoolAddr)
```

gyuguen (commented):
Thank you.
I left some reviews. :)

x/datapool/keeper/pool.go
gyuguen: Ah, I checked and it doesn't seem to be used when creating the module address.
You don't need to change this.

gyuguen: There is a part that gets the same poolAddress on line 130 (duplicate code)

x/datapool/README.md
```diff
@@ -1,43 +1,113 @@
...
-panacead tx datapool create-pool create_pool_sample.json --from $CURATOR $TX_FLAG -y
+panacead tx gov vote {store proposal id} yes --from $VOTER $TX_FLAG -y
+```
+
+### Submit proposal (2): instantiate NFT contract
+
+After store NFT contract passed, instantiate the contract
+
+```shell
+INST_MSG=$(jq -n --arg name "curator" --arg symbol "CUR" --arg minter "panacea1xacc5pqnn00vf4mf8qvhe3y7k0xj4ky2hxgzvz" '{"name": $name, "symbol": $symbol, "minter": $minter}')
```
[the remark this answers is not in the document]
hansol-medi: Good. I missed it. I'll fix it

x/datapool/keeper/pool.go
hansol-medi: Thank you. It is duplicated and confused code. I will fix it

x/datapool/README.md
```diff
@@ -1,43 +1,113 @@
-## Deploy and Register NFT smart contract
+## Deploy and instantiate NFT smart contract
 
+### Setting voter
 ```shell
 TX_FLAG=(--chain-id {your chainID} --gas auto --gas-prices 5umed --gas-adjustment 1.3)
-SENDER=$(panacead keys show {your address or key of sender} -a)
-panacead tx datapool register-contract {your wasm code} --from $SENDER $TX_FLAG -y
+VOTER=$(panacead keys show {your voter} -a)
+panacead tx staking delegate $VALIDATOR 1000000umed --from $VOTER $TX_FLAG -y
```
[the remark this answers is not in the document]
hansol-medi: I'm not sure if I understand this comment correctly, this is just for setting voter(To vote, staking is required as you know). You meant that validator can also be a proposer? or voter?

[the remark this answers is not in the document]
gyuguen: Aha. What I mean that validators can also make proposals and vote. 
So I said it would be unnecessary to add a voter.(Because everything is possible with the validator)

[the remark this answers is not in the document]
hansol-medi: I got it. But, when validator submit a proposal and self-vote yes to the proposal, the proposal is rejected. So I made a voter by staking some coins.

x/datapool/README.md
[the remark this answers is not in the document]
hansol-medi: fixed in [49b64b5](https://github.com/medibloc/panacea-core/pull/303/commits/49b64b5401217524872f5f612a5bd52775b51d0a)

x/datapool/keeper/pool.go
hansol-medi: fixed in [49b64b5](https://github.com/medibloc/panacea-core/pull/303/commits/49b64b5401217524872f5f612a5bd52775b51d0a)

x/datapool/README.md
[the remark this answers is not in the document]
gyuguen: Nice 👍

x/datapool/keeper/pool.go
gyuguen: Good

hansol-medi:
When this PR is merged, e2e test of data validator should be modified accordingly (https://github.com/medibloc/panacea-data-market-validator/blob/main/e2e/init_start_panacea.sh).

x/datapool/README.md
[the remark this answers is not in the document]
gyuguen: As we said earlier, voting with the validator doesn't seem to be a problem.
Therefore, please review the part where the validator uploads the proposal as it is and votes.

gyuguen (approved):
LGTM 👍

x/datapool/types/pool.go
```diff
@@ -33,45 +29,8 @@ func NewPool(poolID uint64, curator sdk.AccAddress, poolParams PoolParams) *Pool
...
-	}
-
-	verifier := sdk.GetConfig().GetAddressVerifier()
-	if verifier != nil {
-		return verifier(bz)
-	}
-
-	return nil
+	poolKey := "pool" + strconv.FormatUint(poolID, 10)
+	return authtypes.NewModuleAddress(poolKey)
```
gyuguen, line 33: If the length of the address in the pool and the length of the general address are the same, is there any problem?
`address.Module(ModuleName, key)` length is 20.
But `authtypes.NewModuleAddress(poolKey)` length is 32.

hansol-medi: Thank you for your double-checking.
IIRC, this kind of module address is also used in `x/datadeal` and there was no specific reason to use 32-length address. 

Rather, 20-length address should be used because wasm keeper checks the length of address is 20. Since module & pool account address is used as a minter in wasm keeper, it makes sense to me.

hansol-medi: If higher version of wasm keeper can take longer address, we can also use it :)

x/datapool/README.md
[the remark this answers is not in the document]
hansol-medi: I see. Let me update README.md as you said 👍

x/datapool/types/pool.go
gyuguen: Ah, wasm checks it.
Then I'd like to go this way.

x/datapool/README.md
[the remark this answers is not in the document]
hansol-medi: updated in [bf74276](https://github.com/medibloc/panacea-core/pull/303/commits/bf7427611628823f82bf410cab814c43396bd692)

youngjoon-lee (commented):
Great! Codes look much cleaner than before.

PR #305, opened 2022-04-08
feat: add query params of datapool module

hansol-medi:
Implemented query params of datapool module. 
It may be used to get NFT contract address registered on `x/datapool` module


inchori (approved):
LGTM 👍

gyuguen (approved):
LGTM 👍

PR #306, opened 2022-04-13
feat: implementation of `MsgBuyDataAccessNFT` Tx

hansol-medi:
Implemented `MsgBuyDataAccessNFT` Tx, not queries.
As we discussed before, if the data pool is activated, the data access NFT is minted to the buyer. Otherwise, the buyer is added to white list of the pool. They are given the NFT when the pool is activated (maybe gyuguen is working now).

I use incrementing number as a id of NFT(`token_id` in contract), like stargaze. It is reasonable because each pool has each contract instance, and the id is used in each contract instance.


proto/panacea/datapool/v2/genesis.proto
```diff
@@ -9,12 +9,11 @@ import "panacea/datapool/v2/pool.proto";
...
-  uint64                                     next_pool_number = 2;
-  repeated Pool                              pools = 3;
-  Params                                     params = 4 [
-    (gogoproto.nullable) = false
-  ];
+  repeated DataValidator data_validators = 1 [(gogoproto.nullable) = false];
+  uint64                 next_pool_number = 2;
+  repeated Pool          pools = 3 [(gogoproto.nullable) = false];
+  Params                 params = 4 [(gogoproto.nullable) = false];
+  repeated WhiteList     white_list = 5 [(gogoproto.nullable) = false];
```
hansol-medi, line 16: I also add `[ (gogoproto.nullable) = false ]` to some fields. 

When I export `genesis.json` with multiple values of those fields, all the same values (I guess the last one) are stored and exported. I referred to other projects and they all used the gogoproto extension. I'm not sure of the reason, it was a workaround for now. If you have any idea, please leave some comments.

x/datapool/keeper/pool.go
```diff
@@ -328,3 +322,121 @@ func (k Keeper) DeployAndRegisterNFTContract(ctx sdk.Context, wasmCode []byte) (
...
+		}
+
+		_, err = k.wasmKeeper.Execute(ctx, contractAcc, poolAcc, mintMsgBz, sdk.NewCoins(types.ZeroFund))
+		if err != nil {
+			return sdkerrors.Wrapf(err, "failed to mint NFT")
+		}
+	} else {
+		// if pool is PENDING state, add buyer to white list
+		k.AddToWhiteList(ctx, poolID, buyer)
+	}
```
hansol-medi, line 376: Minting NFT will be modified after we make a decision of how and when to distribute NFTs.

proto/panacea/datapool/v2/genesis.proto
gyuguen: Are you saying this?
**expected:** 
```json
{
  "whiteList": [
    {
      "data_pool_code_id" : 1,
      "data_pool_nft_contract_adress": "panaceaxxxx1"
    },
    {
      "data_pool_code_id" : 2,
      "data_pool_nft_contract_adress": "panaceaxxxx2"
    }
  ]
}
```

**acutally**
```json
{
  "whiteList": [
    {
      "data_pool_code_id" : 2,
      "data_pool_nft_contract_adress": "panaceaxxxx2"
    }
  ]
}
```

x/datapool/README.md
```diff
@@ -1,4 +1,5 @@
 ## Deploy and instantiate NFT smart contract
+## Deploy and instantiate NFT smart contract
```
gyuguen, line 2: This comment was duplicate.

x/datapool/keeper/pool.go
```diff
@@ -129,7 +129,7 @@ func (k Keeper) CreatePool(ctx sdk.Context, curator sdk.AccAddress, poolParams t
 	// pool address for deposit
 	poolAddress, err := sdk.AccAddressFromBech32(newPoolAddr)
 	if err != nil {
-		return 0, sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "invalid address of pool %s", newPoolAddr)
+		return 0, sdkerrors.Wrapf(err, "invalid address of pool %s", newPoolAddr)
```
gyuguen, line 132: We need to define how to respond to the error.

Some places respond simply by appending a string.
Another place defines and uses the error type in the location below.
https://github.com/medibloc/panacea-core/blob/master/x/datapool/types/errors.go

I usually define the error type and append the error details after it.
Is there any good way?

x/datapool/keeper/pool.go
gyuguen: Right.
In my opinion, when the state of the Pool becomes ACTIVE, NFTs should be distributed along with ShareToken.

proto/panacea/datapool/v2/genesis.proto
```diff
@@ -25,4 +24,10 @@ message Params {
   ];
   uint64 data_pool_code_id = 2;
   string data_pool_nft_contract_address = 3;
+}
+
+// WhiteList define white list of data pool
+message WhiteList {
```
gyuguen, line 30: It looks like a list from the name alone, but it actually contains only one piece of information.
Umm.. Naming is very hard....
Please see below for reference only
```proto
message WatingForTransmissionNFT {
  uint64 pool_id = 1;
  string address = 2;
}

message WhiteList {
  repeat WatingForTransmissionNFT  = 1;
}
```

gyuguen (commented):
Thank you.
I left a few comment :)

proto/panacea/datapool/v2/genesis.proto
gyuguen: Well, no. I think this name is fine 😊

proto/panacea/datapool/v2/genesis.proto
hansol-medi: In your example,

```json
{
  "whiteList": [
    {
      "data_pool_code_id" : 2,
      "data_pool_nft_contract_adress": "panaceaxxxx2"
    },
    {
      "data_pool_code_id" : 2,
      "data_pool_nft_contract_adress": "panaceaxxxx2"
    }
  ]
}
```

The last one was duplicated..

x/datapool/keeper/pool.go
hansol-medi: Agreed. I'm gonna modify this part to distribute NFTs as what you've done in #311

x/datapool/README.md
hansol-medi: Oops, thank you

proto/panacea/datapool/v2/genesis.proto
hansol-medi: Ah, got your point. This `WhiteList` was intended to be a single piece of white list. But in `GenesisState` above, there is an array of white lists which have all pieces of white lists. I should use plural expression of white list in `GenesisState`.

x/datapool/keeper/pool.go
hansol-medi: If there is an error from functions (like `err` from `sdk.AccAddressFromBech32()`) and we check the `err` is not nil, I usually return the `err` wrapping with additional information to inform where the `err` comes from.

Otherwise we use custom logic for checking error case ([example](https://github.com/medibloc/panacea-core/blob/ba80e32aab61b1c9240bd08e8ff8e61b2d91913b/x/datapool/keeper/pool.go#L159)), I defined sentinel error and return it wrapping with some additional information.

I referred this post, https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully. Maybe @youngjoon-lee can give some advices about this.

x/datapool/README.md
hansol-medi: fixed in [e5d8c20](https://github.com/medibloc/panacea-core/pull/306/commits/e5d8c2039353b0cbd51d364469ca6f37e5359ce7)

proto/panacea/datapool/v2/genesis.proto
hansol-medi: In `GenesisState`, renamed it in [e5d8c20](https://github.com/medibloc/panacea-core/pull/306/commits/e5d8c2039353b0cbd51d364469ca6f37e5359ce7)
```proto
message GenesisState {
  ...
  repeated WhiteList     white_lists = 5 [(gogoproto.nullable) = false];
}
```

youngjoon-lee, line 30: IMHO, the word `WhiteList` is not proper for this case because we don't have an opposite concept `BlackList` in this case.
It would be enough to say `NFTTransferQueue` or so.

x/datapool/keeper/pool.go
youngjoon-lee: It depends on whether or not you want to expose the specified sentinel error to the function caller with ignoring whatever the child function returns.

If you think no caller cares about sentinel errors, you can just return a wrapped error. In this case, you actually don't need to use `sdkerrors.Wrapf()`. You can just use the `fmt.Errorf("%v", ...)`.

But, in general, we've returned wrapped sentinel errors in `x/.../keeper/...` packages, even though not all callers expect sentinel errors. So, I think it's better to follow that style for now. Honestly, I don't like the `sdkerrors` stuff.

proto/panacea/datapool/v2/genesis.proto
youngjoon-lee:
```proto
message GenesisState {
   ...
   repeated DelayedNFTTransfer delayed_nft_transfes
}

message DelayedNFTTransfer {
   uint64_t pool_id
   string address
}
```
What do you think about this?

hansol-medi: Good. 
And depending on when and how to distribute this NFT, we don't need this kind of buyer lists.
For now, I'll rename it!

x/datapool/keeper/pool.go
hansol-medi: Ok! I'll change it as gyuguen did. 🙏

x/datapool/keeper/pool.go
```diff
@@ -492,3 +486,121 @@ func (k Keeper) GetDataValidationCertificate(ctx sdk.Context, poolID, round uint
...
+		err := k.cdc.UnmarshalBinaryLengthPrefixed(bz, &list)
+		if err != nil {
+			return []types.WhiteList{}, err
+		}
+
+		whiteLists = append(whiteLists, list)
+	}
+
+	return whiteLists, nil
+}
```
inchori, line 606: It is not a big issue, I have a question. Is we need pagination for not only `whitelist`, but also for the others?

inchori (approved):
LGTM 👍

x/datapool/keeper/pool.go
hansol-medi: Good point. When I wrote the code, I forget pagination.
We may have to consider to add pagination to queries. I'll mark TODO. Thanks.

proto/panacea/datapool/v2/genesis.proto
gyuguen: That's interesting...

proto/panacea/datapool/v2/genesis.proto
gyuguen: Good.
I wish there was a similar name for the pool revenue distribute I'm developing

gyuguen (approved):
LGTM 👍

proto/panacea/datapool/v2/genesis.proto
hansol-medi: There are also differences of types in `GenesisState`. Without gogoproto extension, pointer is added to each type.

Without `[(gogoproto.nullable) = false]`, 
```go
type GenesisState struct {
    DataValidators []*DataValidator `protobuf:"bytes,1,rep,name=data_validators,json=dataValidators,proto3" json:"data_validators"`
    NextPoolNumber uint64           `protobuf:"varint,2,opt,name=next_pool_number,json=nextPoolNumber,proto3" json:"next_pool_number,omitempty"`
    Pools          []*Pool          `protobuf:"bytes,3,rep,name=pools,proto3" json:"pools"`
    Params         *Params          `protobuf:"bytes,4,opt,name=params,proto3" json:"params"`
}
```

With `[(gogoproto.nullable) = false]`, 
```go
type GenesisState struct {
    DataValidators []DataValidator `protobuf:"bytes,1,rep,name=data_validators,json=dataValidators,proto3" json:"data_validators"`
    NextPoolNumber uint64          `protobuf:"varint,2,opt,name=next_pool_number,json=nextPoolNumber,proto3" json:"next_pool_number,omitempty"`
    Pools          []Pool          `protobuf:"bytes,3,rep,name=pools,proto3" json:"pools"`
    Params         Params          `protobuf:"bytes,4,opt,name=params,proto3" json:"params"`
}
```

youngjoon-lee: I just read all changes again. Thank you.

Is this issue still being raised? Then, I will take a look at it.

youngjoon-lee: > There are also differences of types in GenesisState. Without gogoproto extension, pointer is added to each type.

This behavior is intended. But still, it's weird that the list contains all the same value.

hansol-medi: > Is this issue still being raised? Then, I will take a look at it.

This issue is resolved now, but I don't know what's going on.

> This behavior is intended.

Oh I see. I'm curious this part as well. I'll look at gogoproto extension more.

youngjoon-lee: I just removed `[(gogoproto.nullable) = false]` and checked that the exported list in the genesis contains two different elements. As I know, `[(gogoproto.nullable) = false]` shouldn't violate the fundamental behavior of protobuf marshalling/unmarhalling.

In protobuf, the nullability is a bit tricky to handle because everything is optional by default in protobuf 3. That's why all fields are defined with pointer such as `[]*DataValidator` or `*Params`. Previously, the protobuf 2 had the `required` keyword which sets some field not nullable, but that feature was deleted in protobuf 3: https://github.com/protocolbuffers/protobuf/issues/2497.

That's why the gogoproto, which is an extension of protobuf, introduced a new feature `nullable`. If you really want to make each list element not nullable, then you can use `[(gogoproto.nullable) = false]` so that gogoproto defines the list as `[]Element`, not `[]*Element`. I also agree with using it.

youngjoon-lee:
At https://github.com/medibloc/panacea-core/blob/fcab5982f4023b53412c547eb30fb46beedb7e03/x/datapool/keeper/pool.go#L59, you don't needa return an empty slice. Just `return nil, err`. The zero value of slice is nil.

proto/panacea/datapool/v2/genesis.proto
youngjoon-lee: > But still, it's weird that the list contains all the same value.

I guess some Go code wasn't implemented correctly at that time.

youngjoon-lee (approved):
lgtm

proto/panacea/datapool/v2/genesis.proto
hansol-medi: Oh, I see. I'm very thankful to you for your detailed explanation.

>I guess some Go code wasn't implemented correctly at that time.

I think it could be, because as I remember It was exported correctly when I implement `ExportGenesis()` of module params in #294.

inchori (approved):
LGTM 👍

hansol-medi:
> you don't needa return an empty slice which allocates memory. Just return nil, err. The zero value of slice is nil.

Thank you @youngjoon-lee !
It is changed in [77df372](https://github.com/medibloc/panacea-core/pull/306/commits/77df37204023df6cb7e394b3639c9a632e003982)

youngjoon-lee (approved):

x/datapool/keeper/pool.go
```diff
@@ -492,3 +484,61 @@ func (k Keeper) GetDataValidationCertificate(ctx sdk.Context, poolID, round uint
 
 	return cert, nil
 }
+
+func (k Keeper) BuyDataPass(ctx sdk.Context, buyer sdk.AccAddress, poolID, round uint64, payment sdk.Coin) error {
```
gyuguen, line 488: We must distribute a DataPass to a seller every time it is sold. So, when this PR is merged, I will add it to distribute to sellers whenever DataPass is sold.

gyuguen (approved):
LGTM 👍

PR #307, opened 2022-04-14
feat: add get module address query

hansol-medi:
Implemented a query for `x/datapool` [module address](https://github.com/medibloc/panacea-core/blob/64bcad178856d0e1cf2b127a83e4112ea8b80225/x/datapool/types/pool.go#L36).
It is required to register the module as a minter when storing and instantiating wasm contract. 
I also edited `README.md`, accordingly.


gyuguen (approved):
LGTM 👍

inchori (approved):
LGTM 👍

youngjoon-lee (commented):
lgtm

PR #310, opened 2022-04-14
docs: fix roadmap: whitepaper revision schedule

youngjoon-lee:
Previously, the whitepaper revision was in both 2022 Q3 and 2023 Q3. I deleted the one in 2023 Q3.


hansol-medi (approved):
T_T we all missed. sorry about that

tjyoon0324 (approved):
Fine with me! Thanks Youngjoon!

inchori (approved):
Thanks!! 👍

gyuguen (approved):
LGTM 👍

youngjoon-lee:
> T_T we all missed. sorry about that

No problem. I also thought we will probably revise the whitepaper more than twice in the future. But, I think it would be better to mention it only once in the roadmap.

PR #311, opened 2022-04-14
feat: add distribute revenue pools

gyuguen:
* In sellData, when the pool's status becomes ACTIVE, it signals a distribution of the proceeds.
* When a signal is received from BeginBlock of x/datapool, distribution starts to the seller who owns the shareToken.
* When the distribution starts, the curator also gets the deposit back.


gyuguen:
I'm currently thinking with error handling and messages.
We will reflect this in our next PR.

gyuguen:
Oops.. I can't edit the text.. 😂

The contents of `keep/pool.go` were too large, so it was separated into `keep/pool_distribute.go`.

gyuguen:
I will also add a Genesis related function to the next pr

hansol-medi:
> The contents of `keep/pool.go` were too large, so it was separated into `keep/pool_distribute.go`.

Good 👍 . I also thought we need to split the `pool.go`. I will name `pool_*.go` as you did.

x/datapool/types/pool.go
```diff
@@ -39,6 +39,26 @@ func GetModuleAddress() sdk.AccAddress {
...
+
+func (d *DistributeRevenuePools) AppendPoolID(poolID uint64) {
+	d.PoolIds = append(d.PoolIds, poolID)
+}
+
+func (d *DistributeRevenuePools) IsEmpty() bool {
+	return d.PoolIds == nil || len(d.PoolIds) == 0
+}
+
+func (d *DistributeRevenuePools) RemovePeviousIndex(idx int) {
```
hansol-medi, line 58: typo :)
```suggestion
func (d *DistributeRevenuePools) RemovePreviousIndex(idx int) {
```

x/datapool/keeper/pool_distribute.go
```diff
@@ -0,0 +1,177 @@
...
+			)
+
+			poolAddress, err := sdk.AccAddressFromBech32(pool.GetPoolAddress())
+			// stop distribute when an error occurs
+			if err != nil {
+				return true
+			}
+
+			// excluding moduleAccount
+			if k.accountKeeper.GetModuleAddress(types.ModuleName).Equals(addr) {
```
hansol-medi, line 61: You can use `types.GetModuleAddress()` which may returns the same address.

x/datapool/keeper/pool_distribute.go
```diff
@@ -0,0 +1,177 @@
...
+
+			if err != nil {
+				k.Logger(ctx).Error("failed to distribute coin", err)
+				return true
+			}
+			return false
+		})
+
+		// an error occur when distributing revenue to sellers
+		if err != nil {
```
hansol-medi, line 75: This err may not be from the `IterateAllBalances()`. 
The callback, which is an argument of `IterateAllBalances()`, returns true when error occurs, but the function `IterateAllBalances()` itself doesn't return anything. How can we catch the error from the iteration?

x/datapool/keeper/pool_distribute.go
gyuguen: Oh good. Thank you.
I'll changed it

x/datapool/keeper/pool_distribute.go
gyuguen: If any error occurs, this iterator stops running and terminates (return true).
When exiting, there is a value in err, so you can check whether an error has occurred by checking the error in this part.

Is there any part where the err value can be set other than IterateAllBalances()?

x/datapool/keeper/pool_distribute.go
```diff
@@ -0,0 +1,177 @@
...
+	totalShareTokenAmount := k.bankKeeper.GetSupply(ctx).GetTotal().AmountOf(shareTokenDenom)
+
+	// get the pool's total sales revenue
+	poolAddress, err := sdk.AccAddressFromBech32(pool.GetPoolAddress())
+	if err != nil {
+		return sdk.NewInt(0), err
+	}
+	poolSalesBalance := k.bankKeeper.GetBalance(ctx, poolAddress, assets.MicroMedDenom)
+	deposit := k.GetParams(ctx).DataPoolDeposit
+	totalSalesBalance := poolSalesBalance.Sub(deposit)
```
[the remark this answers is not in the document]
gyuguen: Yes, that's right. This part was going to be done in another PR.
As you said, I will write a TODO first.

x/datapool/keeper/pool_distribute.go
```diff
@@ -0,0 +1,177 @@
...
+
+	// calculator the amount to be distributed to each seller
+	if totalSalesBalance.Amount.Equal(sdk.NewInt(0)) {
+		return sdk.NewInt(0), nil
+	}
+	eachDistributionAmount := totalShareTokenAmount.Quo(totalSalesBalance.Amount)
```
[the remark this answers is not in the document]
gyuguen: Oh right. you're right.
I will improve this part so that it can be verified by testing.

x/datapool/keeper/pool_distribute.go
```diff
@@ -0,0 +1,177 @@
...
+
+	return eachDistributionAmount, nil
+}
+
+func (k Keeper) sendDepositToCurator(ctx sdk.Context, pool *types.Pool) error {
+	deposit := k.GetParams(ctx).DataPoolDeposit
```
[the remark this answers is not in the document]
gyuguen: Thank you. I fixed it.

x/datapool/keeper/pool_distribute.go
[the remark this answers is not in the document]
gyuguen: I added this and verified it with a test.
https://github.com/medibloc/panacea-core/pull/311/commits/9ae0249839142138a4922cd0c4ad10a992a44b5b

x/datapool/keeper/pool_distribute.go
gyuguen: Completed!
https://github.com/medibloc/panacea-core/pull/311/commits/9ae0249839142138a4922cd0c4ad10a992a44b5b

x/datapool/keeper/pool_distribute.go
gyuguen: I found the problem you mentioned and fixed it.
Thank you :)

https://github.com/medibloc/panacea-core/pull/311/commits/ac000361124256562269861ca5e2c145265de268#diff-815b7bd8fbeec0907b84af82002300154955256bf90959d4b142a0f84b75830fR54

gyuguen:
As our specifications change, this PR will be improved after drafting.

gyuguen:
Since most of the features here have changed, I will close this PR and reopen it.

PR #314, opened 2022-04-21
refactor: add deposit rate

hansol-medi:
We previously have `data_pool_deposit`, which is a fixed value of deposit for creation data pool. 
In this PR, it is changed to `data_pool_deposit_rate` as a governance parameter, which is proportional to the expected total sales of data pass. (0 <= `data_pool_deposit_rate` <= 1)

_`required_deposit` = (`total_supply_data_pass` * `price_data_pass`) * `deposit_rate`_

Accordingly, check is added whether the deposit curator sent is greater than or equal to the required deposit.
Plus, the deposit field is added to pool, (refer [the comment](https://github.com/medibloc/panacea-core/pull/311#discussion_r851121658))


youngjoon-lee (approved):
lgtm

gyuguen (approved):
LGTM 👍

PR #316, opened 2022-04-26
feat: Implement Redeem Data Pass

inchori:
- The Command-Line for Redeem Data Pass will be implemented in next PR based on this branch.
- Write unit test code for `keeper/RedeemDataPass`.


x/datapool/types/keys.go
```diff
@@ -51,3 +53,9 @@ func GetKeyPrefixDataValidateCertByRound(poolID, round uint64) []byte {
 func GetKeyPrefixDataValidateCert(poolID, round uint64, dataHash []byte) []byte {
 	return append(GetKeyPrefixDataValidateCertByRound(poolID, round), dataHash...)
 }
+
+func GetKeyPrefixNFTRedeemReceipt(poolID, round uint64, redeemer sdk.AccAddress) []byte {
+	poolAppend := append(KeyPrefixNFTRedeemReceipts, sdk.Uint64ToBigEndian(poolID)...)
+	roundAppend := append(poolAppend, sdk.Uint64ToBigEndian(round)...)
```
gyuguen, line 59: My guess is that a value with poolID 1 and round 11 and a value with poolID 11 and round 1 will come out the same.
How about using CombineKey like dataDeal?
https://github.com/medibloc/panacea-core/blob/master/x/datadeal/types/keys.go#L48

gyuguen (commented):
Thank you.
I left few comment :)

x/datapool/keeper/pool.go
```diff
@@ -553,3 +544,129 @@ func (k Keeper) increaseNumIssuedNFT(ctx sdk.Context, pool *types.Pool) {
...
+
+	store := ctx.KVStore(k.storeKey)
+	receiptKey := types.GetKeyPrefixNFTRedeemReceipt(redeemReceipt.PoolId, redeemReceipt.Round, redeemer)
+	bz := k.cdc.MustMarshalBinaryLengthPrefixed(&redeemReceipt)
+	store.Set(receiptKey, bz)
+
+	return nil
+}
+
+func (k Keeper) GetRedeemerDataPass(ctx sdk.Context, poolID uint64, redeemer sdk.AccAddress) ([]string, error) {
```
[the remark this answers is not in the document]
inchori: Yes, my mistake.
I'll change like [this](https://github.com/medibloc/panacea-core/blob/5db2d0e4ed7dad3593505bb4c7e8b17d2cd9d712/x/datapool/keeper/pool.go#L634) in next PR. How about it?

x/datapool/types/keys.go
inchori: Thanks, the key will be changed like [this](https://github.com/medibloc/panacea-core/blob/ft/280/redeem_data_pass_cli/x/datapool/types/keys.go), too. Because of the pagination and the request of data validator for data pass redeem receipts.

x/datapool/keeper/pool.go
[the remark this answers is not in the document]
gyuguen: Well, I'm not sure what the difference is between the link you gave me and the current source. 🤔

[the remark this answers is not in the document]
inchori: Ah, sorry. I was confused the branch. I think your idea is more better. The `GetPool` will be called over and over again. I understand. Thanks!

x/datapool/types/keys.go
gyuguen: I got it. I'll check back in the next PR for changes.

x/datapool/keeper/pool.go
```diff
@@ -553,3 +544,129 @@ func (k Keeper) increaseNumIssuedNFT(ctx sdk.Context, pool *types.Pool) {
 
 	k.SetPool(ctx, pool)
 }
+
+func (k Keeper) RedeemDataPass(ctx sdk.Context, redeemNFT types.MsgRedeemDataPass) (*types.DataPassRedeemReceipt, error) {
```
gyuguen, line 548: What if the same person plays Redeem multiple times?

x/datapool/keeper/pool.go
```diff
@@ -553,3 +544,129 @@ func (k Keeper) increaseNumIssuedNFT(ctx sdk.Context, pool *types.Pool) {
...
+	}
+
+	redeemTokenId, err := k.GetRedeemerDataPass(ctx, pool.PoolId, redeemerAcc)
+	if err != nil {
+		return nil, sdkerrors.Wrapf(types.ErrRedeemDataPass, err.Error())
+	}
+
+	if !contains(redeemTokenId, strconv.FormatUint(redeemNFT.NftId, 10)) {
+		return nil, types.ErrInvalidTokenId
+	}
```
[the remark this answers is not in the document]
inchori: Okay, I'll change the error status. The invalid token id error is quite ambiguous.

x/datapool/types/keys.go
```diff
@@ -51,3 +53,9 @@ func GetKeyPrefixDataValidateCertByRound(poolID, round uint64) []byte {
 func GetKeyPrefixDataValidateCert(poolID, round uint64, dataHash []byte) []byte {
 	return append(GetKeyPrefixDataValidateCertByRound(poolID, round), dataHash...)
 }
+
+func GetKeyPrefixNFTRedeemReceipt(poolID, round uint64, redeemer sdk.AccAddress) []byte {
```
hansol-medi, line 57: This key generated by using pool ID, round, and redeemer's address may cause conflicts.
If a buyer has multiple `DataPass`es and redeem them in the same round, the redeem receipt may have the same key. Thus, we may have to use token ID.

x/datapool/keeper/pool.go
hansol-medi, line 634: The name of this keeper function is confusing. I know it's the hardest thing among what we do 😢 
I suggest a new name.
```suggestion
func (k Keeper) GetAllDataPassByAddr(ctx sdk.Context, poolID uint64, redeemer sdk.AccAddress) ([]string, error) {
```

hansol-medi (commented):
Thank you for your efforts!
I reviewed and left some comments.

x/datapool/keeper/pool.go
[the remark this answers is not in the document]
hansol-medi: > Does @hansol-medi mean taking all the DataPass that the redeemer owns?

No I didn't mean that. This function is for getting a list of DataPassRedeemReceipts used by redeemer in a pool.

If this `GetRedeemerDataPass()` accepts contractAddress as an argument, the caller should always know the contractAddress, which may be retrieved from `GetPool()`. In other words, this function may have a dependency on the `GetPool()` function.

What if we need a query for getting a list of DataPassRedeemReceipts used by redeemer in a pool? Then, I think, the argument of poolID makes sense as an independent function.

Umm, however, if we need that query above, we can get pool contractAddress first, and then get redeem receipts and return it. I'm not sure which is the best way..

x/datapool/keeper/pool.go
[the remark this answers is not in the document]
inchori: How about `ErrNotFoundRedeemerNft` or `ErrNotOwnedRedeemerNft`?

x/datapool/keeper/pool_test.go
```diff
@@ -382,6 +383,62 @@ func (suite poolTestSuite) TestNotEnoughDeposit() {
...
+	err := suite.BankKeeper.AddCoins(suite.Ctx, buyerAddr, fundForBuyer)
+	suite.Require().NoError(err)
+
+	err = suite.DataPoolKeeper.BuyDataPass(suite.Ctx, buyerAddr, poolID, 1, NFTPrice)
+	suite.Require().NoError(err)
+
+	pool, err := suite.DataPoolKeeper.GetPool(suite.Ctx, poolID)
+	suite.Require().NoError(err)
+
+	redeemNFT := types.NewMsgRedeemDataPass(poolID, 1, pool.GetNumIssuedNfts(), buyerAddr.String())
```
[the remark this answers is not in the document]
inchori: Thanks, for now, I'll just use hard coded `tokenID`, 1.

x/datapool/types/keys.go
inchori: Okay, it can be more better. If I applied the `tokenID`, we might need to modify our architecture docs later.

x/datapool/keeper/pool.go
hansol-medi: When redeemer redeems data pass, the data pass(NFT) is transferred to module account, so replay attack can't happen. Did you mean this kind of replay attack, right?

gyuguen: Oh sorry i misunderstood
I thought this function was looking up RedeemDataPass.
You can ignore this comment.

gyuguen: > When redeemer redeems data pass, the data pass(NFT) is transferred to module account, so replay attack can't happen. Did you mean this kind of replay attack, right?
As @hansol-medi said, replay attacks are not a problem.

x/datapool/keeper/pool.go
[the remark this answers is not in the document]
gyuguen: I think `ErrNotOwnedRedeemerNft` looks good.

x/datapool/keeper/pool.go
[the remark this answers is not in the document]
gyuguen: Yes. I thought that it was unnecessary to search the pool again here because it had already looked up the pool once in this upper function.
There is nothing wrong with what @hansol-medi said.

If I could implement it, I think I could try something like this.
Of course, this is just my opinion, and @inchori  can go with the implementation as it is now.

First, since `nftContractAcc` has already been declared in the parent function, it is passed as a parameter like this.
```go
func (k Keeper) GetRedeemerDataPass(ctx sdk.Context, poolID uint64, nftContractAcc, redeemer sdk.AccAddress) ([]string, error) {
    // code...
}
```
If you need to call the GetRedeemer function only with poolID, you can add it as follows.
```go
func (k Keeper) GetRedeemerDataPass(ctx sdk.Context, poolID uint64, redeemer sdk.AccAddress) ([]string, error) {
    pool, err := k.GetPool(ctx, poolID)
    if err != nil {
	return nil, sdkerrors.Wrapf(types.ErrPoolNotFound, err.Error())
    }

    acc, err := sdk.AccAddressFromBech32(pool.NftContractAddr)
    if err != nil {
    	return nil, sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, err.Error())
    }
    return k.GetRedeemerDataPassWithNftContractAcc(ctx, poolID, acc, redeemer)
}

func (k Keeper) GetRedeemerDataPassWithNftContractAcc(ctx sdk.Context, poolID uint64, nftContractAcc, redeemer sdk.AccAddress) ([]string, error) {
    // code...
}
```

Again, this is just my opinion, please keep in mind.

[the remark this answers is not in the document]
hansol-medi: Good. It's enough with the `GetRedeemerDataPassWithNftContractAcc()` for now, I think.
When we need `GetRedeemerDataPass()` or the related query, we can add it.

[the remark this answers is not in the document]
inchori: Thanks for discussion. For now, I'll implement `GetRedeemerDataPassWithNftContractAcc()` and use it. But I'll leave `GetRedeemerDataPass()` method as well.

x/datapool/keeper/pool.go
```diff
@@ -553,3 +544,147 @@ func (k Keeper) increaseNumIssuedNFT(ctx sdk.Context, pool *types.Pool) {
...
+
+	zeroFund := sdk.NewCoins(types.ZeroFund)
+
+	_, err = k.wasmKeeper.Execute(ctx, nftContractAcc, redeemerAcc, transferNFTMsgBz, zeroFund)
+	if err != nil {
+		return nil, sdkerrors.Wrapf(types.ErrRedeemDataPass, err.Error())
+	}
+
+	nftRedeemReceipt := types.NewDataPassRedeemReceipt(redeemNFT.PoolId, redeemNFT.Round, redeemNFT.NftId, uint64(ctx.BlockHeight()), redeemNFT.Redeemer)
+
```
inchori, line 593: For now, I used `ctx.BlockHeight()`. But in our architecture docs, the `if possible, use block_time` was written. So when all of the other derived branches are merged to this branch, I'll try to change `ctx.BlockTime()`.

proto/panacea/datapool/v2/tx.proto
```diff
@@ -27,8 +27,8 @@ service Msg {
   // BuyDataPass defines a method for buying data access NFT
   rpc BuyDataPass(MsgBuyDataPass) returns (MsgBuyDataPassResponse);
 
-  // RedeemDataAccessNFT defines a method for redeeming data access NFT to get data
-  rpc RedeemDataAccessNFT(MsgRedeemDataAccessNFT) returns (MsgRedeemDataAccessNFTResponse);
+  // RedeemDataPass defines a method for redeeming data access NFT to get data
```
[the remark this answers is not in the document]
hansol-medi: It is not applied yet :)

x/datapool/keeper/pool.go
```diff
@@ -553,3 +544,147 @@ func (k Keeper) increaseNumIssuedNFT(ctx sdk.Context, pool *types.Pool) {
...
+		return nil, sdkerrors.Wrapf(types.ErrRedeemDataPass, err.Error())
+	}
+
+	return nftRedeemReceipt, nil
+}
+
+func (k Keeper) GetDataPassRedeemReceipt(ctx sdk.Context, poolID, round uint64, redeemer sdk.AccAddress) (types.DataPassRedeemReceipt, error) {
+	key := types.GetKeyPrefixNFTRedeemReceipt(poolID, round, redeemer)
+	store := ctx.KVStore(k.storeKey)
+
```
hansol-medi, line 605: You may have to check if the store has the key or not.

x/datapool/keeper/pool.go
```diff
@@ -553,3 +544,147 @@ func (k Keeper) increaseNumIssuedNFT(ctx sdk.Context, pool *types.Pool) {
...
+	var res types.QueryTokensResponse
+	err = json.Unmarshal(data, &res)
+	if err != nil {
+		return nil, err
+	}
+
+	return res.Tokens, nil
+}
+
+func (k Keeper) GetRedeemerDataPassWithNFTContractAcc(ctx sdk.Context, poolID uint64, nftContractAcc, redeemer sdk.AccAddress) ([]string, error) {
```
hansol-medi, line 662:
```suggestion
func (k Keeper) GetRedeemerDataPassWithNFTContractAcc(ctx sdk.Context, nftContractAcc, redeemer sdk.AccAddress) ([]string, error) {
```
The argument `poolID` is not used here

x/datapool/keeper/pool.go
```diff
@@ -553,3 +544,147 @@ func (k Keeper) increaseNumIssuedNFT(ctx sdk.Context, pool *types.Pool) {
...
+	data, err := k.viewKeeper.QuerySmart(ctx, acc, queryBz)
+	if err != nil {
+		return nil, err
+	}
+
+	var res types.QueryTokensResponse
+	err = json.Unmarshal(data, &res)
+	if err != nil {
+		return nil, err
+	}
```
hansol-medi, line 657: Instead of this code, you can use the below `GetRedeemerDataPassWithNFTContractAcc()` function.
```suggestion
tokens, err := k.GetRedeemerDataPassWithNFTContractAcc(ctx, acc, redeemer)
```

hansol-medi (commented):
Leave minor comments.
In this PR, the key generated for data pass receipt doesn't use token ID. It is applied in other PR. After merging other PRs into this branch, I want to review all the merged code.

x/datapool/keeper/pool.go
inchori: I'll apply these comments after the other branches are merged to this PR.

x/datapool/keeper/pool.go
```diff
@@ -553,3 +544,147 @@ func (k Keeper) increaseNumIssuedNFT(ctx sdk.Context, pool *types.Pool) {
...
+		return types.DataPassRedeemReceipt{}, sdkerrors.Wrap(types.ErrGetDataPassRedeemReceipt, err.Error())
+	}
+
+	return receipt, nil
+}
+
+func (k Keeper) SetDataPassRedeemReceipt(ctx sdk.Context, redeemReceipt types.DataPassRedeemReceipt) error {
+	redeemer, err := sdk.AccAddressFromBech32(redeemReceipt.Redeemer)
+	if err != nil {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, err.Error())
```
gyuguen, line 620: Since this error is wrapped in the parent function, it would be good to just respond to the error.
```suggestion
		return err
```

x/datapool/keeper/pool.go
```diff
@@ -553,3 +544,147 @@ func (k Keeper) increaseNumIssuedNFT(ctx sdk.Context, pool *types.Pool) {
...
+	store.Set(receiptKey, bz)
+
+	return nil
+}
+
+func (k Keeper) GetRedeemerDataPassByAddr(ctx sdk.Context, poolID uint64, redeemer sdk.AccAddress) ([]string, error) {
+	pool, err := k.GetPool(ctx, poolID)
+	if err != nil {
+		return nil, sdkerrors.Wrapf(types.ErrPoolNotFound, err.Error())
+	}
```
gyuguen, line 635: It seems that `k.GetPool()` already responsd with the same error.
What if it just responds with an error?
```suggestion
	pool, err := k.GetPool(ctx, poolID)
	if err != nil {
		return nil, err
	}
```

gyuguen (approved):
LGTM 👍 
I left a few comment.

x/datapool/keeper/pool.go
inchori: It will be also fixed when after other PR is merged.

youngjoon-lee (approved):
lgtm

x/datapool/keeper/pool.go
youngjoon-lee: IIRC, we don't need the block time because we removed the download_period concept (Please correct me if I'm wrong).
So, I think the block height in the DataPassRedeemReceipt is also not mandatory. But, let's keep it for clarity.

x/datapool/keeper/pool.go
inchori: Fixed in 6fb7f18.

x/datapool/keeper/pool.go
hansol-medi: Right. we don't need block time in redeem receipt anymore.

x/datapool/keeper/pool.go
inchori: Fixed in 6fb7f18 as well.

hansol-medi (approved):
LGTM 👍 
Thank you for your efforts!

x/datapool/keeper/pool.go
inchori: I'll remove `BlockHeight()`. Thanks!

PR #317, opened 2022-04-26
docs: update params & deposit rate in README.md

hansol-medi:
Update `README.md` according to the changes
- key of params
- addition of deposit rate


gyuguen (approved):
LGTM 👍

PR #318, opened 2022-04-27
feat: add distribute revenue pools

### LINKED ISSUE #278: Implement the pool activation
- Probably, using BeginBlocker
  - Iterate existing pools and check the current number of data 
  - If necessary, transit the pool status to ACTIVE.

TODO: Optimize performance
  - If all pools are proceeded in the BeginBlocker of a certain block, it would take a long time to process that block.

gyuguen:
Closes #278 
The following features have been added.
* `shareToken` has been deleted.
* The revenue distribution function is executed in accordance with the following conditions.
  * When a seller sells data
  * When a buyer buys a DataPass
* Seller receives a distribution of proceeds in the order in which data is sold.
* Improved to use commonly available test codes.


gyuguen:
I haven't created a function that distributes revenue to Curator yet.
I'll add that feature in this next PR.

x/datapool/keeper/pool_distribute.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+func (k Keeper) executeRevenueDistribute(ctx sdk.Context, pool *types.Pool) error {
+	availablePoolCoinAmount, err := k.getAvailablePoolCoinAmount(ctx, pool)
+	if err != nil {
+		return err
+	}
+
+	// calculate the revenue to be sent to each seller(totalSalesBalance / totalShareTokenAmount)
+	eachDistributionAmount := k.getEachDistributionAmount(pool)
```
gyuguen, line 161: Calculate the revenue to be distributed to each seller.

x/datapool/keeper/pool_distribute.go
```diff
@@ -0,0 +1,212 @@
...
+		paymentCoin := sdk.NewCoin(assets.MicroMedDenom, paymentAmount)
+		err = k.bankKeeper.SendCoins(
+			ctx,
+			poolAddress,
+			sellerAddr,
+			sdk.NewCoins(paymentCoin))
+		if err != nil {
+			return sdkerrors.Wrap(types.ErrRevenueDistribute, err.Error())
+		}
+		*salesInfo.PaidCoin = salesInfo.PaidCoin.Add(paymentCoin)
```
gyuguen, line 200: Store the distributed amount in the sales history.

x/datapool/types/keys.go
```diff
@@ -43,11 +53,28 @@ func GetKeyPrefixPools(poolID uint64) []byte {
 	return append(KeyPrefixPools, sdk.Uint64ToBigEndian(poolID)...)
 }
 
+func GetKeyPrefixDataValidateCerts(poolID, round *uint64) []byte {
```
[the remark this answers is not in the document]
gyuguen: Yes you are right
I was editing the code here and there and left a strange code.
I will fixed it soon.

gyuguen:
I have applied everything that has been discussed.

x/datapool/keeper/pool_distribute.go
```diff
@@ -0,0 +1,238 @@
...
+	if err != nil {
+		return sdkerrors.Wrap(types.ErrRevenueDistribute, err.Error())
+	}
+
+	salesHistories := k.ListSalesHistory(ctx, pool.PoolId, pool.Round)
+	for _, history := range salesHistories {
+		// if there is no coin available, proceed no further.
+		if availablePoolCoinAmount.Equal(sdk.NewInt(0)) {
+			break
+		}
```
hansol-medi, line 187: This check could be done out of the `salesHistories` loop, maybe after getting `availablePoolCoinAmount`.

x/datapool/keeper/pool_distribute.go
```diff
@@ -0,0 +1,238 @@
...
+		// if it has already been distributed, pass it.
+		if distributedAmount.Equal(paidCoinAmount) {
+			continue
+		}
+
+		paymentAmount := distributedAmount.Sub(paidCoinAmount)
+		// If the transferable amount is less than the payable amount, it is replaced with the transferable amount.
+		if availablePoolCoinAmount.LT(paymentAmount) {
+			paymentAmount = *availablePoolCoinAmount
+		}
```
hansol-medi, line 202: Theoretically, I think the `availablePoolCoinAmount` cannot be less than `paymentAmount` if the `availablePoolCoinAmount` is not zero. Do you have any idea of this case?

x/datapool/keeper/pool_distribute.go
```diff
@@ -0,0 +1,238 @@
...
+	totalNftPriceAmount := maxNftSupply.Mul(nftPriceAmount)
+
+	numIssuedNFTs := sdk.NewIntFromUint64(pool.NumIssuedNfts).ToDec()
+	targetNumData := sdk.NewIntFromUint64(pool.PoolParams.TargetNumData).ToDec()
+
+	distributeRate := numIssuedNFTs.Quo(maxNftSupply)
+	totalNftPriceAmount = totalNftPriceAmount.Mul(distributeRate)
+
+	return totalNftPriceAmount.Quo(targetNumData).TruncateInt()
+}
```
hansol-medi, line 238: This `getEachDistributionAmount()` can be simplified.
```
EachDistributionAmount = (NFTPrice * numIssuedNFTs) / targetNumData
```
which is the amount to be distributed per data.

hansol-medi (commented):
Thank you for quickly reflecting the corrections.
I left some minor comments.

x/datapool/keeper/pool_distribute.go
gyuguen: Oh, Thank you.
I used the previously developed check as it is.
Your comments look good.
I fixed it. (https://github.com/medibloc/panacea-core/pull/318/commits/da195915ca26f7e51121dc3cc674a287c06dcbdb)

x/datapool/keeper/pool_distribute.go
gyuguen: You're right!
In theory, this check is unnecessary.
Previously developed code was affected.
I fixed it. (https://github.com/medibloc/panacea-core/commit/da195915ca26f7e51121dc3cc674a287c06dcbdb)

x/datapool/keeper/pool_distribute.go
gyuguen: Oh! There was a simple way, but I implemented it in a complicated way.
Thank you. 😊
I fixed it. (https://github.com/medibloc/panacea-core/commit/da195915ca26f7e51121dc3cc674a287c06dcbdb)

x/datapool/keeper/pool_distribute.go
hansol-medi: As you mentioned in TODO, BTW, we may have to discuss about error handling of this revenue distribution. I'll think about it too.

gyuguen: Yes. We need a discussion.
Since this is an operation done in the endBlocker of the block chain, if an error occurs, block creation cannot be performed.
(It is likely that blocks will not continue to be created.)
I will discuss a few things soon.

hansol-medi (approved):
It looks good to me 👍

x/datapool/keeper/pool_distribute.go
```diff
@@ -0,0 +1,227 @@
...
+	return &delayedRevenueDistribute
+}
+
+// addInstantRevenueDistribute adds the poolID to distribute. If there are duplicate poolIDs, they are not added.
+func (k Keeper) addInstantRevenueDistribute(ctx sdk.Context, poolID uint64) error {
+	instantRevenueDistribute := k.GetInstantRevenueDistribute(ctx)
+	instantRevenueDistribute.AppendPoolID(poolID)
+	k.SetInstantRevenueDistribute(ctx, instantRevenueDistribute)
+	return nil
+}
```
inchori, line 90: Why this method return `error`? I think that this method has no return values.

gyuguen: Yes, you're right.
I fixed it. (https://github.com/medibloc/panacea-core/pull/318/commits/dfe9a6ca02ba0b18b81276c02872e62313815e5f)

inchori (approved):
LGTM 👍

proto/panacea/datapool/v2/pool.proto
```diff
@@ -19,6 +19,7 @@ message Pool {
   string                   curator = 8;
   cosmos.base.v1beta1.Coin deposit = 9 [(gogoproto.nullable) = false]; // deposit
   string                   nft_contract_addr = 10;                     // NFT contract address of pool
+  bool                     is_paid_deposit = 11;
```
youngjoon-lee, line 22: Does this mean that the curator was paid the deposit to the pool? Or, that the deposit was returned to the curator?

If the latter, it would be clearer to rename this to `was_deposit_returned`.

x/datapool/keeper/pool_distribute.go
```diff
@@ -0,0 +1,226 @@
...
+		}
+	}
+	// delete this pool from instantRevenueDistribute
+	instantRevenueDistribute.RemovePreviousIndex(lastIndex)
+	k.SetInstantRevenueDistribute(ctx, instantRevenueDistribute)
+
+	return nil
+}
+
+// sendDepositToCurator returns to the Curator if the pool status is ACTIVE but the deposit has not yet been returned.
```
youngjoon-lee, line 129:
```suggestion
// sendDepositToCurator returns the deposit to the Curator if the pool status is ACTIVE but the deposit has not yet been returned.
```

proto/panacea/datapool/v2/pool.proto
gyuguen: Yes, you're right.
I will change this name soon.

x/datapool/keeper/pool_distribute.go
```diff
@@ -0,0 +1,226 @@
+package keeper
```
youngjoon-lee, line 1: It would be better to rename this file to `pool_distribution.go` or `pool_dist.go` (as a noun).
Also, please rename all other `*Distribute`s to `*Distribution`.

x/datapool/types/pool.go
```diff
@@ -44,8 +41,35 @@ func GetModuleAddress() sdk.AccAddress {
...
+		}
+	}
+	d.PoolIds = append(d.PoolIds, poolID)
+}
+
+func (d *InstantRevenueDistribute) IsEmpty() bool {
+	return d.PoolIds == nil || len(d.PoolIds) == 0
+}
+
+func (d *InstantRevenueDistribute) RemovePreviousIndex(idx int) {
```
youngjoon-lee, line 59: I would recommend to rename this to `TruncateFromBeginning(count int)`. I didn't get that the `RemovePreviousIndex` means before reading this function body.

x/datapool/keeper/pool_distribute.go
youngjoon-lee: please rename this to `getDistributionPerSeller`.

youngjoon-lee (approved):
lgtm

x/datapool/keeper/pool_distribute.go
```diff
@@ -0,0 +1,226 @@
...
+	instantRevenueDistribute := k.GetInstantRevenueDistribute(ctx)
+
+	if instantRevenueDistribute.IsEmpty() {
+		return nil
+	}
+
+	// This lastIndex is to execute the following iteration in the next block if it does not end in one block.
+	// But We're currently processing it all in one block. :)
+	lastIndex := 0
+	// TODO We need to think about how to deal with the pool where distribution has failed.
```
youngjoon-lee, line 101: From my understanding, if the BeginBlocker panics, the state machine is halted. So, we don't need to worry about how to rollback operations that succeeded partially.
But, we need to test it.

gyuguen: Yes, we will proceed as you understands.
As you said, we need to test this.

x/datapool/keeper/pool_distribute.go
gyuguen: Thank you. I have changed to your recommendation.
https://github.com/medibloc/panacea-core/pull/318/commits/b398ff659c3333c09e78ccf4def88f4dd20dc9f2

x/datapool/types/pool.go
gyuguen: OK.
I fixed it.
https://github.com/medibloc/panacea-core/pull/318/commits/b398ff659c3333c09e78ccf4def88f4dd20dc9f2

x/datapool/keeper/pool_distribute.go
gyuguen: OK. I fixed it.
https://github.com/medibloc/panacea-core/pull/318/commits/5438b937521dcc52e21db57680c0c0b560d7dcce

proto/panacea/datapool/v2/pool.proto
gyuguen: I fixed it.
[change is_paid_deposit to was_deposit_returned](https://github.com/medibloc/panacea-core/pull/318/commits/ecf72ead29d4c0a2cb027b147c8648b03b924d56)

x/datapool/keeper/pool_distribution.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+// addInstantRevenueDistribution adds the poolID to distribution. If there are duplicate poolIDs, they are not added.
+func (k Keeper) addInstantRevenueDistribution(ctx sdk.Context, poolID uint64) {
+	instantRevenueDistribution := k.GetInstantRevenueDistribution(ctx)
+	instantRevenueDistribution.AppendPoolID(poolID)
+	k.SetInstantRevenueDistribution(ctx, instantRevenueDistribution)
+}
+
+func (k Keeper) DistributionRevenuePools(ctx sdk.Context) error {
```
hansol-medi, line 104: We use verb for name of funcs.
So, I think that verb is proper for this func as well, `DistributeRevenuePools()` or `DistributePoolsRevenue()`.

hansol-medi (approved):
LGTM

youngjoon-lee (approved):
thanks

PR #319, opened 2022-04-27
feat: added genesis function for revenue distribution

gyuguen:


go.mod
```diff
@@ -3,107 +3,37 @@ module github.com/medibloc/panacea-core/v2
 go 1.15
 
 require (
```
gyuguen, line 5: Well, in my idea, all the deleted dependencies are not needed.
I think if there are no problems, we can continue as it is.

gyuguen:
I applied what we discussed offline here.

inchori (approved):
LGTM 👍

x/datapool/genesis.go
```diff
@@ -23,6 +23,13 @@ func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState)
 		k.SetPool(ctx, &pool)
 	}
 
+	k.SetInstantRevenueDistribute(ctx, &genState.InstantRevenueDistribute)
+
+	for _, history := range genState.SalesHistories {
+		key := types.GetKeyPrefixSalesHistory(history.PoolId, history.Round, history.SellerAddress)
+		k.SetSalesHistoryByKey(ctx, key, history)
```
hansol-medi, line 30: I think you can use `k.SetSalesHistory()` itself.
```suggestion
		k.SetSalesHistory(ctx, history)
```

x/datapool/keeper/pool_distribute.go
```diff
@@ -32,6 +32,11 @@ func (k Keeper) SetSalesHistory(ctx sdk.Context, salesHistory *types.SalesHistor
 		salesHistory.Round,
 		salesHistory.SellerAddress,
 	)
+	k.SetSalesHistoryByKey(ctx, key, salesHistory)
+}
+
+// SetSalesHistoryByKey stores sales history.
+func (k Keeper) SetSalesHistoryByKey(ctx sdk.Context, key []byte, salesHistory *types.SalesHistory) {
```
hansol-medi, line 39: I just wonder why you made this function splitted which stores history by key. Is there any specific reason?!

hansol-medi (approved):
LGTM

youngjoon-lee (approved):
good

x/datapool/genesis.go
gyuguen: Thank you!

x/datapool/keeper/pool_distribute.go
gyuguen: It is a residue created while editing the code.
I removed this function. :)

PR #323, opened 2022-05-03
feat: Add Genesis for Data Pass Redeem Receipts

inchori:
- Add genesis for data pass redeem receipts.
- Implement `GetAllDataPassRedeemReceipts` in `keeper/pool.go`.


x/datapool/genesis_test.go
```diff
@@ -69,6 +77,11 @@ func (suite genesisTestSuite) TestDataPoolInitGenesis() {
 	// check params
 	paramsFromKeeper := suite.DataPoolKeeper.GetParams(suite.Ctx)
 	suite.Require().Equal(params, paramsFromKeeper)
+
+	// check data pass redeem receipt
+	redeemReceiptFromKeeper, err := suite.DataPoolKeeper.GetDataPassRedeemReceipt(suite.Ctx, poolID, 1, redeemer)
+	suite.Require().NoError(err)
+	suite.Require().Equal(dataPassRedeemReceipt, redeemReceiptFromKeeper)
```
hansol-medi, line 84: Since this is the test for `InitGenesis()`, it would be better to check all the data pass redeem receipts are equal.
In other words, the `redeemReceiptFromKeeper` may be the result of `k.GetAllDataPassRedeemReceipts()`, and it should be equal to the `dataPassRedeemReceipts` above.

inchori: Oh Thanks. I have a question. Is we need to apply for `pool` and `params` later?

hansol-medi: right. I think it would also be good to check all pools are equal.

inchori: Thanks, I just leave these comments, not resolved.

gyuguen (approved):
LGTM 👍 
I left one comment.(not importanted.)

proto/panacea/datapool/v2/genesis.proto
```diff
@@ -12,6 +12,7 @@ message GenesisState {
   uint64                 next_pool_number = 2;
   repeated Pool          pools = 3 [(gogoproto.nullable) = false];
   Params                 params = 4 [(gogoproto.nullable) = false];
+  repeated DataPassRedeemReceipt data_pass_redeem_receipts = 5 [(gogoproto.nullable) = false];
```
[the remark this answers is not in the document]
inchori: Does it mean sort like this?

```proto
message GenesisState {
  repeated DataValidator         data_validators = 1 [(gogoproto.nullable) = false];
  uint64                         next_pool_number = 2;
  repeated Pool                  pools = 3 [(gogoproto.nullable) = false];
  Params                         params = 4 [(gogoproto.nullable) = false];
  repeated DataPassRedeemReceipt data_pass_redeem_receipts = 5 [(gogoproto.nullable) = false];
}
````
or sort by other thing?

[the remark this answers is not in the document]
gyuguen: Oh, I typed it wrong.
Yes, You are right.

PR #326, opened 2022-05-18
refactor: rename nft_id to data_pass_id

hansol-medi:
According to the name of `dataPass`, renamed the variable `nft_id` to `data_pass_id`


x/datapool/client/cli/queryPool.go
```diff
@@ -122,9 +122,9 @@ func CmdGetDataValidationCertificates() *cobra.Command {
 
 func CmdGetDataPassRedeemReceipt() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "get-data-pass-receipt [poolID] [round] [nftID]",
+		Use:   "get-data-pass-receipt [poolID] [round] [dataPassID]",
```
inchori, line 125: I think we need to use underscore case `data_pass_ID`, not camel case. In other cosmos tx commands, they use underscore.

hansol-medi: Thank you for pointing it out!
It seems that dash(-) is used for arg in cosmos sdk, so why don't we use dash too? (e.g. [`x/gov` query cmd](https://github.com/cosmos/cosmos-sdk/blob/bcf20889255a7f7364c9aa54b29da7b47587a1b9/x/gov/client/cli/query.go#L185))

```
Use:   "get-data-pass-receipt [pool-id] [round] [data-pass-id]",
```

hansol-medi: It is changed in [31a0f59](https://github.com/medibloc/panacea-core/pull/326/commits/31a0f598d3ae99d3e831fb257334af5e6fbe3482)

Furthermore, I changed all the cmd args for consistency. 
And I removed `get-` prefix in query cmd because everyone know that query cmd gets something. The 'get-' prefix is not used in the cosmos sdk too.

gyuguen (approved):
LGTM 👍 
Thank you.

youngjoon-lee (approved):
good

PR #328, opened 2022-05-23
feat: add data pass redeem history query

hansol-medi:
In current kv store of redeem receipts, we cannot retrieve data pass redeem receipts by redeemer's address. Because the key is composed of poolID, round, and nftID, not redeemer's address.

With the need of getting receipts by redeemer's address in data validator, I made another kv store that the key is composed of redeemer's address and poolID. When redeemer redeems data pass, the redeem history is also stored/updated (`k.appendDataPassRedeemHistory()`). 

I implemented a query of the redeem history by redeemer's address and pool ID as well.

What I've done

- [x] keeper functions for storing/updating redeem history
- [x] keeper functions for querying redeem history
- [x] query cmd
- [x] Init/Export genesis
- [x] test codes


x/datapool/client/cli/queryPool.go
```diff
@@ -171,7 +171,7 @@ func CmdGetDataPassRedeemReceipts() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "data-pass-redeem-receipts [poolID]",
 		Short: "Query data pass redeem receipts by pool",
-		Args:  cobra.ExactArgs(2),
+		Args:  cobra.ExactArgs(1),
```
hansol-medi, line 174: This cmd seems to accept 1 arg, but was 2. So I fixed it. Please check @inchori

x/datapool/keeper/pool.go
```diff
@@ -615,15 +617,86 @@ func (k Keeper) GetDataPassRedeemReceipt(ctx sdk.Context, poolID, round, nftID u
 	return receipt, nil
 }
 
-func (k Keeper) SetDataPassRedeemReceipt(ctx sdk.Context, redeemReceipt types.DataPassRedeemReceipt) error {
+func (k Keeper) SetDataPassRedeemReceipt(ctx sdk.Context, redeemReceipt types.DataPassRedeemReceipt) {
```
hansol-medi, line 620: This keeper function doesn't need to return error, so I fixed. Please check this as well, @inchori

inchori: Thanks for fixing these codes! 🤩

gyuguen (approved):
LGTM 👍 
Later, if you add a query that retrieves history as a redeemer, you can check the history of all pools of the redemeer.

inchori (approved):
LGTM 👍

youngjoon-lee (approved):
gut

PR #331, opened 2022-05-30
refactor: refactoring deal model & bug fix

hansol-medi:
I ask for your understanding of big changes. But mostly they are variable name changes.
Aside from that, I've done refactoring the codes consistent style with those in `x/datapool` module. 
Main changes are the below:

- change index separator from `[]byte{0x07}` to `[]byte{0xFF}`
- moving constants (about deal status) and `NewDeal` func to types package
- removal of the things related to deal address (`AccDealAddressFromBech32`, `VerifyDealAddressFormat`). plus removal `v044_temp` 
  - I thought we don't have to use the address scheme that osmosis used. So changed `NewDealAddress` function as used in `x/datapool`
- fix isTrustedValidator func
  - If there is no trusted data validator in deal, the `isTrustedValidator` should return true. (It means trust any data validators)

If you have any opinions or advices, please feel free to leave comments 🙏


inchori (approved):
LGTM 👍 Thanks for refactoring codes and fixing bugs!

youngjoon-lee (approved):
great!

x/datadeal/keeper/deal.go
```diff
@@ -77,37 +56,44 @@ func (k Keeper) SetNextDealNumber(ctx sdk.Context, dealNumber uint64) {
...
-		err := k.cdc.UnmarshalBinaryLengthPrefixed(bz, &val)
-		if err != nil {
-			panic(err)
-		}
+	val := gogotypes.UInt64Value{}
 
-		dealNumber = val.GetValue()
+	err := k.cdc.UnmarshalBinaryLengthPrefixed(bz, &val)
+	if err != nil {
+		return 0, nil
```
gyuguen, line 73:
```suggestion
		return 0, err
```

x/datadeal/keeper/deal.go
```diff
@@ -150,79 +136,82 @@ func (k Keeper) ListDeals(ctx sdk.Context) ([]types.Deal, error) {
...
+	return store.Has(dataCertKey)
 }
 
-func (k Keeper) isTrustedValidator(cert types.DataValidationCertificate, findDeal types.Deal) error {
+func (k Keeper) isTrustedDataValidator(cert types.DataValidationCertificate, deal types.Deal) bool {
 	validator := cert.UnsignedCert.GetDataValidatorAddress()
-	trustedValidators := findDeal.GetTrustedDataValidators()
+	trustedValidators := deal.GetTrustedDataValidators()
+
+	if len(trustedValidators) == 0 || trustedValidators == nil {
```
gyuguen, line 204: I think if trustedValidators are nil, len(trustedValidators) will return 0.

```suggestion
	if len(trustedValidators) == 0 {
```

x/datadeal/keeper/deal.go
```diff
@@ -305,16 +290,16 @@ func (k Keeper) VerifyDataCertificate(ctx sdk.Context, validatorAddr sdk.AccAddr
 		return false, sdkerrors.Wrapf(types.ErrInvalidSignature, "%s", cert.GetSignature())
```
gyuguen, line 290: Above this we are using the `isValid` variable like this:
```go
        isValid := validatorPubKey.VerifySignature(unSignedMarshaled, cert.GetSignature())
	if !isValid {
		return false, sdkerrors.Wrapf(types.ErrInvalidSignature, "%s", cert.GetSignature())
	}
```

If you're not going to reuse the `isValid` variable, how about changing it to something like this:
```go
	if !validatorPubKey.VerifySignature(unSignedMarshaled, cert.GetSignature()){
		return false, sdkerrors.Wrapf(types.ErrInvalidSignature, "%s", cert.GetSignature())
	}
```

gyuguen (approved):
Thank you!
I left a few comment.

x/datadeal/keeper/deal.go
hansol-medi: Good point. Thanks!

hansol-medi:
> I left a few comment.

Thanks for detailed review. I've applied your suggestions!

gyuguen (approved):
LGTM 👍

PR #341, opened 2022-06-02
feat: Create a x/oracle module

inchori:
- Create `x/oracle` module.
- This PR is quite big because the all of the existing transaction and query were located in `x/datapool` module.
- When querying an oracle, the command is `panacead q oracle oracle` so that I changed to 
`panacead q oracle oracle-node {addr}`.

#### Needs to be Add ####
- When updating an oracle node, it must be validated that the registered node and requester is same.


x/datapool/types/keys.go
```diff
@@ -26,9 +26,6 @@ const (
 )
 
 var (
-	// KeyPrefixOracles defines key to store oracle
-	KeyPrefixOracles = []byte{0x01}
```
hansol-medi, line 30: We can reduce all the keys by 1, but it's optional

x/oracle/genesis_test.go
```diff
@@ -0,0 +1,57 @@
...
+	oracles = append(oracles, tempOracle)
+
+	genState := &types.GenesisState{
+		Oracles: oracles,
+	}
+
+	oracle.InitGenesis(suite.Ctx, suite.OracleKeeper, *genState)
+
+	// check oracle
+	oracleFromKeeper, err := suite.OracleKeeper.GetOracle(suite.Ctx, oracle1)
```
hansol-medi, line 37: Let's check by getting all the oracles from the keeper and its length to be equal to 1.

x/oracle/keeper/oracle.go
```diff
@@ -0,0 +1,104 @@
...
+	if store.Has(oracleKey) {
+		return sdkerrors.Wrapf(types.ErrOracleAlreadyExist, "the oracle %s is already exists", oracle.Address)
+	}
+
+	oraclePubKey, err := k.accountKeeper.GetPubKey(ctx, oracleAddr)
+	if err != nil {
+		return err
+	}
+	if oraclePubKey == nil {
+		return sdkerrors.ErrKeyNotFound
```
hansol-medi, line 26: Instead of returning `sdkerrors.ErrKeyNotFound`, let's return other error. Because it's an error when the key doesn't exist, not pub key. 

Apart from that, is it required that the pub key of oracle exists? IMHO, it was required for the past because the seller should encrypt data with the oracle pub key. For now, however, we use TLS communication which is decrypted only in SGX so I think it is not required anymore.

x/datapool/keeper/pool.go
```diff
@@ -217,6 +115,12 @@ func (k Keeper) CreatePool(ctx sdk.Context, curator sdk.AccAddress, deposit sdk.
 	return newPool.GetPoolId(), nil
 }
 
+func (k Keeper) isRegisteredOracle(ctx sdk.Context, oracleAddress sdk.AccAddress) bool {
+	store := ctx.KVStore(k.storeKey)
```
hansol-medi, line 119: When checking the oracle is registered or not, we have to use `x/oracle` keeper, not `x/datapool`. Because we store the info of oracle in `x/oracle`. So let's import the `x/oracle` module into the `x/datapool` keeper and use it. 

For that, you may have to add `x/oracle` keeper in `x/datapool` keeper struct (`x/datapool/keeper/keeper.go`).
And then, you can check it by `k.oraclekeeper.isRegisteredOracle()`.

hansol-medi (commented):
Thank you very much. As you said, it's a very big deal to make another whole module.
I left a few comments.

x/oracle/keeper/oracle.go
hansol-medi: Ah, sorry please forget this comment. It is required to verify the signature in `DataCert`. Let's keep it this way :)

gyuguen (approved):
LGTM 👍

x/datapool/types/keys.go
inchori: Does it mean reduce all the keys by 1 in `x/datapool` module?

x/datapool/keeper/pool.go
inchori: Without this comment, I resolved other comments in [658d72e](https://github.com/medibloc/panacea-core/pull/341/commits/658d72e32e99079febba08d26e022865fc500408).

x/datapool/types/keys.go
hansol-medi: yes right!

x/datapool/keeper/pool.go
inchori: You can check this discussion in [bebca0b](https://github.com/medibloc/panacea-core/pull/341/commits/bebca0b48d2680429c31a5c2dbc1eb28915309d0).

inchori: Thanks for resolving it!

hansol-medi (approved):
It looks good to me. Thanks!
After merging this PR, we have to update grpc query in panacea-oracle too.

hansol-medi:
@inchori, I found a word of `data validator` in errors.go, so changed it.

PR #345, opened 2022-06-07
feat: Add Query for Single Data Cert and List of Data Cert

inchori:
- Add cmd and keeper for query a single data cert and list of data cert.


x/datadeal/keeper/grpc_query_deal.go
```diff
@@ -18,8 +21,55 @@ func (k Keeper) Deal(goCtx context.Context, req *types.QueryDealRequest) (*types
...
 }
+
+func (k Keeper) DataCert(goCtx context.Context, req *types.QueryDataCertRequest) (*types.QueryDataCertResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid request")
+	}
+
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	dataCert, err := k.GetDataCert(ctx, req.DealId, []byte(req.DataHash))
```
hansol-medi, line 37: We may have to decode the string using base64 package. 
```suggestion
        bz, err := base64.StdEncoding.DecodeString(req.DataHash)
	if err != nil {
		return nil, err
	}
	dataCert, err := k.GetDataCert(ctx, req.DealId, bz)
```

gyuguen (approved):
LGTM 👍

hansol-medi (approved):
LGTM 👍

inchori:
Query the list of data certificate has something wrong when I did manual test. So I'll convert to draft.

PR #351, opened 2022-06-09
fix: Fix Oracle Genesis State

inchori:
- I just fix genesis state of oracle from array to map data structure.


gyuguen:
Is there any reason to change array to map?

inchori:
Hmm.. There has not special reason for changing it. But I just want to validate genesis state in `types/genesis.go`. If there has no special reason for changing it. I'll close this PR!

gyuguen:
In my opinion, it is unnecessary because it converts an array into a map and exports it, and converts the map to an array when re-importing it.
So, I think it seems to use an array as it is.

inchori:
Thanks for comments! So I'll change to validate using array of oracle genesis state.

x/oracle/types/genesis.go
```diff
@@ -10,6 +14,10 @@ func DefaultGenesis() *GenesisState {
 // Validate performs basic genesis state validation returning an error upon any
 // failure.
 func (gs GenesisState) Validate() error {
-
+	for i, oracle := range gs.Oracles {
+		if gs.Oracles[i].Address != oracle.Address {
+			return sdkerrors.Wrapf(ErrInvalidGenesisOracle, "oracle address: %s", oracle.GetAddress())
+		}
+	}
 	return nil
```
youngjoon-lee, line 22: Hm. I think this logic always returns true.

gyuguen: Umm. What part of genesis are you trying to validate?
Is the address valid?
Or is it a duplicate stored oracle address?

inchori: The address. So I changed to map for genesis state. But the this logic always returns true. 😂 
So I can change like this.
```go
func (gs GenesisState) Validate() error {
	for _, oracle := range gs.Oracles {
		oracleAddr, err := sdk.AccAddressFromBech32(oracle.Address)
		
		if err != nil {
			return sdkerrors.Wrapf(ErrInvalidGenesisOracleAddr, "oracle address: %s", oracleAddr.String())
		}
	}
	return nil
}
```
How about this?

youngjoon-lee: cd3b9e1

gyuguen (approved):

hansol-medi (approved):

PR #353, opened 2022-06-14
feat: fix upgradeHander issue

gyuguen:
We ran into an error situation during the Testnet deployment.
```
Jun 10 03:09:17 testnet-sentry panacead[3707341]: panic: cannot delete latest saved version (12)
Jun 10 03:09:17 testnet-sentry panacead[3707341]: goroutine 1 [running]:
Jun 10 03:09:17 testnet-sentry panacead[3707341]: github.com/cosmos/cosmos-sdk/store/rootmulti.(*Store).pruneStores(0xc0001cc460)
Jun 10 03:09:17 testnet-sentry panacead[3707341]:         /home/ubuntu/go/pkg/mod/github.com/medibloc/cosmos-sdk@v0.42.11-panacea.1/store/rootmulti/store.go:388 +0x26e
```

This error occurs when the following code is excuted.
https://github.com/cosmos/cosmos-sdk/blob/v0.42.11/store/rootmulti/store.go#L380

CosmosHub has solved this related problem as follows.
https://github.com/cosmos/gaia/blob/v5.0.8/app/app.go#L428

We also modified it with reference to this.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -789,7 +789,7 @@ func initParamsKeeper(appCodec codec.BinaryMarshaler, legacyAmino *codec.LegacyA
 	return paramsKeeper
 }
 
-// registerUpgradeHandlers registers upgrade handlers, and sets the storetypes loader if necessary.
+// registerUpgradeHandlers registers upgrade handlers, and sets the store loader if necessary.
 // This function must be called before sealing the BaseApp (i.e. by app.LoadLatestVersion())
 // because the storetypes loader cannot be set if BaseApp is already sealed.
 func (app *App) registerUpgradeHandlers() error {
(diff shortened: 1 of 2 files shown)


inchori (approved):
LGTM 👍

app/app.go
```diff
@@ -785,7 +789,10 @@ func initParamsKeeper(appCodec codec.BinaryMarshaler, legacyAmino *codec.LegacyA
 	return paramsKeeper
 }
 
-func (app *App) registerUpgradeHandlers() {
+// registerUpgradeHandlers registers upgrade handlers, and sets the storetypes loader if necessary.
```
youngjoon-lee, line 792:
```suggestion
// registerUpgradeHandlers registers upgrade handlers, and sets the store loader if necessary.
```
Sorry. Could you fix this typo?

gyuguen: Thank you. This is probably what I changed wrong.

youngjoon-lee (approved):
danke

PR #355, opened 2022-06-17
feat: added cipher text decrypt function

### LINKED ISSUE #354: Added cipherText decrypt function
We decrypt the cipherText using the following command:
```
panacead decrypt encrypted_data.txt --from {key}
```

You must register the key through the command below.
```
panacead keys add {key}
```

gyuguen:
Closes #354 
The following features have been added.
* Added `panacead decrypt (--text {cipherText} | --path {cipherText file path}
* You can decrypt the encrypted text with the ECDSA-based publicKey using the privateKey stored in your local store.

### DIFF
--- a/client/decrypt/decrypt.go
+++ b/client/decrypt/decrypt.go
@@ -1,0 +1,102 @@
+package decrypt
+
+import (
+	"encoding/base64"
+	"encoding/hex"
+	"fmt"
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/cosmos/cosmos-sdk/crypto/keyring"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/tendermint/btcd/btcec"
+	"github.com/tendermint/tendermint/libs/cli"
+	"io/ioutil"
+)
+
+const (
+	FlagCipherText     = "text"
+	FlagCipherTextPath = "path"
+)
+
+func Command(defaultNodeHome string) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "decrypt [name] (--text [ciphertext] | --path [ciphertext-file])",
+		Short: "Decrypt and output the ciphertext file encrypted with ECDSA PublicKey.",
+		Long: `This command can decrypt ciphertext encrypted with ECDSA PublicKey. 
+The file contents must be Standard Base64 encoded.
+Also, make sure your key is stored in the localStore.
+(According to the following command (panacead keys add ...)
+`,
+		Args: cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx := client.GetClientContextFromCmd(cmd)
+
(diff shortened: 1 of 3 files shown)


client/decrypt/decrypt.go
```diff
@@ -0,0 +1,102 @@
+package decrypt
```
gyuguen, line 1: I'm not sure if this name would be good.

`crypto/decrypt`?
`xxx/decrypt`?
difficult!

client/decrypt/decrypt.go
```diff
@@ -0,0 +1,102 @@
...
+	FlagCipherText     = "text"
+	FlagCipherTextPath = "path"
+)
+
+func Command(defaultNodeHome string) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "decrypt [name] (--text [ciphertext] | --path [ciphertext-file])",
+		Short: "Decrypt and output the ciphertext file encrypted with ECDSA PublicKey.",
+		Long: `This command can decrypt ciphertext encrypted with ECDSA PublicKey. 
+The file contents must be Standard Base64 encoded.
```
youngjoon-lee, line 27: I think it would be better to not force users to prepare base64-encoded files. In my opinion, the file needs to contain a plain byte stream that is encrypted, so that users who have files that are not base64-encoded can use this tool.
If the user has a base64-encoded encrypted file, he/she can this tool as below:
```bash
cat my-b64-enc-file.txt | base64 -d > my-enc-file.txt
panacead decrypt my-key --path my-enc-file.txt
```

inchori:
Thank you for your effort!
Does we not need to add `flags.FlagChainID`?

client/decrypt/decrypt.go
gyuguen: OK. I understand.
I will fix it soon.

gyuguen: I fixed it.
https://github.com/medibloc/panacea-core/pull/355/commits/7b214daa7fbdfdd02198eb42c689c280d1128154

inchori (approved):
LGTM 👍

youngjoon-lee (approved):
Thank you!

client/decrypt/decrypt.go
hansol-medi: Yep it looks good to me too!

client/decrypt/decrypt.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/cosmos/cosmos-sdk/crypto/keyring"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/tendermint/tendermint/libs/cli"
+	"io/ioutil"
+)
+
+const (
+	FlagCipherText     = "text"
```
hansol-medi, line 17: Why don't we use "cipher-text"? Although there is no flag under "text", the word "text" is used in output flag. But it's not that important

hansol-medi (approved):
utACK

client/decrypt/decrypt.go
gyuguen: That's good idea.
I fixed it.
Thank you :)

PR #357, opened 2022-06-30
fix: update Dockerfile for wasmvm v1.0.0-beta7

hansol-medi:
Currently, github action of `build and push docker image` is failed.
I missed Dockerfile when upgrading cosmwasm version, so I updated.


inchori (approved):
LGTM 👍

gyuguen (approved):
LGTM 👍

PR #360, opened 2022-07-18
feat: configure datadeal module's `v2alpha2` proto files

inchori:
- Configure `datadeal` module's proto file.

PR #361, opened 2022-07-19
feat: add proto for `v2alpha2` oracle module 

hansol-medi:
Proto files for ver `v2alpha2` of oracle module are added.

When proto-gen, pb.go files from this ver of proto files are also generated under `types` package, which is the same with `v2alpha1`. So I modified `protocgen.sh` to ignore `v2alpha2` temporarily. If you have better idea for this, please feel free to leave comments :)


scripts/protocgen.sh
```diff
@@ -17,7 +17,7 @@ protoc_gen_gocosmos() {
 
 protoc_gen_gocosmos
 
-proto_dirs=$(find ./proto -path -prune -o -name '*.proto' -print0 | xargs -0 -n1 dirname | sort | uniq)
+proto_dirs=$(find ./proto -path -prune -o -name '*.proto' -not -path '*/v2alpha2/*' -print0 | xargs -0 -n1 dirname | sort | uniq)
```
inchori, line 20: Thanks for making this script! For now, I'll apply this script in the branch #360 as well.

proto/panacea/oracle/v2alpha2/oracle.proto
```diff
@@ -0,0 +1,34 @@
...
+
+import "gogoproto/gogo.proto";
+import "google/protobuf/timestamp.proto";
+
+// OracleStatus enumerates the status of oracle.
+enum OracleStatus {
+  option (gogoproto.goproto_enum_prefix) = false;
+
+  // ORACLE_STATUS_UNSPECIFIED
+  ORACLE_STATUS_UNSPECIFIED = 0;
```
gyuguen, line 14: Does this status mean that the Oracle is voting?

gyuguen: And I wonder if this state is maintained even in case of registration failure.

proto/panacea/oracle/v2alpha2/query.proto
```diff
@@ -0,0 +1,108 @@
...
+  // Other oracle can validate whether the oracle set correct trusted block info.
+  int64 trusted_block_height = 5;
+  bytes trusted_block_hash = 6;
+  // Encrypted oracle private key by node public key, which can only be decrypted in SGX securely.
+  bytes encrypted_oracle_priv_key = 7;
+  // The status of OracleRegistration.
+  OracleRegistrationStatus status = 8;
+  VotingPeriod voting_period = 9;
+  TallyResult tally_result = 10;
+}
```
gyuguen, line 108: I think it is better to move this message to `oracle.proto`.
Because this message is stored in the OracleRegistration Store as it is.
What do you think?

proto/panacea/oracle/v2alpha2/tx.proto
```diff
@@ -0,0 +1,73 @@
...
+  rpc RegisterOracle(MsgRegisterOracle) returns (MsgRegisterOracleResponse);
+
+  // VoteOracleRegistration  defines a method for voting registration of oracle.
+  rpc VoteOracleRegistration(MsgVoteOracleRegistration) returns (MsgVoteOracleRegistrationResponse);
+}
+
+// MsgRegisterOracle defines the Msg/RegisterOracle request type.
+message MsgRegisterOracle {
+  OracleRegistrationInfo oracle_registration_info = 1;
+}
```
gyuguen, line 22: How about deleting OracleRegistrationInfo and using it like below?

```suggestion
message MsgRegisterOracle {
  string unique_id = 1;
  string oracle_address = 2;
  bytes node_pub_key = 3;
  bytes node_pub_key_remote_report = 4;
  int64 trusted_block_height = 5;
  bytes trusted_block_hash = 6;
}
```

gyuguen (commented):
Thank you!
I left a few comment :)

proto/panacea/oracle/v2alpha2/oracle.proto
hansol-medi: > Does this status mean that the Oracle is voting?

When we define enum in proto, we are forced to use index 0 as a status of `UNSPECIFIED`. If not, proto lint fail.
As you can see in all the status enum, 0 index is used for `UNSPECIFIED`.

> And I wonder if this state is maintained even in case of registration failure.

This status of `OracleStatus` is for registered oracle. As I understand, when oracle is failed to be registered, the oracle would not be stored in `Oracle` KV store. But it may be stored in `OracleRegistration` KV store with status of `FAILURE`. Do I understand correctly?

proto/panacea/oracle/v2alpha2/query.proto
hansol-medi: Good. I think so. I'll move it.

proto/panacea/oracle/v2alpha2/oracle.proto
gyuguen: > > Does this status mean that the Oracle is voting?
> 
> When we define enum in proto, we are forced to use index 0 as a status of `UNSPECIFIED`. If not, proto lint fail. As you can see in all the status enum, 0 index is used for `UNSPECIFIED`.

Aha. I got it. Thank you.

gyuguen: > > And I wonder if this state is maintained even in case of registration failure.
> 
> This status of `OracleStatus` is for registered oracle. As I understand, when oracle is failed to be registered, the oracle would not be stored in `Oracle` KV store. But it may be stored in `OracleRegistration` KV store with status of `FAILURE`. Do I understand correctly?

Ah, I thought I would also store it in the Oracle KV store, but from what you said, it doesn't seem necessary.
I want it to be like this now.

proto/panacea/oracle/v2alpha2/tx.proto
hansol-medi: We can delete and modify as you suggested.

However, the reason why I make a message `OracleRegistrationInfo` is it would be useful when we handle the `MsgRegisterOracle` in msg server(in `msg_server_oracle.go`).

proto/panacea/oracle/v2alpha2/oracle.proto
hansol-medi: Thanks for asking!
I also thought as you did, but I realized when defining some statuses.

proto/panacea/oracle/v2alpha2/query.proto
hansol-medi: resolved in [ec4c403](https://github.com/medibloc/panacea-core/pull/361/commits/ec4c4032b41a083dd1d5c50ed5b6e8e2d2f01f00)

proto/panacea/oracle/v2alpha2/tx.proto
hansol-medi: I got the point that @gyuguen said. I fixed it in [08c9c9f](https://github.com/medibloc/panacea-core/pull/361/commits/08c9c9f9b88dc30e411e48adfec488a1753f489c)

gyuguen (approved):
LGTM 👍

scripts/protocgen.sh
gyuguen: I think you should change it below. :)
```suggestion
proto_dirs=$(find ./proto -path -prune -o -name '*.proto' -not -path '*/v2alpha1/*' -print0 | xargs -0 -n1 dirname | sort | uniq)
```

gyuguen: I think you should change it below :)
```suggestion
proto_dirs=$(find ./proto -path -prune -o -name '*.proto' -not -path '*/v2alpha1/*' -print0 | xargs -0 -n1 dirname | sort | uniq)
```

We need to */alpha2/* packages.

gyuguen: And if you run proto-gen after making this change, there are a few things that need to be corrected.

hansol-medi: You're right.
As I mentioned in description, `proto-gen` also generates pb.go files under the same package.
So this change was for success of `build-all` and `test` in github actions.

When we develop doracle, version of `v2alpha2`, we should change the `protogen.sh` as you suggested.

gyuguen: Oh man, I understood that I was ignoring `v2alpha1`.
I'm sorry :)

hansol-medi: No problem :)
Let's ignore `v2alpha1` in the next PR with new ver(`v2alpha2`) of doracle!

gyuguen:
And I'm working on a PR related to [this](https://github.com/medibloc/panacea-core/pull/363).
If you apply PR first, the PR I created will rebase the master and create a new PR.

proto/panacea/oracle/v2alpha2/tx.proto
```diff
@@ -0,0 +1,68 @@
...
+
+  // VOTE_OPTION_UNSPECIFIED
+  VOTE_OPTION_UNSPECIFIED = 0;
+
+  // VOTE_OPTION_VALID
+  VOTE_OPTION_VALID = 1;
+
+  // VOTE_OPTION_INVALID
+  VOTE_OPTION_INVALID = 2;
+}
```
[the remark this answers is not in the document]
hansol-medi: @inchori, we made the same `VoteOption` enum type. Could you reuse this `VoteOption` in #360?
If you feel the location of `VoteOption` is awkward, we can move it to separated proto file, e.g. `Vote.proto`.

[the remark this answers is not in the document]
inchori: Ah okay, I understand! I'll import it and reuse it. Thanks for mention it!

proto/panacea/oracle/v2alpha2/genesis.proto
```diff
@@ -0,0 +1,85 @@
...
+  ];
+  string quorum = 3 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
+    (gogoproto.nullable) = false
+  ];
+}
+
+// SlashParams defines the params related to slashing of oracle.
+message SlashParams {
+  string slash_fraction_down_time = 1 [
```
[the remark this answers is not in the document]
hansol-medi: fixed in [b54ebe2](https://github.com/medibloc/panacea-core/pull/361/commits/b54ebe29d9248b3954732b7dc71bde7d87fae59f)

proto/panacea/oracle/v2alpha2/oracle.proto
```diff
@@ -0,0 +1,71 @@
...
+
+  // ACTIVE defines the oracle status that is active
+  ORACLE_STATUS_ACTIVE = 1;
+  // JAILED defines the oracle status that has been jailed
+  ORACLE_STATUS_JAILED = 2;
+  // MIGRATING defines the oracle status that is in migrating to new version of oracle
+  ORACLE_STATUS_MIGRATING = 3;
```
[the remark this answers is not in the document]
hansol-medi: fixed in [b54ebe2](https://github.com/medibloc/panacea-core/pull/361/commits/b54ebe29d9248b3954732b7dc71bde7d87fae59f)

proto/panacea/oracle/v2alpha2/oracle.proto
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// Oracle defines a detail of oracle.
+message Oracle {
+  string address = 1;
+  OracleStatus status = 2;
+  uint64 up_time = 3;
```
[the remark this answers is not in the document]
hansol-medi: fixed in [b54ebe2](https://github.com/medibloc/panacea-core/pull/361/commits/b54ebe29d9248b3954732b7dc71bde7d87fae59f)

youngjoon-lee (approved):
thank you

inchori (approved):
LGTM 👍

PR #363, opened 2022-07-25
feat: initialize all code for OracleRegistration

gyuguen:
I processed with the modification work for the initial setting as follows
* Remove dataPool module
* Initialize oracle and dataDeal module
* Add v2 to go_package in proto files
* Move TallyResult and VotingPeriod from genesis.proto to oracle.proto
* Modify `protocgen.sh`

I'll add a few more comments.


proto/panacea/aol/v2/genesis.proto
```diff
@@ -7,7 +7,7 @@ import "panacea/aol/v2/writer.proto";
 import "panacea/aol/v2/topic.proto";
 
 option java_multiple_files = true;
-option go_package = "github.com/medibloc/panacea-core/x/aol/types";
+option go_package = "github.com/medibloc/panacea-core/v2/x/aol/types";
```
gyuguen, line 10: I changed go_package to v2, the reason is to apply v2 to the package name even when importing the generated *.pb.go file.

gyuguen: Example here(https://github.com/medibloc/panacea-core/pull/363/files#diff-4178f3a958cb7d27bee7d4cc8057249dfd825b9f2104cc1ee0cbc69c97e3942dR10)

proto/panacea/oracle/v2alpha2/genesis.proto
```diff
@@ -1,24 +1,17 @@
...
 import "google/protobuf/duration.proto";
 import "google/protobuf/timestamp.proto";
+import "panacea/oracle/v2alpha2/oracle.proto";
 
 // GenesisState defines the oracle module's genesis state.
 message GenesisState {
-}
-
-// TallyResult defines the tally result of the votes.
-message TallyResult {
```
gyuguen, line 15: I moved this to `oracle.proto`, a cycle error occurred, so I took action.

f it is good that the message exists here, I will return it and find out if there is another way.

scripts/protocgen.sh
```diff
@@ -17,7 +17,7 @@ protoc_gen_gocosmos() {
 
 protoc_gen_gocosmos
 
-proto_dirs=$(find ./proto -path -prune -o -name '*.proto' -not -path '*/v2alpha2/*' -print0 | xargs -0 -n1 dirname | sort | uniq)
+proto_dirs=$(find ./proto -path -prune -o -name '*.proto' -not -path '*/v2alpha1/*' -print0 | xargs -0 -n1 dirname | sort | uniq)
```
gyuguen, line 20: Fixed to generate pb files, excluding those included in v2alpha1.

scripts/protocgen.sh
```diff
@@ -30,5 +30,6 @@ Mgoogle/protobuf/any.proto=github.com/cosmos/cosmos-sdk/codec/types:. \
 done
 
 ## move proto files to the right places
-cp -r github.com/medibloc/panacea-core/* ./
-rm -rf github.com
+find ./x -type f -name "*.pb*.go" -exec rm {} \;
```
gyuguen, line 33: There was a problem that the existing pb file was not deleted. It has been modified to automatically delete it.

scripts/protocgen.sh
```diff
@@ -30,5 +30,6 @@ Mgoogle/protobuf/any.proto=github.com/cosmos/cosmos-sdk/codec/types:. \
...
+cp -rv github.com/medibloc/panacea-core/v*/x/* ./x/
```
gyuguen, line 34: The go_package of the proto file was modified to include v2, but v2 is not included in the actual go folder structure.

gyuguen:
I will reopen this PR after all other related PRs have been corrected.

youngjoon-lee (approved):
good

PR #365, opened 2022-07-27
feat: implement params of oracle module

gyuguen:
Implemented params of Oracle module.


x/oracle/types/params.go
```diff
@@ -0,0 +1,131 @@
...
+	return paramtypes.ParamSetPairs{
+		paramtypes.NewParamSetPair(KeyOraclePublicKey, &p.OraclePublicKey, validateOraclePublicKey),
+		paramtypes.NewParamSetPair(KeyOraclePubKeyRemoteReport, &p.OraclePubKeyRemoteReport, validateOraclePubKeyRemoteReport),
+		paramtypes.NewParamSetPair(KeyUniqueID, &p.UniqueId, validateUniqueID),
+		paramtypes.NewParamSetPair(KeyVoteParams, &p.VoteParams, validateVoteParams),
+		paramtypes.NewParamSetPair(KeySlashParams, &p.SlashParams, validateSlashParams),
+	}
+}
+
+func (p Params) Validate() error {
```
hansol-medi, line 52: Do you have any specific reason for excluding `validateOraclePubKeyRemoteReport` and `validateUniqueID`?

x/oracle/types/params.go
```diff
@@ -0,0 +1,131 @@
...
+	if !ok {
+		return fmt.Errorf("invalid parameter type: %T", i)
+	}
+	if sdk.NewDec(0).Equal(slashParams.SlashFractionDowntime) {
+		return fmt.Errorf("'slashFactionDowntime' cannot be set to zero")
+	}
+
+	if sdk.NewDec(0).Equal(slashParams.SlashFractionForgery) {
+		return fmt.Errorf("'slashFactionForgery' cannot be set to zero")
+	}
```
hansol-medi, line 128: I think the slash fraction can be zero, which means there is no slash. Do we have to prevent from setting to zero?

Rather, the slash fractions should be non-negative and smaller or equal to 1.

x/oracle/types/params.go
gyuguen: The reason is that `validateOraclePubKeyRemoteReport` and `validateUniqueID` are values that cannot be validated by panacea.
I'm not sure what to validate when these two values are non-empty. :)

x/oracle/types/params.go
gyuguen: Yes, your right.
I added a check negative
https://github.com/medibloc/panacea-core/pull/365/commits/70c97ebe7b959e8b87e8311d6fcb16a7bddb33ff

hansol-medi: And it should not be greater than 1, which means 100%. The slash fraction has both lower bound(0) and upper bound(0).

gyuguen: Thank you.
I fixed it. (https://github.com/medibloc/panacea-core/pull/365/commits/87a4cd16f5f4ed37e0d67e04974f3b576f3279b1)

hansol-medi: Thanks!

x/oracle/types/params.go
```diff
@@ -0,0 +1,140 @@
...
+
+func DefaultParams() Params {
+	return Params{
+		OraclePublicKey:          nil,
+		OraclePubKeyRemoteReport: nil,
+		UniqueId:                 "",
+		VoteParams: VoteParams{
+			VotingPeriod: 30 * time.Second,
+			JailPeriod:   10 * time.Minute,
+			Quorum:       sdk.NewDecWithPrec(1, 3),
```
youngjoon-lee, line 33: Is `0.001` the value that you intended? If you wanted `1/3`, you can use `sdk.NewDec(1).Quo(sdk.NewDec(3))`

youngjoon-lee (approved):
lgtm

x/oracle/types/params.go
gyuguen: Right. I meant `0.001`

hansol-medi (approved):

x/oracle/types/params.go
youngjoon-lee: Okay. Then, if only 1 out of 1000 votes yes, the tally result would be finalized with yes (depending on your implementation).

gyuguen: Ah, I misunderstood.
I'll have to set it up as you said in the beginning

PR #364, opened 2022-07-26
feat: handle MsgVoteOracleRegistration transaction

gyuguen:
Closes #362 

A transation that performs a new oracle validation vote.

The follwoing functions will be developed in other PRs.
* Implement `genesis.go`
* implement `params.go`

Things to think about
* error handling
* seperate`oracle_vote.go`?

PR #366, opened 2022-07-28
feat: implements oracle InitGenesis and ExportGenesis

gyuguen:
Implements oracle `InitGenesis` and `ExportGenesis`


x/oracle/keeper/oracle.go
```diff
@@ -134,6 +178,27 @@ func (k Keeper) SetOracleRegistration(ctx sdk.Context, regOracle *types.OracleRe
...
+	defer iterator.Close()
+
+	oracleRegistrationVotes := make([]types.OracleRegistrationVote, 0)
+
+	for ; iterator.Valid(); iterator.Next() {
+		bz := iterator.Value()
+		var oracleRegistrationVote types.OracleRegistrationVote
+		err := k.cdc.UnmarshalLengthPrefixed(bz, &oracleRegistrationVote)
+		if err != nil {
+			return nil, sdkerrors.Wrapf(types.ErrGetOracleRegistration, err.Error())
```
hansol-medi, line 193: When error occurs, an error like `ErrGetOracleRegistrationVote` seems to be returned, not `ErrGetOracleRegistration`.

x/oracle/types/oracle.go
```diff
@@ -13,19 +14,73 @@ func (m SignedOracleRegistrationVote) ValidateBasic() error {
...
+
+func (m OracleRegistration) ValidateBasic() error {
+	if len(m.UniqueId) == 0 {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "uniqueID is empty")
+	}
+	if _, err := sdk.AccAddressFromBech32(m.Address); err != nil {
+		return sdkerrors.Wrapf(err, "oracle address is invalid. address: %s", m.Address)
+	}
+
+	if m.NodePubKey != nil {
```
hansol-medi, line 51: If `m.NodePubKey` is nil, it will pass this validation. So we need to add checks for `m.NodePubKey` not to be nil.

x/oracle/types/oracle.go
```diff
@@ -13,19 +14,73 @@ func (m SignedOracleRegistrationVote) ValidateBasic() error {
...
 }
+
+func (m Oracle) ValidateBasic() error {
+	if _, err := sdk.AccAddressFromBech32(m.Address); err != nil {
+		return sdkerrors.Wrapf(err, "oracle address is invalid. address: %s", m.Address)
+	}
+	return nil
+}
+
+func (m OracleRegistration) ValidateBasic() error {
```
hansol-medi, line 43: Like `Oracle.ValidateBasic()`, some fields are not cared for basic validation. I think the belows should be also handled in this ValidateBasic function. wdyt?

- NodePubKeyRemoteReport : not nil
- VotingPeriod : not nil
- TallyResult : can be nil or valid type of TallyResult

x/oracle/types/oracle.go
```diff
@@ -13,19 +14,73 @@ func (m SignedOracleRegistrationVote) ValidateBasic() error {
...
+	}
+
+	if m.TrustedBlockHash == nil {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "trustedBlockHash is nil")
+	}
+
+	return nil
+}
+
+func (m OracleRegistrationVote) ValidateBasic() error {
```
hansol-medi, line 68: `m.VoteOption` should be one of the enum type of `VoteOption`.

x/oracle/keeper/oracle.go
```diff
@@ -72,6 +72,28 @@ func (k Keeper) validateOracleRegistrationVote(ctx sdk.Context, signedVote *type
 	return nil
 }
 
+func (k Keeper) GetAllOracles(ctx sdk.Context) ([]types.Oracle, error) {
```
hansol-medi, line 75: Like this comments in [previous PR](https://github.com/medibloc/panacea-core/pull/364/files#r931841974), I'd like to discuss about naming of these kinds of functions too.

First of all, I prefer this naming when we get **all** items from KV store for `ExportGenesis`.

Why don't we name this kind of keeper function to `GetAll*`?

Likewise, we may need keeper functions to retrieve multiple items from KV store. In that case, I suggest to use `Get*List`. wdyt?

gyuguen: Yes. It will be fine.
If so, proceed as follows.

`GetAllOracleList` -  gets all Oracles. (using ExportGenesis)
`GetOracleList` -  gets the paged Oracle list.

x/oracle/types/oracle.go
gyuguen: Thank you.
What does valid type mean in TallyResult?

x/oracle/types/oracle.go
gyuguen: Yes, I added validation before [PR](https://github.com/medibloc/panacea-core/pull/364/files#diff-34aa5cd30b13359bc81fc82388624970b489c973ce0b85a7982beb48a4e96b77)

x/oracle/keeper/oracle.go
hansol-medi: It seems good to me!

x/oracle/types/oracle.go
hansol-medi: If the TallyResult is not nil, all the fields in TallyResult should not be nil.

gyuguen: The `Yes`, `No`, `InvalidYes` field inside TallyResult is `sdk.Int`.
I can't do a nil check, but I can do a negative check.

This negative check and consensusValue(->byte array) will be checked nil.

gyuguen: Thank you.
I fixed it.(https://github.com/medibloc/panacea-core/pull/366/commits/3919243b28c5bdbcd393f86c719662138602f480)

x/oracle/keeper/oracle.go
gyuguen: Thank you!
I fixed it.(https://github.com/medibloc/panacea-core/pull/366/commits/5f3e3f88a7a4423f46f0039485e37a3385404dd0)

hansol-medi (approved):
LGTM

proto/panacea/oracle/v2alpha2/genesis.proto
```diff
@@ -10,6 +10,11 @@ import "panacea/oracle/v2alpha2/oracle.proto";
 
 // GenesisState defines the oracle module's genesis state.
 message GenesisState {
+  repeated Oracle oracles = 1 [(gogoproto.nullable) = false];
+  repeated OracleRegistration oracle_registrations = 2 [(gogoproto.nullable) = false];
+  repeated OracleRegistrationVote oracle_registration_votes = 3 [(gogoproto.nullable) = false];
+  Params params = 4 [(gogoproto.nullable) = false];
+  UpgradeOracleInfo upgrade_oracle_info = 5 [(gogoproto.nullable) = false];
```
youngjoon-lee, line 17: i think it's better to rename this to OracleUpgradeInfo.

youngjoon-lee (approved):
lgtm

proto/panacea/oracle/v2alpha2/genesis.proto
gyuguen: Thank you. I fixed it.(https://github.com/medibloc/panacea-core/pull/366/commits/921e33050dbc6008c5e3f7ccf2abdf03b3b2c86e)

PR #367, opened 2022-07-28
fix: [v2.0] use a new version of `cosmos/ledger-go` for Ledger Nano S+

youngjoon-lee:
Based on the `v2.0` branch.

The `cosmos-sdk` uses the `ledger-go` v0.11.1 which doesn't support Ledger Nano S+ (A newly released device).
Fortunately, the Cosmos team made a fix for `ledger-go` a week ago: https://github.com/cosmos/ledger-go/pull/4, and they are working on applying a new version of `ledger-go` to their `cosmos-sdk`. Until then, let's use a new version of `ledger-go` directly in the `panacea-core`.

This fix will be released as v2.0.4 so that the Ledger team can continue their app review process. But, it's not mandatory to upgrade our mainnet to v2.0.4 since this fix affects to only `panacead ... --ledger` commands.

I'll also apply this fix to the `master` branch.


hansol-medi (approved):
LGTM 👍

gyuguen (approved):
LGTM 👍

inchori (approved):
LGTM 👍

PR #372, opened 2022-08-01
feat: implementation of MsgRegisterOracle Tx

### LINKED ISSUE #369: Implement MsgRegisterOracle transaction

hansol-medi:
Close #369 

Implementation of `MsgRegisterOracle` Tx.

To be an oracle, the oracle should be a registered and active validator. 
For that, staking keeper is imported to the oracle module. After getting validator from staking keeper, it is checked if the validator is active (not to be jailed).

I'm not sure if we need a cli for this `MsgRegisterOracle` because it will be done in doracle process automatically. But I made it just in case.

FYI, some refactorings of basic validations (related to node pub key, remote report, and trusted block info) are done to reuse in `MsgRegisterOracle.ValidateBasic`. 

The remained TODO items are as below:
- add event emission of `MsgRegisterOracle`
- add to queue of `MsgRegisterOracle` voting


youngjoon-lee (approved):
lgtm

gyuguen:
How about verifying the `trustedBlockHeight` and `trustedBlockHash` in panacea-core as well?
It seems that `panacea-core` can verify it with its own block information.

hansol-medi:
> How about verifying the trustedBlockHeight and trustedBlockHash in panacea-core as well?
It seems that panacea-core can verify it with its own block information.

It sounds good. I'll try and re-request review after that.

hansol-medi:
@gyuguen,

Do you have any good idea for the verification you said? I have not found a good way to get block info at a specific height without using [clientCtx](https://github.com/cosmos/cosmos-sdk/blob/07f7035f8d88c362ca065e73aa6900ce46f86f52/client/context.go#L27)

gyuguen:
> @gyuguen,
> 
> Do you have any good idea for the verification you said? I have not found a good way to get block info at a specific height without using [clientCtx](https://github.com/cosmos/cosmos-sdk/blob/07f7035f8d88c362ca065e73aa6900ce46f86f52/client/context.go#L27)

Umm... It seems that I can't get it without using the client.
We'd better think about this later.

gyuguen (approved):
LGTM 👍

hansol-medi:
> Umm... It seems that I can't get it without using the client.
We'd better think about this later.

I see. Thank you for you comments!

youngjoon-lee:
> > Umm... It seems that I can't get it without using the client.
> > We'd better think about this later.
> 
> I see. Thank you for you comments!

Yeah. I agree with validating whether the trusted block hash is matched with the trusted block height. We can add this validation later as you said.

Although I don't remember exactly, we can get a block info from the Cosmos layer as the following example that I just wrote: https://github.com/medibloc/panacea-core/commit/9fb48d508aa1d1814bd269f05418beed5673c7f4.
You guys can probably find a better way.

PR #373, opened 2022-08-02
feat: add oracle registration vote queue

gyuguen:
### Implements
* Added queue to get list of `OracleRegistrationVotes` that have closed voting.
    * Only data whose time passed by the key is retrieved.

Reference link: https://github.com/cosmos/cosmos-sdk/blob/main/x/gov/keeper/keeper.go#L203

### DIFF
--- a/x/oracle/keeper/keeper.go
+++ b/x/oracle/keeper/keeper.go
@@ -45,9 +45,9 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 	return ctx.Logger().With("module", fmt.Sprintf("x/%s", types.ModuleName))
 }
 
-func (k Keeper) AddOracleRegistrationVoteQueue(ctx sdk.Context, addr sdk.AccAddress, endTime time.Time) {
+func (k Keeper) AddOracleRegistrationVoteQueue(ctx sdk.Context, uniqueID string, addr sdk.AccAddress, endTime time.Time) {
 	store := ctx.KVStore(k.storeKey)
-	store.Set(types.GetOracleRegistrationVoteQueueKey(addr, endTime), addr)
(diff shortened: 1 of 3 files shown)


x/oracle/keeper/keeper.go
```diff
@@ -43,3 +44,18 @@ func NewKeeper(
 func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 	return ctx.Logger().With("module", fmt.Sprintf("x/%s", types.ModuleName))
 }
+
+func (k Keeper) AddOracleRegistrationVoteQueue(ctx sdk.Context, uniqueID string, addr sdk.AccAddress, endTime time.Time) {
+	store := ctx.KVStore(k.storeKey)
+	store.Set(types.GetOracleRegistrationVoteQueueKey(uniqueID, addr, endTime), addr)
+}
+
+func (k Keeper) GetEndOracleRegistrationVoteQueueIterator(ctx sdk.Context, endTime time.Time) sdk.Iterator {
```
youngjoon-lee, line 53: Would be better to name this as `GetClosedOracleRegistrationVoteIterator`, because the adjective `end` means `final`.

youngjoon-lee, line 53: Because this function returns an iterator for expired(closed) votes, it would be better to name this as `GetClosedOracleRegistrationVoteIterator`. The adjective `end` means `final`.

youngjoon-lee (approved):
lgtm

x/oracle/keeper/keeper.go
gyuguen: You're right. I fixed it.(https://github.com/medibloc/panacea-core/pull/373/commits/9c2aec03501f500b6551d0ed66584341f2fbab4f)

youngjoon-lee (approved):

hansol-medi (approved):
lgtm

PR #374, opened 2022-08-03
feat: Implement Emission of Register Oracle Event

inchori:
Implement Emission of Register Oracle Event for subscribing in `panacea-doracle`.
- `EventType`, `AttributeKey`, and `AttributeValue` were implemented in `types/event.go`.
- For now, `AttributeValue` is `RegisterOracleEvent`.

### DIFF
--- a/x/oracle/keeper/oracle.go
+++ b/x/oracle/keeper/oracle.go
@@ -36,6 +36,7 @@ func (k Keeper) RegisterOracle(ctx sdk.Context, msg *types.MsgRegisterOracle) er
 			types.EventTypeRegisterOracle,
 			sdk.NewAttribute(types.AttributeKeyOracle, types.AttributeValueOracle)),
 	})
+
 	return nil
 }
 
--- a/x/oracle/keeper/oracle.go
+++ b/x/oracle/keeper/oracle.go
@@ -31,7 +31,11 @@ func (k Keeper) RegisterOracle(ctx sdk.Context, msg *types.MsgRegisterOracle) er
 
 	// TODO: Add active queue of Oracle Registration
 
-	// TODO: emit RegisterOracle event
+	ctx.EventManager().EmitEvents(sdk.Events{
+		sdk.NewEvent(
+			types.EventTypeRegisterOracle,
+			sdk.NewAttribute(types.AttributeKeyRegisterOracle, types.AttributeValueRegisterOracle)),
+	})
 	return nil
 }
 
--- a/x/oracle/keeper/oracle.go
+++ b/x/oracle/keeper/oracle.go
@@ -31,7 +31,11 @@ func (k Keeper) RegisterOracle(ctx sdk.Context, msg *types.MsgRegisterOracle) er
 
 	// TODO: Add active queue of Oracle Registration
 
-	// TODO: emit RegisterOracle event
+	ctx.EventManager().EmitEvents(sdk.Events{
+		sdk.NewEvent(
+			types.EventTypeRegisterOracle,
+			sdk.NewAttribute(types.AttributeKeyOracle, types.AttributeValueOracle)),
+	})
 	return nil
 }
 
--- a/x/oracle/keeper/oracle.go
+++ b/x/oracle/keeper/oracle.go
@@ -99,6 +99,7 @@ func (k Keeper) validateOracleRegistrationVote(ctx sdk.Context, vote *types.Orac
 	if err != nil {
 		return err
 	}
+
 	if oracle.Status != types.ORACLE_STATUS_ACTIVE {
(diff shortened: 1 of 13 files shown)


audtlr24 (approved):
lgtm

x/oracle/keeper/oracle.go
```diff
@@ -31,7 +31,11 @@ func (k Keeper) RegisterOracle(ctx sdk.Context, msg *types.MsgRegisterOracle) er
 
 	// TODO: Add active queue of Oracle Registration
 
-	// TODO: emit RegisterOracle event
+	ctx.EventManager().EmitEvents(sdk.Events{
+		sdk.NewEvent(
+			types.EventTypeRegisterOracle,
+			sdk.NewAttribute(types.AttributeKeyRegisterOracle, types.AttributeValueRegisterOracle)),
```
hansol-medi, line 37: `sdk.NewAttribute` accepts key and value of the attribute that you want to emit. 

For example in [`send.go`](https://github.com/cosmos/cosmos-sdk/blob/71035879e4af4551b9bcf16f79bc52785793602e/x/bank/keeper/send.go#L209) of `x/bank` module, the recipient is one of the attributes whose key is `types.AttributeKeyRecipient` and the value is `toAddr.String()`.

In case of `MsgRegisterOracle`, I think the value of the attribute should be the address of oracle
```go
ctx.EventManager().EmitEvents(sdk.Events{
		sdk.NewEvent(
			types.EventTypeRegisterOracle,
			sdk.NewAttribute(types.AttributeKeyRegisterOracle, mgs.oracleAddress)),
})
```

x/oracle/types/event.go
```diff
@@ -0,0 +1,10 @@
+package types
+
+const (
+	AttributeKeyRegisterOracle   = "oracle"
```
hansol-medi, line 4: `AttributeKeyOracle` might be better. Or `AttributeKeyOracleAddress`.

x/oracle/types/event.go
```diff
@@ -0,0 +1,10 @@
...
+	AttributeValueRegisterOracle = "RegisterOracleEvent"
```
hansol-medi, line 5: As I commented, the value is not a const, so we don't need this value.

x/oracle/types/event.go
gyuguen: I agree. :)

gyuguen:
It looks like a mix of PRs from multiple people.
Please tidy up.

youngjoon-lee:
> It looks like a mix of PRs from multiple people. Please tidy up.

@inchori 
Yeah. Please ping me after rebasing your branch to the latest master. Then, I'll revisit.

x/oracle/keeper/oracle.go
gyuguen: It seems that one of the information needed when detecting an event in subscribe is the `EventType`.
I think this EventType would be defined like this:
```
RegisterOracleEvent
RegisterOracleCompleteEvent
SellDataEvent
GetBuyDataEvent
RegisterOracleMigrationEvent
RegisterOracleMigrationCompleteEvent
```

Subscribers should see this event and decide what to do about it.
I think @inchori triggered an event to get this type. Is that right?

gyuguen: If my prediction is correct, it would be better to fire an event like this:
```
vote.type=RegisterOracleEvent
vote.type=RegisterOracleCompleteEvent
vote.type=SellDataEvent
vote.type=GetBuyDataEvent
vote.type=RegisterOracleMigrationEvent
vote.type=RegisterOracleMigrationCompleteEvent
```
The name of `vote.type` deserves some thought.

x/oracle/keeper/oracle.go
```diff
@@ -31,7 +31,11 @@ func (k Keeper) RegisterOracle(ctx sdk.Context, msg *types.MsgRegisterOracle) er
 
 	// TODO: Add active queue of Oracle Registration
 
-	// TODO: emit RegisterOracle event
+	ctx.EventManager().EmitEvents(sdk.Events{
```
gyuguen, line 34: This function seems to be deprecated.
You can use to `ctx.EventManager().EmitTypedEvents(...)`

x/oracle/keeper/oracle.go
inchori: Yes, as we discussed, the `AttributeKey` of `Event` should be like that @gyuguen commented. If we defined the events like that, that can be detected easy for subscriber.

x/oracle/keeper/oracle.go
inchori: Hmm.. I already knew about it. But for now, we just go for `EmitEvents` because other keepers still use `EmitEvents`.

PR #376, opened 2022-08-11
chore: comment implement panic

hansol-medi:
Current version of panacea cannot be initiated and started because panic occurs in validating and initiating the genesis of `datadeal` module. So I just commented them.


audtlr24 (approved):
Lgtm!

gyuguen (approved):
Oh, thank you
I was going to work on it, but you did it first.😄

PR #378, opened 2022-08-12
feat: implement TallyResult calculated by OracleRegistration's vote in EndBlocker

gyuguen:
Added ability to find closed votes in OracleRegistration in EndBlocker and count them as TallyResult.
Additionally, if the vote passes, Oracle is registered.

**Worrying about**
* Handling in case of a tie between `Yes` and `No`
* Handling in case of a tie between `Yes` and `Invalid`
* When an error occurs while EndBlocker is running, a panic occurs, and after that, do we want to stop creating blocks?


x/oracle/abci_test.go
```diff
@@ -0,0 +1,227 @@
...
+
+	oraclePubKey  cryptotypes.PubKey
+	oracleAddr    sdk.AccAddress
+	oraclePubKey2 cryptotypes.PubKey
+	oracleAddr2   sdk.AccAddress
+	oraclePubKey3 cryptotypes.PubKey
+	oracleAddr3   sdk.AccAddress
+
+	newOraclePubKey cryptotypes.PubKey
+	newOracleAddr   sdk.AccAddress
```
gyuguen, line 30: I put these variables in a struct instead of global.
Because if a variable is declared globally, it can be accessed by other test files in the same package.
Separated for test independence.

x/oracle/abci.go
```diff
@@ -0,0 +1,49 @@
...
+		// If ConsensusValue does not exist, consensus has not been passed.
+		if tallyResult.IsPassed() {
+			oracleRegistration.Status = types.ORACLE_REGISTRATION_STATUS_PASSED
+			oracleRegistration.EncryptedOraclePrivKey = tallyResult.ConsensusValue
+
+			oracle := types.NewOracle(oracleRegistration.Address, types.ORACLE_STATUS_ACTIVE)
+			err := keeper.SetOracle(ctx, oracle)
+			if err != nil {
+				panic(err)
+			}
```
H4NLee, line 35:
```suggestion
			if err := keeper.SetOracle(ctx, oracle); err != nil {
				panic(err)
			}
```

x/oracle/abci.go
```diff
@@ -0,0 +1,49 @@
...
+		} else {
+			oracleRegistration.Status = types.ORACLE_REGISTRATION_STATUS_REJECTED
+		}
+
+		oracleRegistration.TallyResult = tallyResult
+
+		err = keeper.SetOracleRegistration(ctx, oracleRegistration)
+		if err != nil {
+			panic(err)
+		}
```
H4NLee, line 45:
```suggestion
		if err = keeper.SetOracleRegistration(ctx, oracleRegistration); err != nil {
			panic(err)
		}
```

H4NLee (approved):
minor comments, but looks good to me.

In case of `x/gov` module, there is a threshold for pass of proposal (https://github.com/cosmos/cosmos-sdk/blob/fabf7399a9a68fc2d91340eeddb145e1e772c12b/x/gov/keeper/tally.go#L120). 
For now, it's not that important but we may have to discuss about the policy of pass of vote. I think current condition of pass of vote (`TallyResult.IsPassed()`) is good enough.

gyuguen:
> minor comments, but looks good to me.
> 
> In case of `x/gov` module, there is a threshold for pass of proposal (https://github.com/cosmos/cosmos-sdk/blob/fabf7399a9a68fc2d91340eeddb145e1e772c12b/x/gov/keeper/tally.go#L120). For now, it's not that important but we may have to discuss about the policy of pass of vote. I think current condition of pass of vote (`TallyResult.IsPassed()`) is good enough.

We discussed this yesterday, and it was agreed that it would only pass if the percentage of people who voted `yes` exceeded the quorum. :)

x/oracle/abci.go
gyuguen: Thank you.
I fixed it. (https://github.com/medibloc/panacea-core/pull/378/commits/fd209b2f029fc714a551c56ca9298797ac66e527)

x/oracle/abci.go
gyuguen: Thank you.
I fixed it. (https://github.com/medibloc/panacea-core/pull/378/commits/fd209b2f029fc714a551c56ca9298797ac66e527)

youngjoon-lee (commented):
lgtm

youngjoon-lee (approved):
lgtm

PR #380, opened 2022-08-12
feat: add grpc keeper func of params

H4NLee:
When querying the module params of `x/oracle`, panic occurred.
Because the implementation of `oracleKeeper.Params()` for gRPC was missed, so I implemented it.

FYI, `QueryParamsResponse` -> `QueryOracleParamsResponse` in `query.proto` for consistency.


H4NLee:
> FYI, QueryParamsResponse -> QueryOracleParamsResponse in query.proto for consistency.

Oh.. proto lint said `RPC response type "QueryOracleParamsResponse" should be named "ParamsResponse" or "QueryParamsResponse"`..

How about `QueryParamsRequest` and `QueryParamsResponse`? or leave it as was?

gyuguen (approved):
LGTM 👍

youngjoon-lee (approved):

PR #382, opened 2022-08-16
feat: add event emission of oracle registration

inchori:
Add event emission of oracle registration for subscribing in ```panacea-doracle```.

- ```EventType```, ```AttributeKey```, and ```AttributeValue``` were implemented in types/event.go.
- For now, AttributeValue is ```RegisterOracleEvent```.


x/oracle/types/event.go
```diff
@@ -0,0 +1,8 @@
+package types
+
+const (
+	EventTypeRegisterOracle = "register"
+
+	AttributeKeyOracle   = "oracle"
+	AttributeValueOracle = "RegisterOracleEvent"
```
gyuguen, line 7: How about changing it like below?
It looks like we can use EventType and AttributeKey in common.

```suggestion
	EventTypeVote= "vote"
	AttributeKeyVoteType   = "voteType"

	VoteTypeRegisterOracle = "RegisterOracle"
	VoteTypeRegisterOracleComplete = "RegisterOracleComplete"
```

x/oracle/keeper/oracle.go
```diff
@@ -31,7 +31,12 @@ func (k Keeper) RegisterOracle(ctx sdk.Context, msg *types.MsgRegisterOracle) er
 
 	// TODO: Add active queue of Oracle Registration
 
-	// TODO: emit RegisterOracle event
+	ctx.EventManager().EmitEvents(sdk.Events{
+		sdk.NewEvent(
+			types.EventTypeRegisterOracle,
+			sdk.NewAttribute(types.AttributeKeyOracle, types.AttributeValueOracle)),
+	})
```
H4NLee, line 38: When you see the [guide](https://docs.cosmos.network/master/core/events.html), the attribute of event is to give more information about the categorized event.
So, we don't have to add a new attribute if we use a fixed value of attribute. I think this event is to alert all oracles no matter who register, we don't need a new attribute.

```suggestion
	ctx.EventManager().EmitEvents(sdk.Events{
		sdk.NewEvent(
			types.EventTypeRegisterOracle,
		),
	})
```

gyuguen: As you said, it seems to work, but all events seem to respond to events in the form `{EventType}.{AttributeKey}={AttributeValue}`.
```json
  {
    "type":"coin_received",
    "attributes":[
      {
        "key":"cmVjZWl2ZXI=",
        "value":"cGFuYWNlYTE3eHBmdmFrbTJhbWc5NjJ5bHM2Zjg0ejNrZWxsOGM1bGFrbHhzNw==",
        "index":true
      },
      {
        "key":"YW1vdW50",
        "value":"MTAwMDAwMHVtZWQ=",
        "index":true
      }
    ]
  },
  {
    "type":"transfer",
    "attributes":[
      {
        "key":"cmVjaXBpZW50",
        "value":"cGFuYWNlYTE3eHBmdmFrbTJhbWc5NjJ5bHM2Zjg0ejNrZWxsOGM1bGFrbHhzNw==",
        "index":true
      },
      {
        "key":"c2VuZGVy",
        "value":"cGFuYWNlYTF6dHl5YTZ6cTduZzNtN2x2ejV4a3UzNmM0dDVqNzAyZ3A0YWc2YQ==",
        "index":true
      },
      {
        "key":"YW1vdW50",
        "value":"MTAwMDAwMHVtZWQ=",
        "index":true
      }
    ]
  }
```


If it is as suggested, I think the result will be answered as below.
```json
{
    "type": "register",
    "attributes": []
}
```

So, I think it would be good for us to respond as well.
```json
{
    "type": "vote",
    "attributes": [
        {
            "key": "voteType",
            "value": "registerOracle"
        }
    ]
}
```

What do you think?

H4NLee: Oh I understood. I think the way @gyuguen suggested is better.

x/oracle/types/event.go
gyuguen: I think you can use this VoteType like an enum.
```
type VoteType string

const (
	VOTE_TYPE_REGISTER_ORACLE VoteType = "OracleRegister"
	VOTE_TYPE_REGISTER_ORACLE_COMPLETE VoteType = "OracleRegisterComplete"
)
```

inchori: Fixed in [2e806ae](https://github.com/medibloc/panacea-core/pull/382/commits/2e806ae059cdeabdd6fcaad7ecb7b96792f27c9e). Plus, I didn't use `VoteType`. If I add the `VoteType` the cosmos-sdk said that convert to string.

youngjoon-lee: From cosmos-sdk, it seems that naming convention is snake_case.
https://github.com/cosmos/cosmos-sdk/blob/eef086f77071e7ccd68e808aa5460de84ecbc387/x/bank/types/events.go#L18

youngjoon-lee: @gyuguen @inchori
Hm.. I think the previous naming made by @inchori is better. 
Does this event present that an oracle was just registered now? Then, I think it's better to emit an event that "A new oracle was registered!!!". 
Then, the next action (voting) should be decided by subscribers.
We don't need to explicitly specify the next action in the event.

gyuguen: Yes, I understood what @youngjoon-lee was saying.

Then we have some Event handling to implement in the future.
* oracle registration requsted
* oracle registration vote ended
* data sell requested
* ...

Are we going to put these behaviors in the EventType?

youngjoon-lee: Yes. I think so. It's similar as what the `x/bank` does. When coins are transferred between accounts, one of events that it emits is:
- event type: `coin_received`
- attr
  - key: `receiver`
    value: `panacea1qewori....`
  - key: `amount`
    value: `100umed`

inchori: Okay, I understand it. For now, I'll make that the attribute key value of oracle registration event is address of oracle.

gyuguen: Aha, so the subscriber will have no problem subscribing to this event, right?

gyuguen: I thought the subscriber should know all this format `{eventType}.{attributeKey}={attributeValue}`

inchori: Ah, I understand @gyuguen 's comment now. How to subscriber know the `attributeValue`? The subscriber use `query` but they already need to know `attributeValue`.

gyuguen: Yes, If the subscriber can detect this Event by its eventType then it should be fine.

inchori: Okay, for now I apply these comments. And I'll test on `panacea-doracle`.

inchori: Fixed in [3bc2bac](https://github.com/medibloc/panacea-core/pull/382/commits/3bc2bacc415dac91757b00d03f7c351147c0340f).

gyuguen:
@inchori  
With this fix, is the subscriber good at event detection?

H4NLee:
I think this PR is still in draft, right?

inchori:
Unfortunately, the subscriber cannot detect the event. So I think we should rollback it as @gyuguen suggested.

@H4NLee Yes, this PR is still in draft. I'll convert to draft it. Thanks!

inchori:
As we discussed, I will close this PR.

PR #384, opened 2022-08-18
feat: added voting deletion after voting ends

gyuguen:
Added a callback function to perform as part of processing each vote.
This will usually include the ability to delete votes.


H4NLee (approved):
lgtm

youngjoon-lee (approved):
lgtm

PR #387, opened 2022-08-19
fix: Check whether voting is possible when calculate voting power

gyuguen:
Fixed to exclude oracles that are unable to vote when calculating total voting power.


gyuguen:
I will close this PR and reopen it with another PR.

PR #388, opened 2022-08-22
feat: improve test code

gyuguen:
Here are some improvements.
* We have created OracleBaseTestSuite, a collection of functions commonly used in Oracle's `TestSuite`.
* Variables used in tests are declared in the suite so that each can be used.


gyuguen:
I don't know why this exists with the master log...
I will regenerate the PR.

PR #389, opened 2022-08-22
fix: check whether voting is possible when calculate voting power

gyuguen:
Fixed to exclude oracles that are unable to vote when calculating total voting power.


H4NLee:
This PR also seems to be from wrong base branch :)

gyuguen:
> This PR also seems to be from wrong base branch :)

Right. After organizing the PR, it was reopened. :)

H4NLee (approved):

youngjoon-lee (approved):
lgtm

PR #391, opened 2022-08-23
feat: added event when OracleRegistration voting starts/ends.

gyuguen:
* OracleRegistration voting start: Occurs when `RegisterOracle` transaction is successful
* OracleRegistration voting end: In Oracle's Endblocker, code is implemented that adds an event when the vote calculate ends.


x/oracle/types/events.go
```diff
@@ -0,0 +1,9 @@
+package types
+
+const (
+	EventTypeOracleRegistration = "oracle_registration"
+
+	AttributeKeyVoteStatus        = "vote_status"
+	AttributeKeyVotingTarget      = "voting_target"
+	AttributeValueStatusCompleted = "completed"
```
youngjoon-lee, line 8:
```suggestion
	AttributeValueVoteStatusCompleted = "completed"
```
probably?

youngjoon-lee (approved):

H4NLee (approved):

x/oracle/types/events.go
gyuguen: Right, I will fix soon.

x/oracle/abci.go
```diff
@@ -49,6 +49,14 @@ func EndBlocker(ctx sdk.Context, keeper keeper.Keeper) {
 			panic(err)
 		}
 
+		ctx.EventManager().EmitEvent(
+			sdk.NewEvent(
+				types.EventTypeOracleRegistration,
+				sdk.NewAttribute(types.AttributeKeyVoteStatus, types.AttributeValueStatusCompleted),
```
H4NLee: This event seems to be emitted when the vote for oracle registration is ended. If you intend to emit event of `end of vote`, I think you need to rename the value of the attribute to `AttributeValueStatusEnded` or something.

If this event emission is for passed vote, it should be moved inside the `if tallyResult.IsPassed()` statement.

gyuguen: This event is emitted when voting is end. 
So I think it would be better to set the status ad `end` as @H4NLee said.

And it would be nice if the `start` event was also emitted when the voting started.

x/oracle/abci_test.go
```diff
@@ -240,4 +250,14 @@ func (suite abciTestSuite) TestEndBlockerVoteReject() {
 	oracleVotes, err = suite.OracleKeeper.GetAllOracleRegistrationVoteList(ctx)
 	suite.Require().NoError(err)
 	suite.Require().Equal(0, len(oracleVotes))
+
+	events := ctx.EventManager().Events()
```
inchori, line 254: It's not a big issue. But you can change it to `suite.Ctx.EventManager().Events()`.

gyuguen: Yes, I used the `ctx` variable from here to `suite.Ctx`

https://github.com/medibloc/panacea-core/pull/391/files#diff-5ab93a9d86b554f353a4615e22adbfef295849e090c0b139aae4fb37bf04f499R159

gyuguen:
@youngjoon-lee @H4NLee @inchori 
I improved it to emit an event when voting starts and ends.
Please review again :)

youngjoon-lee (approved):

x/oracle/keeper/oracle.go
```diff
@@ -36,7 +36,13 @@ func (k Keeper) RegisterOracle(ctx sdk.Context, msg *types.MsgRegisterOracle) er
...
 	k.AddOracleRegistrationQueue(ctx, oracleRegistration.UniqueId, oracleAccAddr, oracleRegistration.VotingPeriod.VotingEndTime)
 
-	// TODO: emit RegisterOracle event
+	ctx.EventManager().EmitEvent(
+		sdk.NewEvent(
+			types.EventTypeOracleRegistration,
+			sdk.NewAttribute(types.AttributeKeyVoteStatus, types.AttributeValueVoteStatusStarted),
+			sdk.NewAttribute(types.AttributeKeyOracleAddress, oracleRegistration.Address),
+		),
+	)
```
youngjoon-lee, line 45: I've heard that this event emission was already implemented by @inchori in the PR https://github.com/medibloc/panacea-core/pull/393/.

Let's define a ground rule of naming convention.
From [Cosmos SDK doc](https://docs.cosmos.network/v0.46/modules/staking/07_events.html), it would be better to name them as:

### MsgRegisterOracle

|Type|Attr key|Attr val|
|---|---|--|
|register_oracle|oracle_address|panacea1...| 

### EndBlocker

|Type|Attr key|Attr val|
|---|---|--|
|complete_vote|vote_type|oracle_registration or ...| 
|                       |oracle_address|panacea1...|

youngjoon-lee (approved):

H4NLee (approved):
Good. 👍

x/oracle/types/events.go
```diff
@@ -0,0 +1,10 @@
+package types
+
+const (
+	EventTypeOracleRegistration = "oracle_registration"
```
H4NLee, line 4: For the name of this event type, how about `EventTypeRegistrationVote`?
When we see other event types, it seems to refer the detailed action of each message.

gyuguen: Right. It seems good.
I fixed it. (https://github.com/medibloc/panacea-core/pull/391/commits/324e15294580b56398d52874e55c3ca2aa0f84cc)

PR #393, opened 2022-08-24
feat: implement event emission for oracle registration

inchori:
- Implement `message.module='oracle'` in `msg_server_oracle.go`.
- Implement `oracle_registered.oracle_address='oracleAddress'` in `keeper/oracle.go`.

### DIFF
--- a/x/oracle/types/events.go
+++ b/x/oracle/types/events.go
@@ -1,0 +1,6 @@
+package types
+
+const (
+	EventTypeRegister         = "oracle_registered"
+	AttributeKeyOracleAddress = "oracle_address"
+)
--- a/x/oracle/types/events.go
+++ b/x/oracle/types/events.go
@@ -1,6 +1,10 @@
 package types
 
 const (
-	EventTypeRegister         = "oracle_registered"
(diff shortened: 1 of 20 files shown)


x/oracle/types/events.go
```diff
@@ -0,0 +1,6 @@
+package types
+
+const (
+	EventTypeRegister         = "oracle_registered"
```
H4NLee, line 4: I think it would be better to make this event type much clearer.
"oracle_registered" seems to be meant that "oracle is registered. so what do you do next?" to me.

This event is emitted when a request for oracle registration is submitted.
Thus, I suggest to rename it something like `EventTypeRegisterRequested` or `EventTypeRegisterSubmitted`. (Actually I don't like this name since it's too long.. but I have no better idea 😭)

H4NLee (approved):

x/oracle/types/events.go
inchori: The event type was renamed as we discussed in #391.

gyuguen (approved):
LGTM 👍

H4NLee (approved):
LGTM

PR #397, opened 2022-08-31
feat: added the ability to read from a file when setting up Genesis oracle

gyuguen:
Add an option to support oracle_pub_key.json, which is created when the `gen-oracle-key` command is executed in panacea-doracle.


H4NLee (approved):
lgtm

youngjoon-lee (approved):

PR #399, opened 2022-09-06
fix: oracle registration status check of RegisterOracle

H4NLee:
A validator, who can apply to be an oracle, could repeatedly request oracle registration regardless of the voting result. It results in the emission of event repeatedly to other oracles, and makes other oracles vote redundantly.

Thus, oracle registration vote can be opened only in the following cases.

- When applying for registration for the first time (not stored in `OracleRegistration` kv store)
- Applied but the previous vote is rejected (`OracleRegistration.Status` == `REJECTED`)
- When a registered oracle want to be shared the oracle private key again (e.g. delete `oracle_priv_key.sealed`) (`OracleRegistration.Status` == `PASSED` && `Oracle.Status` == `JAILED`)

### DIFF
--- a/x/oracle/keeper/oracle.go
+++ b/x/oracle/keeper/oracle.go
@@ -76,22 +76,23 @@ func (k Keeper) checkOracleRegistrationStatus(ctx sdk.Context, uniqueID, oracleA
 		return err
 	}
 
-	if existing.Status == types.ORACLE_REGISTRATION_STATUS_VOTING_PERIOD {
-		return fmt.Errorf("in voting period")
-	}
-
-	if existing.Status == types.ORACLE_REGISTRATION_STATUS_PASSED {
+	switch existing.Status {
+	case types.ORACLE_REGISTRATION_STATUS_PASSED:
 		oracle, err := k.GetOracle(ctx, oracleAddress)
 		if err != nil {
 			return err
 		}
-
 		if oracle.Status != types.ORACLE_STATUS_JAILED {
 			return fmt.Errorf("only jailed oracle can re-register oracle")
 		}
+		return nil
+	case types.ORACLE_REGISTRATION_STATUS_VOTING_PERIOD:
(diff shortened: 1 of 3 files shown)


x/oracle/keeper/oracle.go
```diff
@@ -60,6 +66,34 @@ func (k Keeper) CheckValidatorStatus(ctx sdk.Context, oracleAccAddr sdk.AccAddre
...
+		if err != nil {
+			return err
+		}
+
+		if oracle.Status != types.ORACLE_STATUS_JAILED {
+			return fmt.Errorf("only jailed oracle can re-register oracle")
+		}
+	}
+
+	return nil
```
gyuguen, line 94: How about handling for all states like this?
In this case, if a new state is added later, an error will be raised and unexpected behavior can be detected.

```suggestion
	switch existing.Status {
	case types.ORACLE_REGISTRATION_STATUS_PASSED:
		oracle, err := k.GetOracle(ctx, oracleAddress)
		if err != nil {
			return err
		}

		if oracle.Status != types.ORACLE_STATUS_JAILED {
			return fmt.Errorf("only jailed oracle can re-register oracle")
		}

		return nil
	case types.ORACLE_REGISTRATION_STATUS_VOTING_PERIOD:
		return fmt.Errorf("in voting period")
	case types.ORACLE_REGISTRATION_STATUS_REJECTED:
		return nil
	default:
		return fmt.Errorf("unexpected state. status: %s", existing.Status)
	}
```

gyuguen (approved):
LGTM 👍 
I left one unimportant comment.

x/oracle/keeper/oracle.go
H4NLee: Thank you.
As you said, it would be better to detect unexpected status. I'll improve it as you suggested.

inchori (approved):
LGTM

youngjoon-lee (approved):

PR #407, opened 2022-09-16
fix: x/oracle: change the proto type of base64 values from `bytes` to `string`

### LINKED ISSUE #406: x/oracle: change the proto type of base64 values from `bytes` to `string`
Related to https://github.com/medibloc/panacea-doracle/pull/70

We always put base64 strings to the following fields:
https://github.com/medibloc/panacea-core/blob/fd4a0ce6c4fba633764f7268f5d518bd2170d6bb/proto/panacea/oracle/v2alpha2/genesis.proto#L24-L26

In this case, it seems that Protobuf puts double quotations around the actual base64 string. It can cause a lot of unexpected issues.
So, let's change the type from `bytes` to `strings` for more clarity.

youngjoon-lee:
Closes #406


x/oracle/keeper/oracle.go
```diff
@@ -101,7 +102,8 @@ func (k Keeper) VoteOracleRegistration(ctx sdk.Context, vote *types.OracleRegist
 		return sdkerrors.Wrap(types.ErrOracleRegistrationVote, err.Error())
 	}
 
-	if k.verifyVoteSignature(ctx, vote, signature) {
+	ok, err := k.verifyVoteSignature(ctx, vote, signature)
+	if err != nil || !ok {
```
gyuguen, line 106: Later, when implementing the slashing function, this if statement will have to be implemented in a different way.
If err is not nil, slashing should not be performed because it is an error caused by a cause other than signature verification failure.
We probably only need to slash if no error occurs and the ok value is false.

Of course, you don't have to consider it now.

gyuguen (approved):
LGTM 👍 

I left an unimportant comment.

x/oracle/keeper/oracle.go
youngjoon-lee: Make senses. I will fix it now for safety. I think we can throw a panic if base64 decoding is failed, based on the rule that all values in the `Param` are valid if the `Param` struct was populated successfully.

gyuguen: I agree. :)

youngjoon-lee: Done: 212d423b02f60219d858f0150efc756a4e021414

H4NLee (approved):
lgtm

inchori (approved):
LGTM 👍

PR #418, opened 2022-09-21
feat: Update issue templates

inchori:


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
...
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+
+**Describe the solution you'd like**
+A clear and concise description of what you want to happen.
+
+**Describe alternatives you've considered**
+A clear and concise description of any alternative solutions or features you've considered.
+
+**Additional context**
+Add any other context or screenshots about the feature request here.
```
youngjoon-lee, line 20: It seems that this format is good for projects that were already released and are in their maintenance mode, but we're developing the main feature from scratch now :) Let's me try to see other templates as well soon. If there is nothing better than this, let's go with this template with some modifications.

PR #424, opened 2022-09-27
ci: separate the Lint job with the main job

youngjoon-lee:
As the Lint job usually takes minutes, let's separate it with the main CI job, so that we can see the lint result quickly.


gyuguen (approved):
LGTM 👍

inchori (approved):
LGTM

Issue #414, opened 2022-09-21
Update a Deal Status in Endblocker


inchori:
Update a Deal Status in `EndBlocker`.
- When the deal is ready or being started, the status should be `ACTIVE`.
- When the deal is completed, the status should be `COMPLETED`.

audtlr24:
In the process of processing the vote, it would be better to update the deal status and check the sale in `verifiableCid vote`.
- In the case of exceeding `maxNumData` due to multiple processing at the same time, it is necessary to think about how to handle it.
- Only votes that do not exceed `maxNumData` are added to the `dataDelivery Queue`.

PR #432, opened 2022-09-29
fix: complement deal's validate code and use testutil of datadeal module

inchori:
- Validate the existence of deal when sell data in `SellData` method in `keeper/deal.go`.
- Validate the status of deal when sell data in `SellData` method in`keeper/deal.go`.
- Add failure unit test of above things.
- Use `MakeTestDataSale` in `datadeal`'s `testutil`.


audtlr24 (approved):
LGTM!

x/datadeal/keeper/deal.go
```diff
@@ -149,8 +149,14 @@ func (k Keeper) SellData(ctx sdk.Context, msg *types.MsgSellData) error {
...
-	//TODO: Check the Deal which seller want to sell data to the deal exists
-	//k.GetDeal()
+	deal, err := k.GetDeal(ctx, msg.DealId)
+	if err != nil {
+		return sdkerrors.Wrapf(types.ErrSellData, err.Error())
+	}
+
+	if deal.Status == types.DEAL_STATUS_INACTIVE {
+		return sdkerrors.Wrapf(types.ErrSellData, "deal status is INACTIVE")
+	}
```
H4NLee, line 159: You may have to check if the deal is sold out (or completed?) which means `CurNumData` reaches to `MaxNumData`.
For that, we need a new status for deal accordingly.

inchori: I think that when the `curNumData` reaches to `maxNumData`, the deal's status will be change to `INACTIVE`.
As you say, I agree that add a deal's completed status. So, I add a `COMPLETED` status and re-request your review.
Also, It need to implement well when update a deal's status in endblocker as I written in #414 .

H4NLee: I missed the issue you mentioned. BTW, let's be clear because a completed deal is different from a deactivated deal.

inchori: Applied in [1556530](https://github.com/medibloc/panacea-core/pull/432/commits/1556530df0025eb425665521576fe51f45776e70).

gyuguen (approved):
LGTM 👍

H4NLee (approved):
lgtm

PR #433, opened 2022-09-29
feat: add nonce to Deal & refactor

### LINKED ISSUE #415: Implement a SellData Encryption CLI
When seller add own data to IPFS or storage, the data should be encrypted.
- The data will be encrypted with shared key that consists of `oraclePubKey` and `sellerPrivKey`.
- The encryption of selling data will be implemented in panacea CLI.

audtlr24:
To make shared key, add nonce field to `Deal`
- The nonce is used to create a shared key for encryption/decryption

Apply comments that have not yet been reflected
- fix `NewModuleAccount` name
- add CurNumData validation


x/datadeal/types/deal.go
```diff
@@ -8,10 +8,12 @@ import (
 	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
 )
 
+const NonceSize = 12
+
 func NewDeal(dealID uint64, msg *MsgCreateDeal) *Deal {
 
 	dealAddress := NewDealAddress(dealID)
-
+	nonce := make([]byte, 12)
```
inchori, line 16: I know that nonce will be generated when make a new deal. Or does the `create_deal` file include nonce?

audtlr24: The nonce seems to be included because it calls `NewDeal` inside the `CreateDeal` keeper function.

inchori: If the `nonce` is included in file, does the nonce need to change to `msg.Nonce`?

gyuguen: I also agree with @inchori . It would be better to input nonce from Msg.

x/datadeal/types/deal.go
```diff
@@ -51,9 +54,19 @@ func (m Deal) ValidateBasic() error {
 		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "budget is not a valid Coin object")
 	}
 
+	if m.CurNumData <= 0 {
```
gyuguen, line 64: My guess is that `curNumData` can be zero.
Therefore, it seems to have to be modified to be less than 0.
```suggestion
	if m.CurNumData < 0 {
```

gyuguen (commented):
Thank you.
Please check my comment :)

x/datadeal/types/deal.go
audtlr24: I add `ReadFull` function to make nonce randomly. Thanks! 👍

inchori (approved):
LGTM 👍

x/datadeal/types/deal.go
audtlr24: As @gyuguen said, I move that to msg. Please refer to commit [9a1ed93](https://github.com/medibloc/panacea-core/pull/433/commits/9a1ed9300a9379fdc2c2341850a51b5846229ee4).

gyuguen:
I thought about nonce.
The buyer must decrypt with the nonce of the deal.
If it is a random byte array, the buyer will be exposed as an unknown value when printing.
We have to think about whether it is better to receive this nonce value as a 1 byte string. This is so that people can recognize it when it is printed.
For example, you can enter alphabets and numbers.

x/datadeal/types/deal.go
audtlr24: Lint says we don't need to do that check in case of `uint64`. So, I delete that check code.

gyuguen:
> I thought about nonce. The buyer must decrypt with the nonce of the deal. If it is a random byte array, the buyer will be exposed as an unknown value when printing. We have to think about whether it is better to receive this nonce value as a 1 byte string. This is so that people can recognize it when it is printed. For example, you can enter alphabets and numbers.

Please don't think about this comment. This is not a problem for now. :)

x/datadeal/client/cli/txCreateDeal.go
```diff
@@ -53,5 +55,12 @@ func newCreateDealMsg(file string) (sdk.Msg, error) {
 		return nil, err
 	}
 
+	nonce := make([]byte, 12)
```
gyuguen, line 58:
```suggestion
	nonce := make([]byte, types.NonceSize)
```

gyuguen (approved):
LGTM 👍 
Thank you

H4NLee:
Wait, if the nonce will be generated randomly in cli, the nonce doesn't seem to need to be in the msg.
When creating a new deal, can you make random nonce and set it to deal?

gyuguen:
> Wait, if the nonce will be generated randomly in cli, the nonce doesn't seem to need to be in the msg. When creating a new deal, can you make random nonce and set it to deal?

Wouldn't it be better to set the nonce creation by the cli?
We made it random in cmd, but it can be implemented to receive input from the user in other cli.

H4NLee:
Ah, sorry. Please forget the comment.
Random generation of nonce cannot be in keeper function. This is the right way.

H4NLee (approved):
LGTM

PR #440, opened 2022-10-04
feat: add VoteDataDelivery keeper function

audtlr24:
- add `VoteDataDelivery` keeper function & test codes
- tx, query and genesis state will be developed next PR


x/datadeal/keeper/deal.go
```diff
@@ -230,3 +233,81 @@ func (k Keeper) GetAllDataSaleList(ctx sdk.Context) ([]types.DataSale, error) {
...
+func (k Keeper) validateDataDeliveryVote(ctx sdk.Context, vote *types.DataDeliveryVote) error {
+	dataSale, err := k.GetDataSale(ctx, vote.VerifiableCid, vote.DealId)
+	if err != nil {
+		return err
+	}
+	if dataSale.Status != types.DATA_SALE_STATUS_DELIVERY_VOTING_PERIOD {
+		return fmt.Errorf("the current voted data's status is not 'DELIVERT_VOTING_PERIOD'")
+	}
+	return nil
+}
```
inchori, line 274: I think it need to verify a oracle's status, too

x/datadeal/testutil/suite.go
```diff
@@ -42,3 +42,19 @@ func (suite *DataDealBaseTestSuite) MakeNewDataSale(sellerAddr sdk.AccAddress, v
 		DeliveryTallyResult:     nil,
 	}
 }
+
+func (suite *DataDealBaseTestSuite) MakeNewDataSaleDeliveryVoting(sellerAddr sdk.AccAddress, verifiableCID string) *types.DataSale {
+	return &types.DataSale{
```
inchori, line 47: Is this method return `types.DataSale`? I think this should be return `types.DataDeliveryVote`.
Also, how about change name to `MakeNewDataDeliveryVote`?

inchori (commented):
Thanks for your work! I left a few comments.

x/datadeal/keeper/deal.go
audtlr24: Thanks, I will add it 👍

x/datadeal/testutil/suite.go
audtlr24: Ah, I want to make `DataSale` that `status` is `DATA_SALE_STATUS_DELIVERY_VOTING_PERIOD` not `DATA_SALE_STATUS_VERIFICATION_VOTING_PERIOD`. If a function to change `status` is added later, I will delete it.

inchori: Ah, I got it.

x/datadeal/keeper/deal.go
```diff
@@ -230,3 +233,89 @@ func (k Keeper) GetAllDataSaleList(ctx sdk.Context) ([]types.DataSale, error) {
...
+	return secp256k1.PubKey(oraclePubKeyBz).VerifySignature(voteBz, signature)
+}
+
+func (k Keeper) validateDataDeliveryVote(ctx sdk.Context, vote *types.DataDeliveryVote) error {
+	oracle, err := k.oracleKeeper.GetOracle(ctx, vote.VoterAddress)
+	if err != nil {
+		return err
+	}
+	if oracle.Status != oracletypes.ORACLE_STATUS_ACTIVE {
+		return fmt.Errorf("this oracle is not in 'ACTIVE' state")
```
H4NLee, line 271: There is the same error in #438. It would be also good to make a sentinel error for reuse.

x/datadeal/keeper/deal.go
```diff
@@ -230,3 +233,89 @@ func (k Keeper) GetAllDataSaleList(ctx sdk.Context) ([]types.DataSale, error) {
...
+	}
+
+	dataSale, err := k.GetDataSale(ctx, vote.VerifiableCid, vote.DealId)
+	if err != nil {
+		return err
+	}
+	if dataSale.Status != types.DATA_SALE_STATUS_DELIVERY_VOTING_PERIOD {
+		return fmt.Errorf("the current voted data's status is not 'DELIVERT_VOTING_PERIOD'")
```
H4NLee, line 279:
```suggestion
		return fmt.Errorf("the current voted data's status is not 'DELIVERY_VOTING_PERIOD'")
```

x/datadeal/types/expected_keepers.go
```diff
@@ -10,6 +11,7 @@ type AccountKeeper interface {
 
 	SetAccount(ctx sdk.Context, acc authtypes.AccountI)
 	GetAccount(ctx sdk.Context, addr sdk.AccAddress) authtypes.AccountI
+	GetPubKey(sdk.Context, sdk.AccAddress) (cryptotypes.PubKey, error)
```
H4NLee, line 14: I think there is no need to have this func.

x/datadeal/types/keys.go
```diff
@@ -43,6 +45,10 @@ func GetDataSaleKey(verifiableCID string, dealID uint64) []byte {
 	return append(DataSaleKey, CombineKeys(sdk.Uint64ToBigEndian(dealID), []byte(verifiableCID))...)
 }
 
+func GetDataDeliveryVoteKey(verifiableCID string, voterAddress sdk.AccAddress, dealID uint64) []byte {
+	return append(DataDeliveryVoteKey, CombineKeys([]byte(verifiableCID), voterAddress, sdk.Uint64ToBigEndian(dealID))...)
```
H4NLee, line 49: As mentioned in #438 [comments](https://github.com/medibloc/panacea-core/pull/438#discussion_r987448680), I suggest to change the order.

 ```suggestion
	return append(DataDeliveryVoteKey, CombineKeys(sdk.Uint64ToBigEndian(dealID), []byte(verifiableCID), voterAddress)...)
```

File: x/datadeal/types/datasale.go [hunk 7]
```diff
@@ -77,3 +77,27 @@ func (m DataSale) ValidateBasic() error {
 
 	return nil
 }
+
+func (m DataDeliveryVote) ValidateBasic() error {
+	if m.DealId <= 0 {
```
H4NLee, line 82: It seems to be enough to check if the value is zero or not because the type of `DealId` is `uint64` :)

H4NLee (commented):
Left few comments.
And it seems that we can make `verifyVoteSignature` reusable using `oracletypes.Vote` interface.

x/datadeal/keeper/deal.go
audtlr24: I fixed this part, but please check if it is the correct.

x/datadeal/keeper/deal.go
audtlr24: Thanks, I fix it.

x/datadeal/types/expected_keepers.go
audtlr24: Yes, you're right.

x/datadeal/types/datasale.go
audtlr24: Thanks, I fix it.

x/datadeal/types/keys.go
audtlr24: Thanks, I fix it. But I wonder what importance the order in the key has. Can you explain?

inchori:
In addition, I think that the `DataDeliveryVote` method in `msg_server_deal.go` should be implemented in this branch.

inchori (approved):
LGTM

x/datadeal/keeper/deal.go
```diff
@@ -230,3 +233,89 @@ func (k Keeper) GetAllDataSaleList(ctx sdk.Context) ([]types.DataSale, error) {
...
+}
+
+func (k Keeper) validateDataDeliveryVote(ctx sdk.Context, vote *types.DataDeliveryVote) error {
+	oracle, err := k.oracleKeeper.GetOracle(ctx, vote.VoterAddress)
+	if err != nil {
+		return err
+	}
+	if oracle.Status != oracletypes.ORACLE_STATUS_ACTIVE {
+		return sdkerrors.Wrap(types.ErrOracleNotActive, "")
+	}
```
H4NLee, line 272: You can just return the `types.ErrOracleNotActive` error here :)

x/datadeal/types/keys.go
H4NLee: When querying a list of items (e.g. oracles, deals, etc.), we usually use get the data based on prefix of key.

In this case, when the `dealID` comes first, we can query all the votes of a specific deal using a combination of prefix of `DataDeliveryVoteKey` and `sdk.Uint64ToBigEndian(dealID)`.

Likewise, we can also query all the votes for a specific `verifiableCID` using a combination of prefix of `DataDeliveryVoteKey`, `sdk.Uint64ToBigEndian(dealID)` and `[]byte(verifiableCID)`

audtlr24: Thanks, I understand now 👍 !

youngjoon-lee (approved):
lgtm

x/datadeal/types/keys.go
audtlr24: Thanks, I fix it!

H4NLee (approved):
lgtm

x/datadeal/client/cli/queryDataDeliveryVote.go
```diff
@@ -0,0 +1,47 @@
...
+			dealID, err := strconv.ParseUint(args[0], 10, 64)
+			if err != nil {
+				return err
+			}
+
+			queryClient := types.NewQueryClient(clientCtx)
+			req := &types.QueryDataDeliveryVoteRequest{
+				DealId:        dealID,
+				VerifiableCid: args[1],
+				VoterAddress:  args[2],
```
gyuguen, line 32: I think it would be good to verify the `voterAddress`.

gyuguen (commented):
LGTM
I left a comment, but it doesn't matter

PR #442, opened 2022-10-05
feat: implement encrypt-data cli

### LINKED ISSUE #415: Implement a SellData Encryption CLI

audtlr24:
Close #415 
- add encrypt-data cli that encrypt seller data
- add the `output-file-path` to the cli to save it as a file
- test `encrypt-data` and `decrypt-data` in local panacea environment


inchori (approved):
LGTM Thanks! 👍

cmd/panacead/cmd/encrypt.go
```diff
@@ -0,0 +1,122 @@
...
+	cmd.Flags().String(FlagCipherText, "", "Cipher text in the form of a byte array")
+	cmd.Flags().String(FlagCipherTextPath, "", "Path to the file where the cipher text in the form of a byte array is stored")
+
+	flags.AddQueryFlagsToCmd(cmd)
+
+	return cmd
+}
+
+func encrypt(clientCtx client.Context, keyName string, originData []byte, oraclePubKey string, nonce []byte) ([]byte, error) {
+	privKeyHex, err := keyring.NewUnsafe(clientCtx.Keyring).UnsafeExportPrivKeyHex(keyName)
```
youngjoon-lee, line 92: It is unsafe to use `keyring.NewUnsafe(...)` because of the type assertion internally, but it seems that there's no choice for us now. 
Let's keep it in mind for now (Also, in the [`decrypt()`](https://github.com/medibloc/panacea-core/blob/0f1d22f75cc97c5e7c9ca2f077b88058a0a7cfaa/cmd/panacead/cmd/decrypt.go#L91)). I'll check whether there is any better way.

youngjoon-lee (approved):
lgtm

cmd/panacead/cmd/encrypt.go
audtlr24: Thanks, I'll keep that in mind for now.

gyuguen (commented):
LGTM

PR #444, opened 2022-10-06
feat: implement distribution rewards to oracles

### LINKED ISSUE #435: Implement distribution of reward to oracles
## Background

Oracles can be rewarded for validating and delivering data. The reward will be proportional to the voting power of each oracle.

## Implement

When `DataDeliveryVote` is passed, the oracles who cast votes for both `DataVerificationVote` and `DataDeliveryVote` normally will be rewarded proportional to their voting power.
The reward may be distributed to use [`distrKeeper.AllocateTokensToValidator`](https://github.com/cosmos/cosmos-sdk/blob/a9f02d9cb7447b3da1fb00b4458588db55e9b5fb/x/distribution/keeper/allocation.go#L61), which enables to distribute the reward to the delegator as well as oracle proportional to voting power.

H4NLee:
Close #435 

in `x/oracle` module,
implemented `DistributeRewardToOracle()` using `k.distrKeeper.AllocateTokensToValidator()` which distributes reward to a single validator and its delegators

in `x/datadeal` module,
implemented `DistributeRewards` using the above `DistributeRewardToOracle()` which distributes reward to all oracles. 
The way of receiving voted oracles for both data verification and delivery (currently a simple `map[string]*oracletypes.OracleValidatorInfo`) can be changed.


x/datadeal/keeper/reward.go
```diff
@@ -0,0 +1,58 @@
...
+	dealAddress, err := sdk.AccAddressFromBech32(deal.GetAddress())
+	if err != nil {
+		panic(err)
+	}
+
+	// calculate oracle commission
+	totalBudget := deal.GetBudget().Amount.ToDec()
+	maxNumData := sdk.NewIntFromUint64(deal.GetMaxNumData()).ToDec()
+	oracleCommission := k.oracleKeeper.GetParams(ctx).OracleCommissionRate
+	pricePerData := sdk.NewCoin(assets.MicroMedDenom, totalBudget.Quo(maxNumData).TruncateDec().Mul(oracleCommission).TruncateInt())
```
audtlr24, line 28: Thanks for your work! 👍  But, I have one question. Currently, consideration of the tx fee is not included in the calculation process, is this not necessary? I wonder if it is an automated process or if the fee itself is not used.

x/datadeal/keeper/reward.go
```diff
@@ -0,0 +1,58 @@
...
+	totalVotingPower := sdk.ZeroInt()
+	for _, oracle := range oracles {
+		if oracle.IsPossibleVote() {
+			totalVotingPower = totalVotingPower.Add(oracle.BondedTokens)
+		}
+	}
+	totalVotingPowerDec := totalVotingPower.ToDec()
+
+	// distribute rewards
+	for _, oracle := range oracles {
```
gyuguen, line 50: We use distributeKeeper to distribute rewards, so it doesn't seem like it would be easy to track the history.
How about adding an event here so we can check it in the block info?

gyuguen (approved):
Thank you!
I left a comment, but it doesn't matter

And it's probably not a problem, but it might be a good idea to manually test later to see if the rewards are properly distributed to the delegates.

x/datadeal/keeper/reward.go
H4NLee: Yes. AFAIK, it is consumed automatically whenever there is a read or write to the store. 
You can refer the docs [here](https://docs.cosmos.network/v0.46/basics/gas-fees.html#main-gas-meter) :)

H4NLee:
> And it's probably not a problem, but it might be a good idea to manually test later to see if the rewards are properly distributed to the delegates.

Good point. Actually I tried to add delegators in this test case, but I faced some errors in delegation.
I will figure it out and add test case with delegators as well.

x/datadeal/keeper/reward.go
H4NLee: `distrKeeper.AllocateTokensToValidator()` also emits events regarding to commission and reward.
But it would be also good to add event emission of **reward distribution for data verification or delivery**.
I'll add it 👍

audtlr24 (approved):
LGTM!

inchori (approved):
LGTM 👍

PR #445, opened 2022-10-06
feat: add dataDeliveryVote info to genesis state

audtlr24:
- add `DataDeliveryVote` info to genesis state
- change `GetDeal`, `SetDeal` to be consistent with other functions
- `genesis_test` code will be developed later while refactoring all at once


inchori (commented):
Thanks for your work!
I think you need to add `DataDeliveryVote` in `DefaultGenesis()` function and `Validate()` method which is located in `datadeal/types/genesis.go`.

inchori (approved):
lgtm

audtlr24:
Thanks, I add `DefaultGenesis()` and `Validate()` functions. And I change the verification of `DeliveredCid`, because I think `DeliveredCid` can be empty when the vote option is `VOTE_OPTION_NO`.

x/datadeal/genesis.go
```diff
@@ -30,25 +35,34 @@ func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState)
...
+	genesis.DataDeliveryVotes = dataDeliveryVotes
 
 	return &types.GenesisState{
-		Deals:          deals,
-		NextDealNumber: nextDealNum,
+		Deals:             deals,
+		NextDealNumber:    nextDealNum,
+		DataSales:         dataSales,
+		DataDeliveryVotes: dataDeliveryVotes,
 	}
```
H4NLee, line 67: You can just return `genesis`.

gyuguen (commented):
LGTM

Issue #448, opened 2022-10-11
Add a DataHash field in DataSale


inchori:
## Background

When the seller using the other addresses sells own data, it could emerge the duplicated data sale issue.

## Implement

Add a `DataHash` field of raw data in `DataSale` type so that oracles can compare the original data hash with data sale's data hash and vote for it.

PR #447, opened 2022-10-07
feat: add DataDeliveryVote EndBlocker

audtlr24:
Closes #419 
- add `EndBlocker()` that calculate `DataDeliveryVote` tally and update `DataSale`
- Updating and checking `Deal` info(`curNumData`, `status`) will be developed in Issue #414

PR #449, opened 2022-10-11
feat: add DataVerificationVote EndBlocker

### LINKED ISSUE #412: Implement a function to calculate DataVerificationVote
Implement calculating `tallyResult` at `EndBlocker` when the `DataVerificationVote` is ended
- add `Endblocker` to `abci.go` that calculate `tallyResult`, save the `tallyResult` to `DataSale`, update `DealStatus` of `DataSale` and emit event
- add QueueIterator functions to keeper
- Implement by referencing or using the tally functions of `x/oracle`

inchori:
Closes #412 
- Add a calculate tally result of `DataVerificationVote` in `Endblocker`.
- Add a test code with failure test in `abci_test.go`.


audtlr24 (approved):
LGTM!

x/datadeal/abci.go
```diff
@@ -0,0 +1,59 @@
...
+		dataSale.VerificationTallyResult = tallyResult
+		if err := keeper.SetDataSale(ctx, dataSale); err != nil {
+			panic(err)
+		}
+
+		ctx.EventManager().EmitEvent(
+			sdk.NewEvent(
+				types.EventTypeDataVerificationVote,
+				sdk.NewAttribute(types.AttributeKeyVoteStatus, types.AttributeValueVoteStatusEnded),
+				sdk.NewAttribute(types.AttributeKeyVerifiableCID, dataSale.VerifiableCid),
```
audtlr24, line 53: Thanks,  There is one comment.`dealID` attribute is required to determine which `dataSale` is in `doracle`.

inchori: Does it mean the `doracle` need to get `Deal` for determining `DataSale` after the verifying data and voting?

inchori: Yes I agree. I'll add it. Thanks!

x/datadeal/abci.go
```diff
@@ -1,13 +1,66 @@
...
+			panic(err)
+		}
+
+		if tallyResult.IsPassed() {
+			dataSale.Status = types.DATA_SALE_STATUS_DELIVERY_VOTING_PERIOD
+			dataSale.VerifiableCid = string(tallyResult.ConsensusValue)
+
+			keeper.AddDataDeliveryQueue(ctx, dataSale.VerifiableCid, dataSale.DealId, oracleKeeper.GetVotingPeriod(ctx).VotingEndTime)
+		} else {
+			dataSale.Status = types.DATA_SALE_STATUS_FAILED
```
H4NLee, line 43: [As I commented in the previous PR](https://github.com/medibloc/panacea-core/pull/447#discussion_r993030944), I think we need to set this status `DATA_VERIFICATION_FAILED`. In case of failure of delivery vote(whatever the reason is), buyer can re-request delivery to the oracles.

inchori: Yes, this can be more clearer. So I'll change to `DATA_SALE_STATUS_VERIFICATION_FAILED`.

inchori: So our `FAILED` statuses have three types. The `VERIFICATION_FAILED`, `DELIVERY_FAILED`, and `FAILED`.
Or is it have two types except `FAILED`?

H4NLee: I think it's enough to have two status, `VERIFICATION_FAILED` and `DELIVERY_FAILED`. Do we need a status `FAILED` as well?

x/datadeal/abci.go
H4NLee: Good. Then, why don't we add the `dealID` attribute to the delivery vote?
Even if the `doracle` don't use the attribute for now, it would be good to make the event clear.

inchori: It makes sense. If add a `dealID` attribute, the `doracle` can notice that the delivery vote included in which deal.

x/datadeal/keeper/deal.go
```diff
@@ -162,7 +162,7 @@ func (k Keeper) SellData(ctx sdk.Context, msg *types.MsgSellData) error {
 	}
 
 	getDataSale, _ := k.GetDataSale(ctx, msg.VerifiableCid, msg.DealId)
-	if getDataSale != nil && getDataSale.Status != types.DATA_SALE_STATUS_FAILED {
+	if getDataSale != nil && getDataSale.Status != types.DATA_SALE_STATUS_VERIFICATION_FAILED {
 		return sdkerrors.Wrapf(types.ErrSellData, "data already exists")
 	}
```
H4NLee, line 167: Should we distinguish between failed and rejected in data verification?
IMHO, if the data is rejected in verification, seller cannot sell the same data. wdyt?

x/datadeal/abci.go
```diff
@@ -1,13 +1,66 @@
...
+		}
+
+		ctx.EventManager().EmitEvent(
+			sdk.NewEvent(
+				types.EventTypeDataVerificationVote,
+				sdk.NewAttribute(types.AttributeKeyVoteStatus, types.AttributeValueVoteStatusEnded),
+				sdk.NewAttribute(types.AttributeKeyVerifiableCID, dataSale.VerifiableCid),
+				sdk.NewAttribute(types.AttributeKeyDealID, strconv.FormatUint(dataSale.DealId, 10)),
+			),
+		)
```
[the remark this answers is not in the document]
inchori: How about leave a data verification vote end event and add a data delivery vote started event?

x/datadeal/keeper/deal.go
H4NLee: On the other hand, if the verification of data is failed(due to the downtime of oracles), the seller can try to sell the same data again.

inchori: Does it means the `REJECTED` and `FAILED` are different?
So, if the status is `FAILED`, seller can sell the same data again. Am I understand right?

inchori: I have the question. So, when the `REJECTED` status need?
Does the `REJECTED` status will be added in `Endblocker`?

H4NLee: > Does it means the REJECTED and FAILED are different?
So, if the status is FAILED, seller can sell the same data again. Am I understand right?

Right. If the result of vote is `No`(which may mean 2/3+ `No`?), the data verification is `REJECTED`.
And if there is no consensus value, the status of data sale might be `FAILED`.

H4NLee: > I have the question. So, when the REJECTED status need?
Does the REJECTED status will be added in Endblocker?

If I am right and we need to distinguish between `FAILED` and `REJECTED`, we need to handle the result differently.

H4NLee: Give me some seconds. I'll think more about it.

x/datadeal/abci.go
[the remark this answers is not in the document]
H4NLee: I think it would be good. However, I concern that there is a tendency to overuse the emission of event.
If there is no worry to overuse, it's good!

x/datadeal/keeper/deal.go
gyuguen: It would be good to distinguish between `FAILED` and `REJECTED`, but I don't think it's necessary to distringuish them.
Because if oracle vote with `No` once, then oracle will continue to vote with `No`.
So, the seller just keeps paying the fee, and there seems to be no problem functionally.

x/datadeal/abci.go
```diff
@@ -1,13 +1,66 @@
...
+			},
+		)
+
+		if err != nil {
+			panic(err)
+		}
+
+		if tallyResult.IsPassed() {
+			dataSale.Status = types.DATA_SALE_STATUS_DELIVERY_VOTING_PERIOD
+			dataSale.VerifiableCid = string(tallyResult.ConsensusValue)
```
gyuguen, line 39: I think it would be better to compare the verifiableCID of the consensus verifiableCID with the verifiableCID of dataSale if the vote passed.
Because verifiableCID is already stored in dataSale.

gyuguen (commented):
Thank you.
I left a comment.

x/datadeal/keeper/deal.go
H4NLee: Right. I agree that it's not essential.
Let's keep this way for now.

x/datadeal/abci.go
H4NLee: Actually, there is no consensus value in data verification vote, right?

inchori: When the issue #448 resolved, the data verification vote will not have consensus value.

audtlr24: After this branch is merged, I will reflect this part in the PR that adds dataHash (#448).

x/datadeal/abci.go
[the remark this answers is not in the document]
inchori: Applied in [b57333c](https://github.com/medibloc/panacea-core/pull/449/commits/b57333c3f6321de4fa9afa74f420923362cc5b3a).

x/datadeal/abci.go
inchori: Applied in [b57333c](https://github.com/medibloc/panacea-core/pull/449/commits/b57333c3f6321de4fa9afa74f420923362cc5b3a).

x/datadeal/abci.go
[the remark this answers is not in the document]
H4NLee: It seems that other attributes, verifiableCID(maybe later dataHash) and dealID, are also required for `doracle`.

x/datadeal/abci.go
```diff
@@ -1,13 +1,72 @@
...
+			panic(err)
+		}
+
+		if tallyResult.IsPassed() {
+			dataSale.Status = types.DATA_SALE_STATUS_DELIVERY_VOTING_PERIOD
+			if dataSale.VerifiableCid != string(tallyResult.ConsensusValue) {
+				panic("invalid verifiable CID consensus value")
+			}
+
+			keeper.AddDataDeliveryQueue(ctx, dataSale.VerifiableCid, dataSale.DealId, oracleKeeper.GetVotingPeriod(ctx).VotingEndTime)
```
H4NLee, line 43: When the data verification vote is passed, a new vote for data delivery starts.
Accordingly, we need to update the voting period :)

x/datadeal/abci.go
```diff
@@ -1,13 +1,72 @@
...
+		ctx.EventManager().EmitEvents(sdk.Events{
+			sdk.NewEvent(
+				types.EventTypeDataVerificationVote,
+				sdk.NewAttribute(types.AttributeKeyVoteStatus, types.AttributeValueVoteStatusEnded),
+				sdk.NewAttribute(types.AttributeKeyVerifiableCID, dataSale.VerifiableCid),
+				sdk.NewAttribute(types.AttributeKeyDealID, strconv.FormatUint(dataSale.DealId, 10)),
+			),
+			sdk.NewEvent(
+				types.EventTypeDataDeliveryVote,
+				sdk.NewAttribute(types.AttributeKeyVoteStatus, types.AttributeValueVoteStatusStarted),
```
H4NLee, line 63: This event should be emitted only when the verification vote is passed.

x/datadeal/abci.go
inchori: Yes, I wonder about that. How to get the updated voting period?

x/datadeal/abci.go
inchori: So, this event should be in the `if tallyResult.IsPassed()` condition?
IMHO, I think that this event emission also means the verification vote is passed.

x/datadeal/abci.go
gyuguen: You can refer this part to get `votingPeriod`. (https://github.com/medibloc/panacea-core/blob/master/x/oracle/keeper/oracle.go#L378)

Umm... As we decide to vote twice, I think the votingPeriod field should also be split.
```protobuf
message DataSale {
  string seller_address = 1;
  uint64 deal_id = 2;
  string verifiable_cid = 3;
  string delivered_cid = 4;
  DataSaleStatus status = 5;
  panacea.oracle.v2alpha2.VotingPeriod verification_voting_period = 6;
  panacea.oracle.v2alpha2.VotingPeriod delivery_voting_period = 7;
  panacea.oracle.v2alpha2.TallyResult verification_tally_result = 8;
  panacea.oracle.v2alpha2.TallyResult delivery_tally_result = 9;
}
```
What do you think?

H4NLee: Thanks @gyuguen!
Yes we can do. Or can set newly updated `votingPeriod` to data sale and use it.
Both are ok to me.

inchori: Yes, I also used a `GetVotingPeriod()` method. So, I'll change to `dataSale.DeliveryVotingPeriod.VotingEndTime`.

x/datadeal/abci.go
H4NLee: Right. let's move this emission inside the if statement.

x/datadeal/abci.go
inchori: I think this part can be refactored quite a lot. So I'll make a new branch and work on there.

x/datadeal/abci.go
inchori: Fixed in [5790d95](https://github.com/medibloc/panacea-core/pull/449/commits/5790d955cd663d845094563b94667d232c5dadaa).

x/datadeal/abci.go
gyuguen: OK

gyuguen (approved):
LGTM

x/datadeal/abci.go
inchori: This comments reflected in PR #456. Please refer to the branch.

H4NLee (approved):
Looks good to me too

PR #454, opened 2022-10-12
feat: implements oracle upgrade proposal

### LINKED ISSUE #427: Implement OracleUpgradeProposal
## Background
In order to perform Oracle's upgrade, consensus of all members is required.
We also need to decide which Oracle to upgrade to and when to upgrade.

## Implement
Basically, refer to `SofrwareUpgradeProposal`
It can be executed with the following command:
```
panacead tx gov submit-proposal oracle-upgrade --upgrade-unique-id {uniqueID} --upgrade-block-height {height} --title {title} --description {description} --deposit 1000000umed
```
It can also be done with a file.
```
panacead tx gov submit-proposal oracle-oupgrade --upgrade-file-path {path}
```
```json
{
	"title": "Oracle Upgrade",
	"description": "...",
	"deposit":{
		"denom": "umed",
		"amount": "1000000000"
	},
	"content": {
		"uniqueID": "...",
		"blockHeight": 26666
	}
}
```

If `OracleUpgradeProposal` is passed, `UniqueID` and `BlockHeight` should be stored in `OracleUpgradeInfo` KVStore.

gyuguen:
Close #427 

I'm sorry. This PR was so huge that it was not easy to separate.

It implements the following features:
* OracleUpgradeInfo can be chagned through governance.
* Provides `cmd` and `grpcCleint` to query OracleUpgradeInfo.

I'll leave some helpful comments in the review


app/app.go
```diff
@@ -469,6 +470,7 @@ func New(
 		app.GetSubspace(oracletypes.ModuleName),
 		app.StakingKeeper,
 	)
+	govRouter.AddRoute(oracletypes.RouterKey, oracle.NewOracleProposalHandler(app.oracleKeeper))
```
gyuguen, line 473: You must register this router to perform proposal tx normally.

x/oracle/proposal_handler.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
+	"github.com/medibloc/panacea-core/v2/x/oracle/keeper"
+	"github.com/medibloc/panacea-core/v2/x/oracle/types"
+)
+
+func NewOracleProposalHandler(k keeper.Keeper) govtypes.Handler {
```
gyuguen, line 11: This handler is executed once when the first proposal is registered, and then again when the proposal is passed.

When the first proposal is registered, it is stored in the cache, so it is not stored in the kvStore.
After the proposal is passed, it is stored in the kvStore.

app/app.go
```diff
@@ -162,6 +162,7 @@ func getGovProposalHandlers() []govclient.ProposalHandler {
 		upgradeclient.ProposalHandler,
 		upgradeclient.CancelProposalHandler,
 		// this line is used by starport scaffolding # stargate/app/govProposalHandler
+		oracleclient.ProposalHandler,
```
gyuguen, line 165: If it is not registered here, it will not be registered in the command below.
`panacead tx gov submit-proposal ...`

app/app.go
```diff
@@ -162,6 +162,7 @@ func getGovProposalHandlers() []govclient.ProposalHandler {
...
 	)
 	govProposalHandlers = append(govProposalHandlers, wasmclient.ProposalHandlers...)
```
inchori, line 167: Is we need to append `oracleclient.ProposalHandlers`, too?

gyuguen: Yes. You can see it added right above.

x/oracle/proposal_handler.go
```diff
@@ -0,0 +1,33 @@
...
+		case *types.OracleUpgradeProposal:
+			return handlerOracleUpgradeProposal(ctx, k, c)
+		default:
+			return sdkerrors.Wrapf(sdkerrors.ErrUnknownRequest, "unrecognized oracle upgrade proposal content type: %T", c)
+		}
+
+	}
+}
+
+func handlerOracleUpgradeProposal(ctx sdk.Context, k keeper.Keeper, p *types.OracleUpgradeProposal) error {
```
H4NLee, line 23: We can check if current block height is less than the target height of upgrade as upgrade keeper's handler do.
https://github.com/cosmos/cosmos-sdk/blob/fa2cb40892e4653926dd8ff9fb6702e80313af33/x/upgrade/keeper/keeper.go#L186

x/oracle/client/cli/txUpgradeOracleProposal.go
```diff
@@ -0,0 +1,89 @@
...
+
+			content, err := parseArgsToContent(cmd)
+			if err != nil {
+				return err
+			}
+
+			msg, err := govtypes.NewMsgSubmitProposal(content, deposit, from)
+			if err != nil {
+				return err
+			}
```
H4NLee, line 50: Why don't we add `msg.ValidateBasic()` here?

x/oracle/client/cli/txUpgradeOracleProposal.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	uniqueID, err := cmd.Flags().GetString(FlagUpgradeUniqueID)
+	if err != nil {
+		return nil, err
+	}
+
+	plan := types.Plan{UniqueId: uniqueID, Height: height}
```
H4NLee, line 86: We can check the validity of plan from cmd using `plan.ValidateBasic()` before generate tx

proto/panacea/oracle/v2alpha2/oracle.proto
```diff
@@ -123,4 +123,13 @@ enum VoteOption {
 
   // VOTE_OPTION_INVALID
   VOTE_OPTION_NO = 2;
+}
+
+// UpgradeOracleInfo defines the info of oracle upgrade, which includes the target height of upgrade and unique ID of the new version of oracle
+message OracleUpgradeInfo {
+  option (gogoproto.equal) = true;
```
H4NLee, line 130: May I ask do you have any specific reason to set this option?

x/oracle/client/cli/queryOracleUpgrade.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/medibloc/panacea-core/v2/x/oracle/types"
+	"github.com/spf13/cobra"
+)
+
+func CmdGetOracleUpgradeInfo() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "oracle-upgrade-info",
+		Short: "Query a oracle upgrade information",
```
H4NLee, line 13:
```suggestion
		Short: "Query an oracle upgrade information",
```

x/oracle/client/cli/txUpgradeOracleProposal.go
```diff
@@ -0,0 +1,89 @@
...
+	FlagUpgradeHeight   = "upgrade-height"
+)
+
+func CmdUpgradeOracleProposal() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "oracle-upgrade (--upgrade-unique-id [uniqueID]) (--upgrade-height [height]) [flags]",
+		Args:  cobra.ExactArgs(0),
+		Short: "Submit a oracle upgrade proposal",
+		Long: "Submit a oracle upgrade along with an initial deposit.\n + " +
+			"You must enter the uniqueID and block height of the Oracle to perform the upgrade.",
```
H4NLee, line 24:
```suggestion
		Long: "Submit an oracle upgrade proposal along with an initial deposit.\n + " +
			"You must enter the uniqueID of a new version of oracle and target block height for upgrade.",
```

x/oracle/client/cli/txUpgradeOracleProposal.go
```diff
@@ -0,0 +1,89 @@
...
+
+			return tx.GenerateOrBroadcastTxCLI(clientCtx, cmd.Flags(), msg)
+		},
+	}
+
+	cmd.Flags().String(govcli.FlagTitle, "", "title of proposal")
+	cmd.Flags().String(govcli.FlagDescription, "", "description of proposal")
+	cmd.Flags().String(govcli.FlagDeposit, "", "deposit of proposal")
+	cmd.Flags().Int64(FlagUpgradeHeight, 0, "The height at which the upgrade must happen")
+	cmd.Flags().String(FlagUpgradeUniqueID, "", "Oracle's uniqueID to be upgraded")
```
H4NLee, line 60: If this two flags are required, why don't we mark these two flags as requirement?

H4NLee (approved):
Left minor comments and looks good to me.

x/oracle/client/cli/txUpgradeOracleProposal.go
gyuguen: You're right. I'll add

x/oracle/client/cli/txUpgradeOracleProposal.go
gyuguen: Good. I'll add

H4NLee:
For now, only one upgrade-info is stored. Don't we need to do store all the upgrade info for archiving history?

app/app.go
inchori: Ah, yes. I missed it 😂

gyuguen:
> For now, only one upgrade-info is stored. Don't we need to do store all the upgrade info for archiving history?

Yes. I don't think history is necessary now.
I think history can be viewed as a governance.

proto/panacea/oracle/v2alpha2/oracle.proto
gyuguen: This was simply copied from x/upgrade's proto and moved there.
This setting creates an equal function.

Since there is this setting in SoftwareUpgradeProposal, it is applied the same, and the Plan struct used internally also needs equal.

```go
func (this *Plan) Equal(that interface{}) bool {
	if that == nil {
		return this == nil
	}

	that1, ok := that.(*Plan)
	if !ok {
		that2, ok := that.(Plan)
		if ok {
			that1 = &that2
		} else {
			return false
		}
	}
	if that1 == nil {
		return this == nil
	} else if this == nil {
		return false
	}
	if this.UniqueId != that1.UniqueId {
		return false
	}
	if this.Height != that1.Height {
		return false
	}
	return true
}
```

x/oracle/client/cli/txUpgradeOracleProposal.go
gyuguen: Thanks. I added it.(https://github.com/medibloc/panacea-core/pull/454/commits/d45982338677ec29cb075c64c43aa842c350f3e9)

inchori (approved):
LGTM 👍 Thanks!

proto/panacea/oracle/v2alpha2/oracle.proto
H4NLee: Oh the `Plan` struct used internally. Thanks for your explanation!

x/oracle/proposal_handler.go
```diff
@@ -0,0 +1,37 @@
...
```
H4NLee, line 23: Wait. When this proposal is passed, the uniqueID in `x/oracle` module params should be changed at target height. Then, I think we need to add the logic to `x/oracle` end blocker. Am I right?

gyuguen: Yes, that's right. Development for endblocker is going on in another PR.

gyuguen: When the upgrade height is reached, the params of x/oracle will be modified and the contents of OracleUpgradeInfo will be cleared.

H4NLee (approved):
LGTM

audtlr24 (approved):
Thanks, LGTM!

PR #460, opened 2022-10-13
feat: added a oracle upgrade transaction

### LINKED ISSUE #428: Implement OracleUpgrade transaction
## Background
If `OracleUpgradeProposal` is passed, the chain must implement a transaction so that the oracle of the new binary can be registered.

## Implement
Add the following message to proto.
* The type must be added to `OracleRegistration`
  * We have to distinguish whether this is a new Oracle or an Oracle to be upgraded.
```protobuf
message MsgUpgradeOracle {
  string unique_id = 1;
  string oracle_address = 2;
  bytes node_pub_key = 3;
  bytes node_pub_key_remote_report = 4;
  int64 trusted_block_height = 5;
  bytes trusted_block_hash = 6;
  bytes nonce = 7;
}

message MsgUpgradeOracleResponse {

}

message OracleRegistration {
   string unique_id = 1;
  string address = 2;
  bytes node_pub_key = 3;
  bytes node_pub_key_remote_report = 4;
  int64 trusted_block_height = 5;
  bytes trusted_block_hash = 6;
  bytes encrypted_oracle_priv_key = 7;
  OracleRegistrationStatus status = 8;
  VotingPeriod voting_period = 9;
  TallyResult tally_result = 10;
  bytes nonce = 11;

  // Type of registration through which path
  OracleRegistrationType type = 12;
}

enum OracleRegistrationType {
  option (gogoproto.goproto_enum_prefix) = false;
  ORACLE_REGISTRATION_TYPE_UNSPECIFIED = 0;
  ORACLE_REGISTRATION_TYPE_NEW = 1;
  ORACLE_REGISTRATION_TYPE_UPGRADE = 2;
}
```

Implemented differently depending on whether OracleUpgrade is in progress.
* If the Oracle upgrade is not in progress, fail.
* If Oracle upgrade is in progress, follow the steps below.
  *  Compare if the UniqueID to be upgraded matches the UniqueID of MsgUpdateOracle.
  * If the verification passes, the Oracle information to be upgraded is stored in OracleRegistration KVStore.
  * If it is saved successfully, an event should be occured.
    * eventType: `oracle_upgrade`
    * attribute
      * unique_id
      * oracle_address
      * status: `started`

gyuguen:
Close #428 

## Implementation
* Validates and stores upgrade request message from oracle
* Since it performs a function similar to RegistrationOracle, it has been refactored.


x/oracle/keeper/upgrade.go
```diff
@@ -36,3 +38,32 @@ func (k Keeper) GetOracleUpgradeInfo(ctx sdk.Context) (*types.OracleUpgradeInfo,
...
+	oracleRegistration := msg.ToOracleRegistration()
+	compareFn := func(uniqueID string) error {
+		// check unique id
+		upgradeInfo, err := k.GetOracleUpgradeInfo(ctx)
+		if err != nil {
+			return err
+		}
+		if upgradeInfo.UniqueId != uniqueID {
+			return fmt.Errorf("The uniqueID to upgrade does not match. expected(%s), received(%s), ", upgradeInfo.UniqueId, msg.UniqueId)
+		}
```
H4NLee, line 52: Block height should also be checked.
It should be greater than `upgradeInfo.Height`.
Because this `UpgradeOracle` may cause new vote, and if the vote is passed, it will update the oracle info with new uniqueID. Then, the oracle's vote might be invalid.
Please check if I'm right. @gyuguen

H4NLee: I misunderstood. Please forget this comment. 

FYI, all the oracles should upgrade before reaching target height. If some oracles don't until target height, they have to re-register oracle with new version of oracle. [(ref comment)](https://github.com/medibloc/panacea-core/pull/454#discussion_r995379719)

x/oracle/keeper/upgrade.go
```diff
@@ -36,3 +38,32 @@ func (k Keeper) GetOracleUpgradeInfo(ctx sdk.Context) (*types.OracleUpgradeInfo,
 
 	return &upgradeInfo, nil
 }
+
+func (k Keeper) UpgradeOracle(ctx sdk.Context, msg *types.MsgUpgradeOracle) error {
+	oracleRegistration := msg.ToOracleRegistration()
```
audtlr24, line 43: IMO, the name of `ToOracleRegistration` is little bit confused. How about changing the name to `UpgradeOracleRegistration`?

audtlr24 (approved):
lgtm, thanks!
I left an unImportented comment

x/oracle/keeper/upgrade.go
gyuguen: Thank you comment.
This function is changing MsgUpgradeOracle to OracleRegistration.
Umm... I don't think this name would be interpreted differently.

audtlr24: Ah, I misunderstood. For now, I think we should keep the name as it is.

x/oracle/keeper/oracle.go
```diff
@@ -10,38 +10,51 @@ import (
...
+func (k Keeper) RegisterOracle(ctx sdk.Context, msg *types.MsgRegisterOracle) error {
+	oracleRegistration := types.NewOracleRegistration(msg)
+	compareFn := func(uniqueID string) error {
+		// check unique id
+		params := k.GetParams(ctx)
+		if params.UniqueId != uniqueID {
+			return fmt.Errorf("is not match the currently active uniqueID")
+		}
+		return nil
+	}
```
inchori, line 52: I have a question. Why the method is changed like this?

inchori (approved):
LGTM 👍

x/oracle/keeper/oracle.go
gyuguen: This is because RegisterOracle and UpgradeOracle have almost the same functionality. 
The only difference is that it compares uniqueIDs. 
RegisterOracle is taken from params of x/oracle. 
UpgradeOracle gets information from OracleUpgradeInfo KVStore.

Similar functions were combined into one, and only the other parts were passed to the function and imported.

inchori: I got it! Thank you for explanation!

proto/panacea/oracle/v2alpha2/oracle.proto
```diff
@@ -131,4 +133,27 @@ enum VoteOption {
...
+  option (gogoproto.goproto_enum_prefix) = false;
+
+  // UNSPECIFIED type
+  ORACLE_REGISTRATION_TYPE_UNSPECIFIED = 0;
+
+  // New oracle type
+  ORACLE_REGISTRATION_TYPE_NEW = 1;
+
+  // Upgrade oracle type
+  ORACLE_REGISTRATION_TYPE_UPGRADE = 2;
```
youngjoon-lee, line 158:
```suggestion
  // For oracles newly added 
  ORACLE_REGISTRATION_TYPE_NEW = 1;

  // For oracles to be upgraded
  ORACLE_REGISTRATION_TYPE_UPGRADE = 2;
```
Good to have these types :) I just added a simple comment for comments :)

gyuguen: Thank you!

youngjoon-lee (approved):
lgtm

PR #463, opened 2022-10-14
fix: change to hex decoding for trusted block hash

H4NLee:
Hex encoding/decoding is used for block hash, so I fixed it.


gyuguen (approved):
Thank you!

inchori (approved):
LGTM

PR #464, opened 2022-10-14
feat: implement voting consensus function for oracle upgrade

### LINKED ISSUE #429: Implement voting consensus function for Oracle upgrade request
## Background
We need to calculate the voting result to migrate the OraclePrivKey to the new binary oracle.

## Implement
The current implementation of OracleRegistration requires some modifications to the voting calculation.
The processing method is different depending on the type defined in OracleRegistration.
* `ORACLE_REGISTRATION_TYPE_NEW`: When the consensus is complete, a new Oracle is registered.
* `ORACLE_REGISTRATION_TYPE_UPGRADE`: When the consensus is complete, only the status of the existing oracle will change.

The rest will follow the process of existing OracleRegistration.

gyuguen:
Close #429 

### Implementation
When the Oracle registration consensus is finish, if the purpose is to upgrade, it will not be stored in the Oracle KVStore.
It will only change the status of the oracleRegistration


audtlr24 (approved):
lgtm!

inchori (approved):
LGTM !

youngjoon-lee (approved):
LGTM for now. In near future, I guess we need to abstract the voting & consensus logic for all other voting types such as sell-data, and so on.

gyuguen:
> LGTM for now. In near future, I guess we need to abstract the voting & consensus logic for covering all other voting types such as sell-data, and so on.

Yes, I agree :)

PR #466, opened 2022-10-16
feat: [v2.0.x] upgrade cosmos-sdk to v0.45.4

youngjoon-lee:
This PR is supposed to be merged to the `release/v2.0.x` branch, and be released with the tag `v2.0.5` after some additional PRs are merged as well.

Just cherry-picked from 664d5460e4c036ceb380cde0c0f74cb29c956458

TODO: should be tested if the chain can be upgraded smoothly from 2.0.3 to 2.0.5 without hard forks


app/app.go
```diff
@@ -720,44 +806,61 @@ func initParamsKeeper(appCodec codec.BinaryMarshaler, legacyAmino *codec.LegacyA
...
-
-				// call the before-modification hook since we're about to update the commission
-				app.StakingKeeper.BeforeValidatorModified(ctx, v.GetOperator())
-
-				app.StakingKeeper.SetValidator(ctx, v)
-			}
+// registerUpgradeHandlers registers upgrade handlers, and sets the store loader if necessary.
+// This function must be called before sealing the BaseApp (i.e. by app.LoadLatestVersion())
+// because the storetypes loader cannot be set if BaseApp is already sealed.
+func (app *App) registerUpgradeHandlers() error {
```
youngjoon-lee, line 818: If we upgrade our mainnet to v2.0.5 with this upgrade handler, I think we must remove the upgrade handler in the `master` branch which is similar as this.

app/app.go
```diff
@@ -560,7 +644,9 @@ func New(
 	app.ScopedTransferKeeper = scopedTransferKeeper
 	app.ScopedWasmKeeper = scopedWasmKeeper
 
-	app.registerUpgradeHandlers()
+	if err := app.registerUpgradeHandlers(); err != nil {
+		panic("Failed to register upgradeHandler: " + err.Error())
+	}
```
H4NLee, line 649: AFAIK, `registerUpgradeHandlers()` might be called before sealing.

gyuguen (approved):
LGTM

H4NLee:
When upgrading to v0.45.9, configuration for grpc web is added to `app.toml`.
If needed, we may have to give some guide for that.

app/app.go
H4NLee: fixed in [5dbc7e6](https://github.com/medibloc/panacea-core/pull/466/commits/5dbc7e60f4a50044ed9c840d7eee5091a459feb3)

gyuguen: Thank you!

youngjoon-lee: Danke. I even forgot about it even though we had already handled this issue on the master branch.

youngjoon-lee:
> When upgrading to v0.45.9, configuration for grpc web is added to app.toml.
If needed, we may have to give some guide for that.

Yeah. I'll put a note in the gov proposal, that the grpc-web will be enabled by default.

H4NLee (approved):
lgtm

PR #469, opened 2022-10-17
feat: distribute data sale reward to seller

### LINKED ISSUE #434: Implement distribution of reward to seller
## Background

When selling data, the seller can receive reward from deal after the verification of the data is completed.

## Implement

When `DataVerificationVote` is passed in end blocker, the seller will receive the reward except oracle commission.
The amount which seller receive will be `Budget/MaxNumData*(1-OracleCommission)`

H4NLee:
Close #434 

FYI, the `DistributeOracleRewards` will be refactored and used in `DistributeVerificationRewards` in the next PR.


x/datadeal/keeper/reward_test.go
```diff
@@ -85,6 +85,55 @@ func (suite *rewardTestSuite) BeforeTest(_, _ string) {
...
+		DealId:        1,
+		VerifiableCid: "verifiableCID",
+		SellerAddress: suite.sellerAccAddr.String(),
+	})
+
+	// distribute data verification rewards
+	err = suite.DataDealKeeper.DistributeVerificationRewards(ctx, dataSale)
+	suite.Require().NoError(err)
+
+	// TODO check balance of oracles
```
gyuguen, line 126: How about adding a test to verify the current Oracle's balance?
If the Oracle reward distribution logic is added later, this test will fail.

gyuguen (approved):
Thank you!
I left one comment.

x/datadeal/keeper/reward_test.go
H4NLee: Right. If the distribution of oracle rewards is added, the balance of deal would be empty.
So I marked as TODO here.
I'll cover it when I add oracle reward distribution in the next PR.

youngjoon-lee (commented):
lgtm

PR #472, opened 2022-10-19
feat: implement oracle upgrade in endblocker

### LINKED ISSUE #430: Implement oracle upgrade in endblocker
## Background
When the OracleUpgradeProposal passes and the block's height reaches the upgrade height, the chain must perform Oracle's upgrade operation.
Once the Oracle upgrade is complete, the old Oracles can no longer participate in voting, and only the Oracles to be upgraded can vote.

## Implement
In the endblocker of the oracle module, every block, query the OracleUpgradeInfo KVStore to see if there is any upgrade information.
If Oracle upgrade information exists and the block height reaches the upgrade height, the following upgrade operation is performed.
* Modify the oracle module's param to the `UniqueID` to be upgraded.
* Trigger an event
  * eventType: `oracle_upgrade`
  * attribute
    * uniqueID

gyuguen:
Close #430 

## Implement
* Check if oracleUpgradeInfo exists in every block
* Modify uniqueID in params if oracleUpgradeInfo exists and reached block height
* Delete oracleUpgradeInfo after upgrade is completed


audtlr24 (approved):
Thanks, LGTM!

x/oracle/keeper/upgrade.go
```diff
@@ -67,3 +73,20 @@ func (k Keeper) UpgradeOracle(ctx sdk.Context, msg *types.MsgUpgradeOracle) erro
 	)
 	return nil
 }
+
+func (k Keeper) ApplyUpgrade(ctx sdk.Context, info *types.OracleUpgradeInfo) error {
```
H4NLee, line 77: Why don't we add a log when oracle upgrade is applied?

H4NLee (approved):
lgtm

x/oracle/keeper/upgrade.go
gyuguen: Thank you!
I added a log!(https://github.com/medibloc/panacea-core/pull/472/commits/b88c8912f70630aa1fd1ea9deb72198d6893870f)

inchori (approved):
lgtm

PR #473, opened 2022-10-19
feat: add distribution of oracle rewards for data verification

H4NLee:
Based on #469 
Added oracle rewards distribution for data verification. Tests have been modified accordingly.

For simplicity, 50% of oracle commission is distributed to oracles who cast valid vote.
The rest of commission (rest of 50%) will be distributed when data delivery vote is passed.


audtlr24 (approved):
LGTM! 👍 
`distributeOracleRewards` in `DataDelivery` is applied in the next PR, right?

And, there is one thing I would like to ask for your opinion. How about defining `NewDecWithPrec(5, 1)` for 50% of oracle rewards as a const value. It can be used in `DataDelivery` oracle rewards.

H4NLee:
> `distributeOracleRewards` in DataDelivery is applied in the next PR, right?

Yes. `distributeOracleRewards` will also be used to distribute rewards for data delivery and it will be implemented in the next PR. 

> And, there is one thing I would like to ask for your opinion. How about defining NewDecWithPrec(5, 1) for 50% of oracle rewards as a const value. It can be used in DataDelivery oracle rewards.

Ok. it is done in [eb001cb](https://github.com/medibloc/panacea-core/pull/473/commits/eb001cb4a3358ce61d83e71cebfa8e8a796389ee)

gyuguen (approved):
LGTM

youngjoon-lee (commented):
lgtm. danke

PR #474, opened 2022-10-20
fix: add a deal ID when emit sell data event

inchori:
Add a deal ID when emit sell data event. So `doracle` can get data sale from panacea.


audtlr24 (approved):
Thansk, lgtm!

gyuguen (approved):
LGTM

H4NLee (approved):
lgtm

PR #476, opened 2022-10-21
fix: OracleUpgradeInfo in genesis to be nullable

gyuguen:
## Implement
OracleUpgradeInfo in genesis has been modified to be nullable.
If OracleUpgradeInfo is not nullable, the initial genesis.json contains a specific non-null empty value.
After the upgrade is completed, OracleUpgradeInfo KVstore is deleted and becomes null, so it is necessary to meet this specification.


audtlr24 (approved):
LGTM!

H4NLee (approved):
It makes sense

PR #481, opened 2022-10-27
fix: remove validation encryptedOraclePrivKey in oracleRegistrationVote

gyuguen:
## Implement
If oracle voted no, `encryptedOraclePrivKey` should be null.
Current our validation requires `encryptedOraclePrivKey`.
So I removed validation a `encryptedOraclePrivKey`


audtlr24 (approved):
Good, Thanks!

PR #486, opened 2022-11-01
feat: add check seller address when SellData

### LINKED ISSUE #462: Add sellerAddress check In SellData
Implement `sellerAddress` check in `SellData` for already existing `dataSale`.
- Currently, the `SellData` request with the same key(dealID + dataHash) is being rejected.
- For resale requests for data, enable request from `dataSale` with status `DATA_SALE_STATUS_VERIFICATION_FAILED`
- Add `sellerAddress` check logic to prevent malicious users put the `dataSale` into voting period using meta information of `dataSale`.

inchori:
Closes #462 
- Add check `dataSale` status and enable reselling data if the sellerAddress equals.
- Passed in unit test.

### DIFF
--- a/x/datadeal/keeper/deal.go
+++ b/x/datadeal/keeper/deal.go
@@ -1,6 +1,7 @@
 package keeper
 
 import (
+	"errors"
 	"fmt"
 	"strconv"
 
@@ -173,11 +174,6 @@ func (k Keeper) IncrementCurNumDataAtDeal(ctx sdk.Context, dealID uint64) error 
 }
 
 func (k Keeper) SellData(ctx sdk.Context, msg *types.MsgSellData) error {
-	_, err := sdk.AccAddressFromBech32(msg.SellerAddress)
-	if err != nil {
-		return err
-	}
-
 	deal, err := k.GetDeal(ctx, msg.DealId)
 	if err != nil {
 		return sdkerrors.Wrapf(types.ErrSellData, err.Error())
@@ -187,8 +183,7 @@ func (k Keeper) SellData(ctx sdk.Context, msg *types.MsgSellData) error {
 		return sdkerrors.Wrapf(types.ErrSellData, "deal status is not ACTIVE")
 	}
 
-	err = k.checkDataSaleStatus(ctx, msg.SellerAddress, msg.DataHash, msg.DealId)
-	if err != nil {
+	if err := k.checkDataSaleStatus(ctx, msg.SellerAddress, msg.DataHash, msg.DealId); err != nil {
 		return sdkerrors.Wrapf(types.ErrSellData, err.Error())
 	}
 
@@ -214,15 +209,21 @@ func (k Keeper) SellData(ctx sdk.Context, msg *types.MsgSellData) error {
 }
 
 func (k Keeper) checkDataSaleStatus(ctx sdk.Context, sellerAddress, dataHash string, dealID uint64) error {
-	getDataSale, _ := k.GetDataSale(ctx, dataHash, dealID)
+	getDataSale, err := k.GetDataSale(ctx, dataHash, dealID)
+
+	if err != nil {
+		if errors.Is(types.ErrDataSaleNotFound, err) {
+			return nil
+		} else {
+			return err
+		}
+	}
 
 	if getDataSale != nil {
 		if getDataSale.Status != types.DATA_SALE_STATUS_VERIFICATION_FAILED {
 			return fmt.Errorf("data already exists")
-		} else {
-			if getDataSale.SellerAddress != sellerAddress {
-				return fmt.Errorf("invalid data seller address")
-			}
+		} else if getDataSale.SellerAddress != sellerAddress {
+			return fmt.Errorf("invalid data seller address")
 		}
 	}
 
--- a/x/datadeal/keeper/deal.go
+++ b/x/datadeal/keeper/deal.go
@@ -187,9 +187,9 @@ func (k Keeper) SellData(ctx sdk.Context, msg *types.MsgSellData) error {
 		return sdkerrors.Wrapf(types.ErrSellData, "deal status is not ACTIVE")
 	}
 
-	getDataSale, _ := k.GetDataSale(ctx, msg.DataHash, msg.DealId)
-	if getDataSale != nil && getDataSale.Status != types.DATA_SALE_STATUS_VERIFICATION_FAILED {
-		return sdkerrors.Wrapf(types.ErrSellData, "data already exists")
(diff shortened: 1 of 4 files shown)


audtlr24 (approved):
LGTM! And, It would be good to make a test code for a case that fails because the sellerAddress is different.

x/datadeal/keeper/deal.go
```diff
@@ -187,9 +187,9 @@ func (k Keeper) SellData(ctx sdk.Context, msg *types.MsgSellData) error {
 		return sdkerrors.Wrapf(types.ErrSellData, "deal status is not ACTIVE")
 	}
 
-	getDataSale, _ := k.GetDataSale(ctx, msg.DataHash, msg.DealId)
-	if getDataSale != nil && getDataSale.Status != types.DATA_SALE_STATUS_VERIFICATION_FAILED {
-		return sdkerrors.Wrapf(types.ErrSellData, "data already exists")
+	err = k.checkDataSaleStatus(ctx, msg.SellerAddress, msg.DataHash, msg.DealId)
+	if err != nil {
+		return sdkerrors.Wrapf(types.ErrSellData, err.Error())
```
gyuguen, line 192: This is only code style, I prefer below :)
```suggestion
	
	if err := k.checkDataSaleStatus(ctx, msg.SellerAddress, msg.DataHash, msg.DealId); err != nil {
		return sdkerrors.Wrapf(types.ErrSellData, err.Error())
```

gyuguen (commented):
Thank you.
Check my comment please :)

x/datadeal/keeper/deal.go
```diff
@@ -213,6 +213,22 @@ func (k Keeper) SellData(ctx sdk.Context, msg *types.MsgSellData) error {
 	return nil
 }
 
+func (k Keeper) checkDataSaleStatus(ctx sdk.Context, sellerAddress, dataHash string, dealID uint64) error {
+	getDataSale, _ := k.GetDataSale(ctx, dataHash, dealID)
```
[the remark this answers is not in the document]
inchori: Because if the first data requested to sell, there will be return an error.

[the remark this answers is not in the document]
H4NLee: Then, I think it would be good to handle errors explicitly.

```suggestion
	getDataSale, err := k.GetDataSale(ctx, dataHash, dealID)
	if err != nil {
	    if errors.Is(types.ErrDataSaleNotFound, err) {
	    	    return nil
	    } else {
		    return err
	    }
	}
	
```

x/datadeal/keeper/deal.go
```diff
@@ -213,6 +213,22 @@ func (k Keeper) SellData(ctx sdk.Context, msg *types.MsgSellData) error {
...
+
+	if getDataSale != nil {
+		if getDataSale.Status != types.DATA_SALE_STATUS_VERIFICATION_FAILED {
+			return fmt.Errorf("data already exists")
+		} else {
+			if getDataSale.SellerAddress != sellerAddress {
+				return fmt.Errorf("invalid data seller address")
+			}
+		}
```
H4NLee, line 226:
```suggestion
		} else if getDataSale.SellerAddress != sellerAddress {
			return fmt.Errorf("invalid data seller address")
		}
```

x/datadeal/keeper/deal_test.go
```diff
@@ -183,7 +183,7 @@ func (suite *dealTestSuite) TestSellDataSuccess() {
 	suite.Require().Equal(dataSale.DataHash, suite.dataHash1)
 }
 
-func (suite *dealTestSuite) TestSellDataStatusFailed() {
+func (suite *dealTestSuite) TestReSellData() {
```
H4NLee, line 186: Good.
I think it would be good to check the desired events are emitted.

H4NLee (commented):
BTW, we don't need [line 176-179 validation](https://github.com/medibloc/panacea-core/pull/486/files#diff-da8d47c4b3fd413a36c0e0f5bdb94f8dbc05a194a544090542d8b28bd72fc76eL176-L179) because it is done in `msg.ValidateBasic()`

x/datadeal/keeper/deal.go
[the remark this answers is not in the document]
gyuguen: I agree :)

gyuguen (approved):
LGTM

x/datadeal/keeper/deal_test.go
inchori: Fixed in [9be3da9](https://github.com/medibloc/panacea-core/pull/486/commits/9be3da9de8b7d7787d0cecf1a4c8d022d5dd02b3).

x/datadeal/keeper/deal.go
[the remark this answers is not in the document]
inchori: Fixed in [9be3da9](https://github.com/medibloc/panacea-core/pull/486/commits/9be3da9de8b7d7787d0cecf1a4c8d022d5dd02b3).

x/datadeal/keeper/deal_test.go
```diff
@@ -174,13 +174,22 @@ func (suite *dealTestSuite) TestSellDataSuccess() {
...
 	suite.Require().Equal(dataSale.DataHash, suite.dataHash1)
+
+	events := suite.Ctx.EventManager().Events()
+
+	suite.Require().Equal(3, len(events[0].Attributes))
+	suite.Require().Equal(types.EventTypeDataVerificationVote, events[0].Type)
+	suite.Require().Equal(types.AttributeKeyVoteStatus, string(events[0].Attributes[0].Key))
+	suite.Require().Equal(types.AttributeValueVoteStatusStarted, string(events[0].Attributes[0].Value))
+	suite.Require().Equal(types.AttributeKeyDealID, string(events[0].Attributes[1].Key))
+	suite.Require().Equal(types.AttributeKeyDataHash, string(events[0].Attributes[2].Key))
```
H4NLee, line 192: Please refer the previous tests for event.
If some event is added, the order might be changed. So I suggest this way.

https://github.com/medibloc/panacea-core/blob/38cb583403c3a81fd9a6bf1484cf67bd0e4dc132/x/datadeal/abci_test.go#L237

inchori: Yes, I saw this test code. But the there has an only one event type in `SellData`. So I just wrote like this.

H4NLee: Ok. Let's keep this way, then.

H4NLee (approved):
good 👍

PR #488, opened 2022-11-02
feat: add event a unique id to registration oracle

gyuguen:


inchori (approved):
LGTM Thanks!

audtlr24 (approved):
LGTM

H4NLee (approved):
lgtm
BTW, the uniqueID is required to vote oracle registration?

gyuguen:
> lgtm BTW, the uniqueID is required to vote oracle registration?

I think oracle validates and using uniqueID as below.
* Compare my uniqueID and event's uniqueID
* Execute vote trasnaction using event's uniqueID

If event's uniqueID is not matched oracle's uniqueID, oracle shoud vote `No` as event's uniqueID.
So I added uniqueID in registrationOracle event.

H4NLee:
I got it. Thank you for your explanation.

In my opinion, it's enough to compare my uniqueID(`e.reactor.EnclaveInfo().UniqueIDHex()`) and oracle registration's uniqueID(`oracleRegistration.UniqueId`) which is verified by light client.

gyuguen:
Thanks comment,
Umm, I think if get oracleRegistration by oracle's uniqueID is failed, the oracle confuse how using unqueID.(Using oracle's uniqueID?)

H4NLee:
If oracle fails to get `OracleRegistration`, oracle can simply vote `NO`.
No more verification is needed I think.

gyuguen:
Yes, What you said about the verification is correct.

The uniqueID parameter is required when Oracle runs the vote.
I meant that what kind of using uniqueID. :)

oracle's uniqueID?
event's uniqueID?

H4NLee:
> The uniqueID parameter is required when Oracle runs the vote.
I meant that what kind of using uniqueID. :)
oracle's uniqueID?
event's uniqueID?

oracle's uniqueID.
I think oracle "should" use their uniqueID for every vote. The validity uniqueID will be checked in panacea chain by comparing with current unique ID.

gyuguen:
> > The uniqueID parameter is required when Oracle runs the vote.
> > I meant that what kind of using uniqueID. :)
> > oracle's uniqueID?
> > event's uniqueID?
> 
> oracle's uniqueID. I think oracle "should" use their uniqueID for every vote. The validity uniqueID will be checked in panacea chain by comparing with current unique ID.

Yes. I got it.
So, if the event's unqiueID and oracle's unqiueID are different, do you think Oracle should vote with oracle's uniqueID?

gyuguen:
Of course, this is unlikely to happen. :)

H4NLee:
> So, if the event's unqiueID and oracle's unqiueID are different, do you think Oracle should vote with oracle's uniqueID?

Right. I think it's like a signature of binary.

gyuguen:
OK. Thank you.
I will apply your discussion.
And apart from this discussion, I think add uniqueID is not bad in event.

H4NLee:
> And apart from this discussion, I think add uniqueID is not bad in event.

I agree with you 👍

PR #489, opened 2022-11-03
feat: added voterUniqueID and validate to compare activation uniqueID

gyuguen:
## Implement
Current, we don't compare an active uniqueID to a voter uniqueID. 
So old oracles are voteable.

I improved it to compare voter's uniqueID and activate uniqueID.


inchori (approved):
LGTM

audtlr24 (approved):
LGTM!

H4NLee (approved):
lgtm

PR #499, opened 2022-11-09
feat: added event when upgrade completion

gyuguen:
## Implement
An event should occur when OracleUpgrade completes.
So I added event like below:
```
oracle_upgrade.vote_status='upgrade_ended'
```


audtlr24 (approved):
lgtm!

H4NLee (approved):
LGTM.
BTW, does doracle subscribe this event? Or just to inform?

gyuguen:
> LGTM. BTW, does doracle subscribe this event? Or just to inform?

Oracle should be subscribe this event.
Because if changed uniqueID in blockchain, old oracle is ignored vote and new oracle is starting vote.

Issue #492, opened 2022-11-03
Upgrade DeactivateDeal function


inchori:
When buyer deactivated the deal, the remained vote need to be removed.
- Implement a `RemoveAllDataVerificationQueue` and `RemoveAllDataDeliveryQueue` in `keeper.go`

audtlr24:
When buyer deactivated the deal, `DataVerification/Delivery` queue does not removed.

Instead, deal status become `INACTIVE_REQUESTED`, and in this state, the previously requested `sellData` will be voted. When the deal status is `INACTIVE_REQUESTED`, additional `sellData` requests will be rejected.

After all previously requested `sellData` have been voted, deal status become `INACTIVE` and buyer can get a refund remaining coins in the deal account.

So, let's implement the following.
- change `DeactivateDeal` keeper function
- implement check function that all previously requested `sellData` have been voted or not
- implement refund function when all previously requested `sellData` have been voted
- applying above functions

audtlr24:
There are some changes like below:
- deal status `INACTIVE_REQUESTED` renamed `DEACTIVATING`
- deal status `INACTIVE` renamed `DEACTIVATED`
- when the deal_status is `DEACTIVATING`, panacea calculate `deactivate_height` by `cur_height + k * voting_period` (the `k` is set by params)
-  when the block reach `deactivate_height`, deal status become `DEACTIVATED` and all the remaining coins in deal account refund to buyer

youngjoon-lee:
I think we can close this issue because this is related with the `panacea-doracle` (deprecated).

PR #502, opened 2022-11-10
feat: split DeactivateDeal to two phase

audtlr24:
To solve issue #492, split `DeactivateDeal` to `RequestDeactivateDeal` and `DeactivateDeal`.

When a `DeactivateDeal` request is received, it first enters the `DEAL_STATUS_DEACTIVATING` status, and is added to the `DealQueue` with the deactivation height. 
`EndBlocker` handle `DealQueue` and change the deal status to DEAL_STATUS_DEACTIVATED with a refund when the deactivation height is reached.

There are remaining Todo.
- implement function that remove DataVerification/DeliveryVote Queue at `DeactivateDeal`
- change `DealDeactivationParam` to datadeal params
- add `DealQueue` and datadeal params to genesis state


gyuguen (commented):
Thank you.
Please check my comment.

x/datadeal/keeper/keeper.go
```diff
@@ -173,3 +173,38 @@ func (k Keeper) IterateClosedDataDeliveryQueue(ctx sdk.Context, endTime time.Tim
...
 }
+
+func (k Keeper) AddDealQueue(ctx sdk.Context, dealID uint64, deactivationHeight int64) {
+	store := ctx.KVStore(k.storeKey)
+	store.Set(types.GetDealQueueKey(dealID, deactivationHeight), sdk.Uint64ToBigEndian(dealID))
+}
+
+func (k Keeper) GetClosedDealQueueIterator(ctx sdk.Context, deactivationHeight int64) sdk.Iterator {
+	store := ctx.KVStore(k.storeKey)
+	return store.Iterator(types.DealQueueKey, sdk.PrefixEndBytes(types.GetDealQueueByHeight(deactivationHeight)))
```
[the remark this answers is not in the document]
audtlr24: I tested this case, but there is no error (When the block height is 20, the deal is not fetched, when the block height is 100, the deal is actually fetched)

x/datadeal/keeper/deal.go
```diff
@@ -544,45 +550,82 @@ func (k Keeper) DeactivateDeal(ctx sdk.Context, msg *types.MsgDeactivateDeal) er
...
+
+	err = k.SetDeal(ctx, deal)
+	if err != nil {
+		return sdkerrors.Wrapf(types.ErrDealDeactivate, err.Error())
+	}
+
+	params := k.oracleKeeper.GetParams(ctx)
+	VotingPeriod := params.VoteParams.VotingPeriod
+
+	deactivationHeight := ctx.BlockHeader().Height + DealDeactivationParam*int64(VotingPeriod/BlockPeriod)
```
gyuguen, line 563: I think it would be good add more block height. (+1 or higher)
```suggestion
	deactivationHeight := ctx.BlockHeader().Height + DealDeactivationParam*int64(VotingPeriod/BlockPeriod) +1
```

x/datadeal/keeper/keeper.go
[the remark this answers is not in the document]
gyuguen: OK. good.
Thank you

x/datadeal/keeper/deal.go
audtlr24: Okay, Thanks. It's good because it seems to have a minimum limit.

x/datadeal/keeper/deal.go
```diff
@@ -544,45 +550,82 @@ func (k Keeper) DeactivateDeal(ctx sdk.Context, msg *types.MsgDeactivateDeal) er
...
 
-	err = k.bankKeeper.SendCoins(ctx, dealAcc, requesterAcc, sdk.Coins{dealBalance})
+	err = k.bankKeeper.SendCoins(ctx, dealAcc, buyerAcc, sdk.Coins{dealBalance})
 	if err != nil {
 		return sdkerrors.Wrapf(types.ErrDealDeactivate, err.Error())
 	}
 
-	//TODO: Implement Remove the DataVerification/DeliveryVote Queue after PR #449 merged
-	//https://github.com/medibloc/panacea-core/pull/449
+	//Todo:Remove DataVerification/DeliveryVote Queue
```
gyuguen, line 606: Please remove this comment.
We discussed that the voting queue is not being removed.

x/datadeal/keeper/deal.go
```diff
@@ -544,45 +550,82 @@ func (k Keeper) DeactivateDeal(ctx sdk.Context, msg *types.MsgDeactivateDeal) er
...
 func (k Keeper) ReRequestDataDeliveryVote(ctx sdk.Context, msg *types.MsgReRequestDataDeliveryVote) error {
 
+	deal, err := k.GetDeal(ctx, msg.DealId)
+	if err != nil {
+		return sdkerrors.Wrapf(types.ErrReRequestDataDeliveryVote, err.Error())
+	}
+
+	if deal.Status != types.DEAL_STATUS_ACTIVE && deal.Status != types.DEAL_STATUS_COMPLETED {
+		return sdkerrors.Wrapf(types.ErrReRequestDataDeliveryVote, "not in deal state to make a request")
+	}
```
gyuguen, line 620: We ensure that we receive all requested sales data.
So I think that we don't need check deal status.

x/datadeal/keeper/deal.go
```diff
@@ -14,6 +15,11 @@ import (
 	oracletypes "github.com/medibloc/panacea-core/v2/x/oracle/types"
 )
 
+const (
+	BlockPeriod           = 6 * time.Second
```
H4NLee, line 19: I suggest to make the `blockTimeSec` in `cmd/init.go` public and use it?

x/datadeal/keeper/deal.go
```diff
@@ -14,6 +15,11 @@ import (
...
+	DealDeactivationParam = int64(3) // Todo: make DealDeactivationParam to Params of DataDeal
```
H4NLee, line 20: I think use much appropriate variable name such as `DeactivationPeriodMultiplier`.
If you are gonna move this to module params, it doesn't matter for now.

H4NLee (approved):
Left minor comments

x/datadeal/keeper/deal.go
H4NLee: AFAIK, we need to check the status because there is no budget for data delivery in deactivated deal.

gyuguen: I forgot this spec!
Please ignore my comment.
Thank you.

x/datadeal/keeper/deal.go
audtlr24: Okay, thanks. I change to use it.

audtlr24: When I checked because the CI test failed, an import cycle error occurred in that case. So, I reverted the commit. I think it looks good to keep it this way.

x/datadeal/keeper/deal.go
audtlr24: The naming you suggested looks better. If it's not important, I think it's okay to change it all at once in refactoring later.

x/datadeal/keeper/deal_test.go
```diff
@@ -873,3 +865,57 @@ func (suite *dealTestSuite) TestRequestDataDeliveryVoteFailedInvalidStatus() {
...
+		),
+		"deal"+strconv.FormatUint(deactivatingDeal.Id, 10)),
+	)
+	suite.AccountKeeper.SetAccount(ctx, acc)
+
+	err = suite.BankKeeper.SendCoins(ctx, suite.buyerAccAddr, dealAddress, coins)
+	suite.Require().NoError(err)
+
+	err = suite.DataDealKeeper.SetDeal(ctx, deactivatingDeal)
+	suite.Require().NoError(err)
```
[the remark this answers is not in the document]
audtlr24: It seems like it could be done simpler. Thanks!

PR #503, opened 2022-11-11
feat: add function that remove DataVerification/DeliveryVote Queue

audtlr24:
Add `RemoveVotingQueue` function that remove all DataVerification/DeliveryVote Queue related deactivating deal.

Minor changes:
- The order of key parameters is uniformly matched.

### DIFF
--- a/x/datadeal/keeper/deal.go
+++ b/x/datadeal/keeper/deal.go
@@ -238,7 +238,7 @@ func (k Keeper) checkDataSaleStatus(ctx sdk.Context, sellerAddress, dataHash str
 
 func (k Keeper) GetDataSale(ctx sdk.Context, dataHash string, dealID uint64) (*types.DataSale, error) {
 	store := ctx.KVStore(k.storeKey)
-	key := types.GetDataSaleKey(dataHash, dealID)
+	key := types.GetDataSaleKey(dealID, dataHash)
 
 	bz := store.Get(key)
 	if bz == nil {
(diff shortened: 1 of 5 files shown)


x/datadeal/keeper/deal.go
```diff
@@ -603,7 +609,36 @@ func (k Keeper) DeactivateDeal(ctx sdk.Context, dealID uint64) error {
 		return sdkerrors.Wrapf(types.ErrDealDeactivate, err.Error())
 	}
 
-	//Todo:Remove DataVerification/DeliveryVote Queue
+	return nil
+}
+
+func (k Keeper) RemoveVotingQueue(ctx sdk.Context, DeactivateDealID uint64) error {
```
gyuguen, line 615: I think it would be good not to remove the voting queue. Because this code will run very slowly if the  voting queue contains a lot of data.(searched all data).

Umm. How about ignoring deactive deal data as an endblocker.

gyuguen (commented):
Check my comment please.

x/datadeal/keeper/deal.go
audtlr24: I think the method you suggested is better. It would be better to check the deal status in the verfication/delivery endblocker and ignore it when it is in the deactivate related state.

gyuguen: So, I think this PR can be closed.
What do you think?

audtlr24: Okay, sure. I closed this PR.

PR #504, opened 2022-11-11
feat: add datadeal params and DealQueue at genesis state

audtlr24:
- add datadeal params containing `dealDeactivationParam`.
- add `DealQueue` to genesis state


x/datadeal/types/keys.go
```diff
@@ -105,8 +105,8 @@ func GetDealQueueByHeight(deactivationHeight int64) []byte {
 	return append(DealQueueKey, sdk.Uint64ToBigEndian(uint64(deactivationHeight))...)
 }
 
-func SplitDealQueueKey(key []byte) uint64 {
-	return sdk.BigEndianToUint64(key[10:])
+func SplitDealQueueKey(key []byte) (int64, uint64) {
```
gyuguen, line 108: I think it would be good to add comment what returns data name.(maybe votingEndBlockHeight, dealID?)

gyuguen (approved):
LGTM
I left one comment.

x/datadeal/types/keys.go
audtlr24: Ah, it might be confusing. I add it in [988dba4](https://github.com/medibloc/panacea-core/pull/504/commits/988dba40043d294d28c49287b060a59e1e2c4971).

H4NLee (approved):
utACK

PR #515, opened 2022-12-02
feat: define skeleton proto files

### LINKED ISSUE #508: Define x/oracle and x/datadeal with proto files
Based on a new `main` branch (branched from `release/v2.0.x`).

gyuguen:
Closes #508


proto/panacea/datadeal/v2/consent.proto
```diff
@@ -0,0 +1,16 @@
+syntax = "proto3";
```
gyuguen, line 1: I don't know this file name is correct.
certificate.go? consent.go? 
Please give me a good idea!

audtlr24 (approved):
lgtm!

proto/panacea/datadeal/v2/consent.proto
```diff
@@ -0,0 +1,22 @@
...
+// Certificate defines a certificate with signature
+message Certificate {
+  UnsignedCertificate unsigned_certificate = 1;
+  bytes signature = 2;
+}
+
+// UnsignedCertificate defines a certificate information
+message UnsignedCertificate {
+    string cid = 1;
+    string operator_address = 2;
```
H4NLee, line 18: why don't we use `oracle_address` to be clear?
I think we have many similar words can be confused: node, oracle, operator, validator, etc.
For oracle, it would be better to unify with the word `oracle`.

proto/panacea/datadeal/v2/deal.proto
```diff
@@ -0,0 +1,37 @@
...
+message Deal {
+  uint64 id = 1;
+  string address = 2;
+  repeated string data_schema = 3;
+  cosmos.base.v1beta1.Coin budget = 4;
+  uint64 max_num_data = 5;
+  uint64 cur_num_data = 6;
+  string consumer_address = 7;
+  DealStatus status = 8;
+  bytes nonce = 9;
```
H4NLee, line 19: Do we still need this nonce? I think it was required to make consensus between oracles.

proto/panacea/datadeal/v2/tx.proto
```diff
@@ -0,0 +1,54 @@
...
+  rpc SubmitConsent(MsgSubmitConsent) returns (MsgSubmitConsentResponse);
+}
+
+// MsgCreateDeal defines the Msg/CreateDeal request type.
+message MsgCreateDeal {
+  repeated string data_schema = 1;
+  cosmos.base.v1beta1.Coin budget = 2;
+  uint64 max_num_data = 3;
+  string buyer_address = 4;
+  bytes nonce =5;
```
H4NLee, line 29: Same here :)
If needed, please ignore this comment as well.

proto/panacea/oracle/v2/oracle.proto
```diff
@@ -0,0 +1,48 @@
+syntax = "proto3";
+package panacea.oracle.v2;
+
+option go_package = "github.com/medibloc/panacea-core/x/oracle/types";
+
+import "gogoproto/gogo.proto";
+
+// Oracle defines a detail of oracle.
+message Oracle {
+  string operator_address = 1;
```
H4NLee, line 10: How about `oracle_address` here too?

proto/panacea/oracle/v2/oracle.proto
```diff
@@ -0,0 +1,48 @@
...
+  OracleStatus status = 2;
```
H4NLee, line 11: I think this status is not needed. 
If it is for upgrade, we can use `unique_id` to know whether the oracles have been upgraded

proto/panacea/oracle/v2/oracle.proto
```diff
@@ -0,0 +1,48 @@
...
+  bytes node_pub_key_remote_report = 4;
+  // The trusted block info is required for light client.
+  // Other oracle can validate whether the oracle set correct trusted block info.
+  int64 trusted_block_height = 5;
+  bytes trusted_block_hash = 6;
+  string endpoint = 7;
+  string oracle_commission_rate = 8 [(gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec", (gogoproto.nullable) = false];
+  // Nonce required for AES256-GCM encryption
+  bytes nonce = 9;
+}
```
H4NLee, line 48: We may need to have status of `OracleRegistration`.
- pending : waiting for approval of oracle registration
- completed : received at least 1 approval of oracle registration

And also `encrypted_oracle_priv_key` may be required which is empty when `pending` status and updated when the first oracle approves.

H4NLee (commented):
Left some comments :)

proto/panacea/datadeal/v2/deal.proto
audtlr24: The nonce is required for `AES256` encryption. It is used to store the nonce used when performing encryption/decryption with combineKey. In deal, it is used to encrypt and decrypt the `certificate` with combineKey. It seems that there will be no problem if this value is `Params`, but in the previous implementation, each deal had a separate nonce.

proto/panacea/datadeal/v2/consent.proto
gyuguen: Yes, We used a lot of words for Oracle's address. 
We set this name to `operator_address` in the architecture discussion.
But I think `oracle_address` is better. :)

proto/panacea/datadeal/v2/deal.proto
gyuguen: And the CID of IPFS is always changed. Because encrypted data always changes when using a random nonce.

proto/panacea/oracle/v2/oracle.proto
gyuguen: It is used only for Oracle register information. We don't manage register oracle status.
We can also get encrypted data from `MsgApproveOracleRegistration` tx.
New oracles need to subscirbe to `MsgApproveOracleRegistration` tx

proto/panacea/oracle/v2/oracle.proto
gyuguen: You're right.
I think currently not required.
I will remove soon.

gyuguen: fixed: https://github.com/medibloc/panacea-core/pull/515/commits/8eccb41513d2d18b87ff343c73155def0844c388

proto/panacea/datadeal/v2/consent.proto
gyuguen: fixed: https://github.com/medibloc/panacea-core/pull/515/commits/8eccb41513d2d18b87ff343c73155def0844c388

proto/panacea/datadeal/v2/deal.proto
H4NLee: > The nonce is required for AES256 encryption. It is used to store the nonce used when performing encryption/decryption with combineKey. In deal, it is used to encrypt and decrypt the certificate with combineKey. It seems that there will be no problem if this value is Params, but in the previous implementation, each deal had a separate nonce.

We can encrypt/decrypt data using `AES256` with randomly generated nonce.
Since the data is validated by the only one oracle, I think the nonce is not required.
In doracle, if we don't use nonce, all the oracles encrypt the data using each nonce, leading to each different value of `encrypted_oracle_priv_key` and consensus can't be made.
However in this scalable oracle, we don't need any consensus so that I thought the randomly generated nonce can be used. If I miss something, please let me know :pray:

H4NLee: > And the CID of IPFS is always changed. Because encrypted data always changes when using a random nonce.

I don't understand the meaning of `the CID of IPFS is always changed`. Could you elaborate more??

gyuguen: Yes, We can use a random nonce as you said.
However I'm worried that the data stored in IPFS keeps changing CID.(Even though it is the same data).

H4NLee: Ah, I got your point.

H4NLee: Right. As you said, we'd better to use nonce in encryption. Thank you all!

third_party/proto/cosmos/staking/v1beta1/staking.proto
```diff
@@ -85,8 +85,8 @@ message Validator {
   option (gogoproto.goproto_stringer) = false;
   option (gogoproto.goproto_getters)  = false;
 
-  // operator_address defines the address of the validator's operator; bech encoded in JSON.
-  string              operator_address = 1 [(gogoproto.moretags) = "yaml:\"operator_address\""];
+  // oracle_address defines the address of the validator's operator; bech encoded in JSON.
+  string              oracle_address = 1 [(gogoproto.moretags) = "yaml:\"oracle_address\""];
```
H4NLee, line 89: Oh, `staking.proto` has also been changed. Please check again :)

gyuguen: Oh. I missed it. 
Thank you.

proto/panacea/oracle/v2/oracle.proto
H4NLee: I got it. The `encrypted_oracle_priv_key` will not be stored, just subscribing the Tx.

Then, multiple `MsgApproveOracleRegistration`s from multiple oracles can be executed successfully because no status or state is stored, right?

gyuguen: Exactly.

H4NLee (approved):
LGTM :+1:

inchori (approved):
LGTM 👍

youngjoon-lee (approved):

proto/panacea/datadeal/v2/consent.proto
gyuguen: OK. I changed name to `certificate.proto`.
Thank you

proto/panacea/datadeal/v2/deal.proto
gyuguen: Confirm. Remove the field 'nonce'.

PR #525, opened 2022-12-07
feat: add create a deal and query a deal keeper function

### LINKED ISSUE #512: Implement CreateDeal tx and Query a Deal
- Implement CreateDeal Tx and Query a Deals.
- Implement genesis state of Deal.

inchori:
Closes #512

- Add keeper functions of `CreateDeal` and query a `Deal`.
- The CLI function of `CreateDeal` and `Deal` will be implemented in next branch based on this branch.


audtlr24 (approved):
lgtm

x/datadeal/keeper/deal.go
```diff
@@ -0,0 +1,148 @@
...
+	deal, err := k.GetDeal(ctx, dealID)
+	if err != nil {
+		return false, err
+	}
+	if deal.Status == types.DEAL_STATUS_COMPLETED {
+		return true, nil
+	} else {
+		return false, nil
+	}
+}
```
H4NLee, line 133: Why don't we make this function like below?
```go
func (d Deal) IsCompleted() bool {
    return d.Status == types.DEAL_STATUS_COMPLETED
}
```

x/datadeal/keeper/deal.go
```diff
@@ -0,0 +1,148 @@
...
+
+func (k Keeper) IncrementCurNumDataAtDeal(ctx sdk.Context, dealID uint64) error {
```
H4NLee, line 135: Let's use verbs as function name
```suggestion
func (k Keeper) IncreaseCurNumDataOfDeal(ctx sdk.Context, dealID uint64) error {
```

H4NLee (approved):

x/datadeal/keeper/deal.go
inchori: If I change like this, does the function need to move under the `types` package?

youngjoon-lee: You can:
```
func (k Keeper) IsDealCompleted(ctx sdk.Context, dealID uint64) (bool, error) {
	deal, err := k.GetDeal(ctx, dealID)
	if err != nil {
		return false, err
	}
        return deal.Status == types.DEAL_STATUS_COMPLETED, nil
}
```
, if you wanna put both `k.GetDeal` and status check in one function.

x/datadeal/keeper/deal.go
```diff
@@ -0,0 +1,148 @@
...
+	if err := k.cdc.UnmarshalLengthPrefixed(bz, &val); err != nil {
+		return 0, sdkerrors.Wrapf(err, "Failed to get next deal number")
+	}
+
+	dealNumber = val.GetValue()
+
+	return dealNumber, nil
+}
+
+func (k Keeper) GetNextDealNumberAndIncrement(ctx sdk.Context) (uint64, error) {
```
youngjoon-lee, line 83: Usually, we name this kind of functions as `Verb1AndVerb2...`. So, we can rename this to `GetAndIncreaseNextDealNumber`.
https://en.wikipedia.org/wiki/Compare-and-swap

x/datadeal/keeper/deal.go
```diff
@@ -0,0 +1,148 @@
...
+
+func (k Keeper) CreateDeal(ctx sdk.Context, consumerAddress sdk.AccAddress, msg *types.MsgCreateDeal) (uint64, error) {
+	dealID, err := k.GetNextDealNumberAndIncrement(ctx)
+	if err != nil {
+		return 0, sdkerrors.Wrapf(err, "failed to get next deal num")
+	}
+
+	newDeal := types.NewDeal(dealID, msg)
+
+	coins := sdk.NewCoins(*msg.Budget)
```
youngjoon-lee, line 21: would be easier if this is called as `budget`.

youngjoon-lee (approved):
lgtm

x/datadeal/keeper/deal.go
inchori: I'll reflect these comments after the other derived branches are merged into this branch.

x/datadeal/keeper/deal.go
H4NLee: > If I change like this, does the function need to move under the types package?

It might be :)

gyuguen (approved):

x/datadeal/keeper/deal.go
```diff
@@ -0,0 +1,169 @@
...
+	"strconv"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	gogotypes "github.com/gogo/protobuf/types"
+	"github.com/medibloc/panacea-core/v2/x/datadeal/types"
+)
+
+func (k Keeper) CreateDeal(ctx sdk.Context, consumerAddress sdk.AccAddress, msg *types.MsgCreateDeal) (uint64, error) {
```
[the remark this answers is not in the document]
inchori: Hmm... I think that I referred to before version of `CreateDeal`. I'll fix it!

x/datadeal/keeper/deal.go
inchori: Fixed in [8077e85](https://github.com/medibloc/panacea-core/pull/525/commits/8077e85ac4a079ae1a55d5c181a0047c1cb22a22)

PR #532, opened 2022-12-07
feat: add a genesis state of Deal and CLI of Deals

inchori:
Add a genesis state of Deal and add a CLI of query `Deals`.


types/testsuite/suite.go
```diff
@@ -230,6 +230,8 @@ func newTestCodec() params.EncodingConfig {
 	authtypes.RegisterInterfaces(interfaceRegistry)
 	banktypes.RegisterInterfaces(interfaceRegistry)
 	didtypes.RegisterInterfaces(interfaceRegistry)
+	//TODO: When other Msgs are implemented, I'll remove this comment.
+	//datadealtypes.RegisterInterfaces(interfaceRegistry)
```
inchori, line 234: I didn't understand when I add this code, the test was failed. And the error message is like this:
```
 test panicked: type *types.MsgDeactivateDeal doesn't actually implement interface types.Msg
 ```
 So I think other Msg function is not implemented yet, so the errors were occurred. So for now, I add TODO to this part.

youngjoon-lee: It's natural because `*types.MsgDeactivateDeal` doesn't fit the requirement (spec) of the interface `types.Msg` yet. It's good that you commented this line out. We can uncomment it after implementing the `*types.MsgDeactivateDeal`.

youngjoon-lee (approved):
lgtm

H4NLee (approved):
lgtm

audtlr24 (approved):
lgtm!

PR #535, opened 2022-12-08
fix: remove duplicated oracle in app.go

H4NLee:
In current main branch, there are duplicated codes for oracle module.
When I try to init and start chain, I got an error like below.
`panic: Potential key collision between KVStores:oracle - oracle`


gyuguen (approved):
Thank you.
It seems that several PRs are mixed.

audtlr24 (approved):

PR #540, opened 2022-12-08
feat: add UpdateOracleInfo tx

### LINKED ISSUE #511: Implement UpdateOracleInfo tx
- To update endpoint and commission-rate
- SGX-based signature isn't needed.

audtlr24:
closes #511.
- add `oracle_commission_max_rate` and `oracle_commission_max_change_rate`
- check commission change (max_rate, max_change_rate, updateTime)
- change `RegisterOracle` and related code to include that commission part


x/oracle/client/cli/txRegisterOracle.go
```diff
@@ -16,9 +16,9 @@ import (
 
 func CmdRegisterOracle() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "register-oracle [unique ID] [node public key] [node public key remote report] [trusted block height] [trusted block hash] [endpoint] [oracle's commission rate]",
+		Use:   "register-oracle [unique-ID] [node-public-key] [node-public-key-remote-report] [trusted-block-height] [trusted-block-hash] [endpoint] [oracle-commission-rate] [oracle-commission-max-rate] [oracle-commission-max-change-rate]",
```
gyuguen, line 19: I implemented this code to receive input as an argument, but it was better to change the flag.
Because it doesn't look good.
wdyt?

x/oracle/client/cli/txUpdateOracleInfo.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/cosmos/cosmos-sdk/client/tx"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/medibloc/panacea-core/v2/x/oracle/types"
+	"github.com/spf13/cobra"
+)
+
+func CmdUpdateOracleInfo() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "update-oracle-info [endpoint] [oracle-commission-rate]",
```
gyuguen, line 16: I think this is similar. How about change to using flag?

x/oracle/keeper/oracle.go
```diff
@@ -42,6 +43,53 @@ func (k Keeper) RegisterOracle(ctx sdk.Context, msg *types.MsgRegisterOracle) er
...
+
+	case newRate.IsNegative():
+		// new rate cannot be negative
+		return types.ErrCommissionNegative
+
+	case newRate.GT(oracle.OracleCommissionMaxRate):
+		// new rate cannot be greater than the max rate
+		return types.ErrCommissionGTMaxRate
+
+	case newRate.Sub(oracle.OracleCommissionRate).GT(oracle.OracleCommissionMaxChangeRate):
```
gyuguen, line 85: I think you need to change like below:
```suggestion
	case newRate.Sub(oracle.OracleCommissionRate).Abs().GT(oracle.OracleCommissionMaxChangeRate):
```

x/oracle/keeper/oracle_test.go
```diff
@@ -206,17 +225,118 @@ func (suite *oracleTestSuite) TestRegisterOracleNotSameUniqueID() {
...
+	msgUpdateOracleInfo := &types.MsgUpdateOracleInfo{
+		OracleAddress:        suite.oracleAccAddr.String(),
+		Endpoint:             suite.newEndpoint,
+		OracleCommissionRate: sdk.NewDecWithPrec(3, 1),
+	}
+	ctx = ctx.WithBlockTime(ctx.BlockTime().Add(30 * time.Hour))
+	err = suite.OracleKeeper.UpdateOracleInfo(ctx, msgUpdateOracleInfo)
+	suite.Require().Error(err, types.ErrUpdateOracle)
+	suite.Require().ErrorContains(err, "commission cannot be more than the max rate")
+}
```
gyuguen, line 342: Please add test code to down commission.(e.g 10% -> 0%)

x/oracle/types/message_oracle.go
```diff
@@ -59,6 +61,12 @@ func (m *MsgRegisterOracle) ValidateBasic() error {
 	if m.OracleCommissionRate.LT(sdk.ZeroDec()) || m.OracleCommissionRate.GT(sdk.OneDec()) {
 		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "oracleCommissionRate must be between 0 and 1")
 	}
+	if m.OracleCommissionMaxRate.LT(sdk.ZeroDec()) || m.OracleCommissionMaxRate.GT(sdk.OneDec()) {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "OracleCommissionMaxRate must be between 0 and 1")
+	}
+	if m.OracleCommissionMaxChangeRate.LT(sdk.ZeroDec()) || m.OracleCommissionMaxChangeRate.GT(sdk.OneDec()) {
```
gyuguen, line 67: I think you should use `m.OracleCommissionMaxRate` as the maximum value.
```suggestion
	if m.OracleCommissionMaxChangeRate.LT(sdk.ZeroDec()) || m.OracleCommissionMaxChangeRate.GT(m.OracleCommissionMaxRate) {
```

gyuguen (commented):
Thank you
Please check my comment.

x/oracle/client/cli/txRegisterOracle.go
gyuguen: But it doesn't matter. :)

x/oracle/client/cli/txUpdateOracleInfo.go
inchori: How about using endpoint by argument and oracle commission rate to using flag?

x/oracle/client/cli/txRegisterOracle.go
audtlr24: Thanks, I changed it using flag ([ab40af0](https://github.com/medibloc/panacea-core/pull/540/commits/ab40af0819b45dee477db3cb1317e287c0a0efdd)).

x/oracle/client/cli/txUpdateOracleInfo.go
audtlr24: Thanks, I changed it using flag ([ab40af0](https://github.com/medibloc/panacea-core/pull/540/commits/ab40af0819b45dee477db3cb1317e287c0a0efdd)).

x/oracle/keeper/oracle.go
audtlr24: This part requires consideration of limitations when reducing the commission rate. In staking module, there is no limitation about this. 
In my opinion, when the commission rate is reduced, the provider does not lose money, so it would be okay to give this part as freedom.

x/oracle/client/cli/txRegisterOracle.go
```diff
@@ -101,6 +138,16 @@ func CmdRegisterOracle() *cobra.Command {
 		},
 	}
 
+	cmd.Flags().String(flagOracleUniqueID, "", "unique ID for oracle")
```
gyuguen, line 141: And you can set required values as below.
```go
if err := cmd.MarkFlagRequired(flagOracleUniqueID); error != nil {
    panic (error)
}
```

audtlr24: Thanks, I fixed it.

x/oracle/client/cli/txRegisterOracle.go
```diff
@@ -16,39 +16,70 @@ import (
...
+				return fmt.Errorf("failed to get oralce end point")
+			}
 
-			oracleCommissionRateStr := args[6]
+			oracleCommissionRateStr, err := cmd.Flags().GetString(flagOracleCommRate)
+			if err != nil {
+				return fmt.Errorf("failed to get oralce commission rate")
+			}
 
 			if len(oracleCommissionRateStr) == 0 {
```
H4NLee, line 84: There is no need to check the length of commission rate because `sdk.NewDecFromStr` will do.

x/oracle/client/cli/txRegisterOracle.go
```diff
@@ -59,15 +90,45 @@ func CmdRegisterOracle() *cobra.Command {
...
 			}
 
+			oracleCommissionMaxRateStr, err := cmd.Flags().GetString(flagOracleCommMaxRate)
+			if err != nil {
+				return fmt.Errorf("failed to get oralce commission max rate")
+			}
+
+			if len(oracleCommissionMaxRateStr) == 0 {
+				return fmt.Errorf("oracleCommissionMaxRate is empty")
+			}
```
H4NLee, line 100: Same here :)

x/oracle/client/cli/txRegisterOracle.go
```diff
@@ -59,15 +90,45 @@ func CmdRegisterOracle() *cobra.Command {
...
+
+			oracleCommissionMaxChangeRateStr, err := cmd.Flags().GetString(flagOracleCommMaxChangeRate)
+			if err != nil {
+				return fmt.Errorf("failed to get oralce commission max change rate")
+			}
+
+			if len(oracleCommissionMaxChangeRateStr) == 0 {
+				return fmt.Errorf("oracleCommissionMaxChangeRate is empty")
+			}
```
H4NLee, line 114: same here :)

x/oracle/client/cli/txUpdateOracleInfo.go
```diff
@@ -0,0 +1,60 @@
...
+
+			oracleCommissionRateStr, err := cmd.Flags().GetString(flagOracleCommRate)
+			if err != nil {
+				return fmt.Errorf("failed to get oralce commission rate")
+			}
+
+			if len(oracleCommissionRateStr) == 0 {
+				return fmt.Errorf("oracleCommissionRate is empty")
+			}
+
```
H4NLee, line 39:
```suggestion
```

x/oracle/client/cli/txUpdateOracleInfo.go
```diff
@@ -0,0 +1,60 @@
...
+			if err := msg.ValidateBasic(); err != nil {
+				return err
+			}
+
+			return tx.GenerateOrBroadcastTxCLI(clientCtx, cmd.Flags(), msg)
+		},
+	}
+
+	cmd.Flags().String(flagOracleEndpoint, "", "oracle's endpoint")
+	cmd.Flags().String(flagOracleCommRate, "", "oracle's commission rate")
```
H4NLee, line 55: please mark as required field as done in register-oracle

x/oracle/keeper/msg_server_oracle.go
```diff
@@ -30,7 +30,20 @@ func (m msgServer) ApproveOracleRegistration(ctx context.Context, registration *
...
+		return nil, err
+	}
+
+	ctx.EventManager().EmitEvent(
+		sdk.NewEvent(
+
+			sdk.EventTypeMessage,
+			sdk.NewAttribute(sdk.AttributeKeyModule, types.ModuleName),
+		),
+	)
```
H4NLee, line 46:
```suggestion
	ctx.EventManager().EmitEvent(
		sdk.NewEvent(
			sdk.EventTypeMessage,
			sdk.NewAttribute(sdk.AttributeKeyModule, types.ModuleName),
		),
	)
```

x/oracle/keeper/oracle.go
```diff
@@ -42,6 +43,53 @@ func (k Keeper) RegisterOracle(ctx sdk.Context, msg *types.MsgRegisterOracle) er
...
+
+	if err := k.SetOracle(ctx, oracle); err != nil {
+		return sdkerrors.Wrapf(types.ErrUpdateOracle, err.Error())
+	}
+	return nil
+}
+
+// ValidateOracleCommission validate an oracle's commission rate.
+// An error is returned if the new commission rate is invalid.
+func (k Keeper) ValidateOracleCommission(oracle *types.Oracle, blockTime time.Time, newRate sdk.Dec) error {
```
H4NLee, line 72: It seems that this function doesn't have to be a keeper function.
Instead, how about to be a function of oracle? like,

```go
func (o Oracle) ValidateCommissionRate(blockTime time.Time, newRate sdk.Dec) error {
    // validate commission rate
}
```

x/oracle/keeper/oracle.go
H4NLee: Oh, I didn't know that staking module only restrict max-change-rate when increase the commission. As @audtlr24 said, it seems good to follow the way the staking module do.

x/oracle/keeper/oracle.go
```diff
@@ -42,6 +43,53 @@ func (k Keeper) RegisterOracle(ctx sdk.Context, msg *types.MsgRegisterOracle) er
...
+	blockTime := ctx.BlockHeader().Time
+
+	if !oracle.OracleCommissionRate.Equal(msg.OracleCommissionRate) {
+		if err := k.ValidateOracleCommission(oracle, blockTime, msg.OracleCommissionRate); err != nil {
+			return sdkerrors.Wrapf(types.ErrUpdateOracle, err.Error())
+		}
+	}
+
+	oracle.UpdateTime = blockTime
+	oracle.Endpoint = msg.Endpoint
```
H4NLee, line 61: Why don't we update the endpoint only when the length of `msg.Endpoint` is greater than 0?
In other words, there is no need to enter endpoint when user want to change commission rate only.
```suggestion
        if len(msg.Endpoint) > 0 {
            oracle.Endpoint = msg.Endpoint
        }
```

H4NLee (commented):
Left some comments. Thx!

x/oracle/keeper/oracle_test.go
audtlr24: Since a commission rate of 0% is possible now, I add negative commission rate test code at [7d4ca68](https://github.com/medibloc/panacea-core/pull/540/commits/7d4ca686bba1834fc458c939a864058fa7a0fe73).

x/oracle/client/cli/txRegisterOracle.go
audtlr24: Oh, Thanks! I fixed it!

x/oracle/keeper/oracle.go
audtlr24: Thanks, there is no need for a keeper function because there is no store access. It's corrected.

x/oracle/keeper/oracle.go
audtlr24: I missed this part. thank you

x/oracle/client/cli/txUpdateOracleInfo.go
audtlr24: It has not been added as part of this section. This is because it was thought that this CLI should work even if only one of the two values is entered. 
I thought about a part that handles when there are no values, but I excluded it for now because I thought this process would become more complicated as the number of update info increases later. wdyt?

H4NLee: I got it. Then, I think `UpgradeOracleInfo` should update commission rate only when the `msg.CommissionRate` is not nil.

audtlr24: When I think about it again,  it is better to apply `MarkFlagRequired` because the commission rate returns an error when it is empty. 
And I made some modifications to the keeper function. If there is no change in the input information, an error is returned in order not to update `UpdateTime`. And, commission rate only update when the commission rate is changed.

x/oracle/types/errors.go
```diff
@@ -9,4 +9,10 @@ var (
 	ErrOracleNotFound            = sdkerrors.Register(ModuleName, 4, "oracle not found")
 	ErrInvalidUniqueID           = sdkerrors.Register(ModuleName, 5, "invalid unique id")
 	ErrApproveOracleRegistration = sdkerrors.Register(ModuleName, 6, "error while approving an oracle registration")
+	ErrUpdateOracle          = sdkerrors.Register(ModuleName, 7, "error while updating oracle info")
+
+	ErrCommissionNegative        = sdkerrors.Register(ModuleName, 8, "commission must be positive")
+	ErrCommissionGTMaxRate       = sdkerrors.Register(ModuleName, 9, "commission cannot be more than the max rate")
+	ErrCommissionUpdateTime      = sdkerrors.Register(ModuleName, 10, "commission cannot be changed more than once in 24h")
+	ErrCommissionGTMaxChangeRate = sdkerrors.Register(ModuleName, 11, "commission cannot be changed more than max change rate")
```
inchori, line 17: Please `go fmt`.

inchori (approved):
LGTM!
I left minor comments.

x/oracle/keeper/oracle.go
```diff
@@ -116,6 +119,37 @@ func (k Keeper) validateApproveOracleRegistration(ctx sdk.Context, msg *types.Ms
...
+			return sdkerrors.Wrapf(types.ErrUpdateOracle, err.Error())
+		}
+		oracle.OracleCommissionRate = msg.OracleCommissionRate
+	}
+
+	if len(msg.Endpoint) > 0 {
+		oracle.Endpoint = msg.Endpoint
+	}
+
+	oracle.UpdateTime = blockTime
```
youngjoon-lee, line 145: I think the update time should be updated only when the comm rate is changed.

audtlr24: That's right. It's fixed.

x/oracle/client/cli/txUpdateOracleInfo.go
audtlr24: Thank you for your suggestions. This way, the user seems to be able to request the oracle commission as a nil value as well. Please check [d35db47](https://github.com/medibloc/panacea-core/pull/540/commits/d35db4770e1fb267b350286643ee78ea6351cfe9).

H4NLee (approved):
Good!

gyuguen (approved):
LGTM

PR #542, opened 2022-12-08
feat: added verify uniqueID of certificate

gyuguen:
### Implement
* Added `unique_id` field in `UnsignedCertificate`
* Added verify uniqueID of certificate
* Added test msg validation of deal


x/datadeal/types/message_deal.go
```diff
@@ -82,46 +82,6 @@ func (m *MsgSubmitConsent) GetSigners() []sdk.AccAddress {
 	return []sdk.AccAddress{oracleAddress}
 }
 
-func (m *Certificate) ValidateBasic() error {
```
gyuguen, line 85: I moved it to `x/datadeal/types/certificate.go`

x/datadeal/types/message_deal_test.go
```diff
@@ -0,0 +1,166 @@
+package types
```
inchori, line 1: Does has the reason to write this test code?

inchori (approved):
lgtm

x/datadeal/types/message_deal_test.go
gyuguen: Because I wanted to do a single test for `ValidateBasic()`

audtlr24 (approved):
good

H4NLee (approved):
lgtm

PR #545, opened 2022-12-12
feat: added decrypt cli

### LINKED ISSUE #514: Add CLI: decrypt
- To decrypt something with a combined key (provided by oracles)

gyuguen:
close #514


H4NLee (approved):
LGTM
I saw the TODO, but I think it's good to keep this way. Could you explain why do you think that?

gyuguen:
> LGTM I saw the TODO, but I think it's good to keep this way. Could you explain why do you think that?

We know the oracle endpoint from the oracle store, we can provide enough encryptedCombinedKey to fetch on our own.
I think we can get oracleAddress and get encryptionCombinedKey from it.

But it just only comfort.

cmd/panacead/cmd/root.go
```diff
@@ -115,6 +115,7 @@ func initRootCmd(rootCmd *cobra.Command, encodingConfig params.EncodingConfig) {
 		AddGenesisWasmMsgCmd(app.DefaultNodeHome),
 		AddGenesisOracleCmd(app.DefaultNodeHome),
 		EncryptDataCmd(app.DefaultNodeHome),
+		DecryptData(app.DefaultNodeHome),
```
inchori, line 118: How about we change to `DecryptData` as other cmd function name?

inchori (approved):
LGTM!
I left a not important comment.

audtlr24 (approved):
lgtm

cmd/panacead/cmd/root.go
gyuguen: fixed: https://github.com/medibloc/panacea-core/pull/545/commits/b432c3840b8748e98751b9b541387e82d7aaf4c4

PR #553, opened 2022-12-14
fix: OracleCommissionRate upper limit

### LINKED ISSUE #551: Add upper limit to `OracleCommissionRate` in `MsgRegisterOracle`
`OracleCommissionRate` should not be greater than `OracleCommissionMaxRate` when oracle request registration.
Check should be added in `ValidateBasic` of `MsgRegisterOracle`.

H4NLee:
close #551

### DIFF
--- a/x/oracle/types/oracle.go
+++ b/x/oracle/types/oracle.go
@@ -86,16 +86,12 @@ func (m *OracleRegistration) ValidateBasic() error {
 		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "trusted block hash should not be nil")
 	}
 
-	if m.OracleCommissionRate.LT(sdk.ZeroDec()) || m.OracleCommissionRate.GT(sdk.OneDec()) {
-		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "oracleCommissionRate must be between 0 and 1")
-	}
-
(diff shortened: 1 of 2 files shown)


x/oracle/types/oracle.go
```diff
@@ -94,6 +94,10 @@ func (m *OracleRegistration) ValidateBasic() error {
 		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "OracleCommissionMaxRate must be between 0 and 1")
 	}
 
+	if m.OracleCommissionRate.GT(m.OracleCommissionMaxRate) {
```
gyuguen, line 97: I think we can change like below.(line: 89)

```go
if m.OracleCommissionRate.LT(sdk.ZeroDec()) || m.OracleCommissionRate.GT(m.OracleCommissionMaxRate) {
	return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "oracleCommissionRate must be between 0 and 1")
}
```

How about it?

gyuguen (approved):
LGTM
I left one comment.

x/oracle/types/oracle.go
H4NLee: Got it. It looks better. I'll change it

audtlr24 (approved):
Good, Thanks.

youngjoon-lee (approved):

PR #565, opened 2022-12-21
feat: add CLI for issuing JWT

### LINKED ISSUE #558: Add CLI: `panacead jwt ...`
For example,
```bash
panacead jwt issue --from key-1 --expiration 1h
```

You can find a code snippet for creating a JWT from https://github.com/medibloc/panacea-oracle/blob/7c2eb368e1db299e25298975f772cd1e9aa699b4/server/middleware/auth_test.go#L108

inchori:
Closes #558 

FYI)
- You can test like this:
```bash
panacead issue-jwt consumer 1h
```


cmd/panacead/cmd/jwt.go
```diff
@@ -0,0 +1,93 @@
...
+The key to be used for issuing a JWT should be stored in the localStore.
+If not stored, please add the key first.
+The [expiration] should use the format such as "1h", "30m", "2h45m" etc. 
+Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
+		`,
+		Args: cobra.ExactArgs(2),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx := client.GetClientContextFromCmd(cmd)
+
+			fromAddress := clientCtx.GetFromAddress()
```
H4NLee, line 33: When I try to use `--from` flag, `unknown flag` error occurs.
BTW, both `key-name`(arg[0]) and `--from` flag are required? IMO, we can use only `key-name` and get issuer address from it.

gyuguen: Yes, we don't need to `--from` flag.

cmd/panacead/cmd/jwt.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	signedJWT, err := jwt.Sign(token, jwt.WithKey(jwa.ES256K, privKey.ToECDSA()))
+	if err != nil {
+		return nil, err
+	}
+
+	return signedJWT, nil
```
gyuguen, line 92:
```suggestion
	return jwt.Sign(token, jwt.WithKey(jwa.ES256K, privKey.ToECDSA()))
```

gyuguen (approved):
LGTM
Please check my comment.

cmd/panacead/cmd/jwt.go
inchori: Or how about using a `--from` flag? Because when issuing a JWT, I think that the issuer should be address of key. But I couldn't find a get address from [key-name]. So how about using a `--from` flag?

H4NLee: It's up to you.

FYI, you can get address from private key as below
```go
btcecPrivKey, _ := btcec.PrivKeyFromBytes(btcec.S256(), privKeyBz)

privKey := secp256k1.PrivKey{
	Key: btcecPrivKey.Serialize(),
}

addr := sdk.AccAddress(priv.PubKey().Address()).String()
```

inchori: Hmm.. Yes I think the converting is not safe. So I'll use `--from` flag!

H4NLee: It seems if the private key is needed, we have only one option to use `key-name` argument and `UnsafeExportPrivKeyHex`.

H4NLee: I didn't know the `GetFromName()`. Looks good!

H4NLee (approved):
lgtm

audtlr24 (approved):
Thanks!

Issue #554, opened 2022-12-15
use stdout instead of `output-filepath` for CLI `encrypt/decrypt-data`


youngjoon-lee:
if other outputs don't pollute stdout already

PR #566, opened 2022-12-22
fix: Change byte array to string when decypt-data cli output logs

gyuguen:
related: #554 

When outputting the `decrypt-data` cmd we need to change the type byte array to string.


inchori (approved):
LGTM Thanks! 👍

audtlr24 (approved):
Thanks!

H4NLee (approved):

PR #573, opened 2022-12-23
fix: refunds for invalid deactivation deal

### LINKED ISSUE #572: Sending the wrong coins to the consumer address when executing a deactivation transaction.
When a consumer deactivates a deal, refunds do not proceed normally.
Because the refund amount is fixed in the deal's budget.
I will change the refund amount to the remaining amout of the deal.

gyuguen:
close #572 

We should refund the remaining amount of the deal instead of the budget.

And I will add test code. 
But it need to refactorying `deal_test.go`, `certificate_test.go`


audtlr24 (approved):

H4NLee (approved):
lgtm

inchori (approved):
lgtm Thanks!

PR #578, opened 2022-12-27
feat: add oracle upgrade endblocker

### LINKED ISSUE #570: Implement oracle upgrade handler at endblocker
- oracle upgrade handler check block height at `endblocker` when oracle upgrade proposal is passed
- update oracle params and `Oracles` when block height reaches upgrade height

audtlr24:
closes #570 .
- TODO: add a part to update `Oracle` in `ApplyOracle`


H4NLee (approved):
lgtm

inchori (approved):
LGTM

x/oracle/abci.go
```diff
@@ -0,0 +1,31 @@
...
+		} else {
+			panic(err)
+		}
+	}
+
+	if upgradeInfo.ShouldExecute(ctx) {
+		if err := keeper.ApplyUpgrade(ctx, upgradeInfo); err != nil {
+			panic(err)
+		}
+		keeper.RemoveOracleUpgradeInfo(ctx)
```
H4NLee, line 29: How about not deleting the upgrade info?
When a new upgrade proposal passes, it will be overwritten anyway.

The reason I'm talking about this is because of the following case:

If the oracles didn't upgrade until the target height, they had to re-request `register-oracle` transaction again to be shared the oracle private key in the new binary(upgrade version). 
However, at this time, the oracles will be able to enter the commission rate, which overwrites the existing oracle commission rate. 
In other words, this case should also be done through `upgrade-oracle`, and to do so, it seems that upgrade info should be left. upgrade info will mean upcoming or recent upgrade information.

--
FYI) Plus, in `register-oracle`, the check that if the oracle has already been set or not should be added.

gyuguen: That's a good idea.
Then, how about `OracleKeyMigration` rather than a name `OracleUpgrade`?
OracleKeyMigration works only when it has the same value as uniqueID of `OracleUpgradeInfo`.

gyuguen: Oh, please forget the name change. I'll think about it more.

audtlr24: Okay, I understand the problem. 
If the oracles didn't upgrade until the target height, oracle operator have to use `upgrade-oracle` because the oracle is already in `Oracle` store.
For now, I think @H4NLee 's suggestion is the best way.

gyuguen: If you change this flow, please left comment :)

audtlr24: I fixed it in [8e1a3d8](https://github.com/medibloc/panacea-core/pull/578/commits/8e1a3d8338c0d7fdc61e0a27556102f708c5fe00).

x/oracle/types/oracle.go
```diff
@@ -121,3 +121,10 @@ func (m *Oracle) ValidateOracleCommission(blockTime time.Time, newRate sdk.Dec)
 
 	return nil
 }
+
+func (m *OracleUpgradeInfo) ShouldExecute(ctx sdk.Context) bool {
+	if m.Height > 0 {
+		return m.Height <= ctx.BlockHeight()
```
gyuguen, line 127: we decision not deleted `uniqueID` for `OracleUpgradeInfo`. 
So I think you need to change condition like below:
```suggestion
		return m.Height == ctx.BlockHeight()
```

gyuguen (approved):
LGTM
Please check my review.

x/oracle/types/oracle.go
H4NLee: really good point!

audtlr24: good!

youngjoon-lee (commented):
lgtm

PR #579, opened 2022-12-28
feat: add upgrade oracle tx & cli

audtlr24:
closes #571.

PR #582, opened 2022-12-29
feat: implement approve oracle upgrade tx

### LINKED ISSUE #574: Implement approve oracle upgrade tx
After verifying signature with oracle public key, set `EncryptedOraclePrivKey` for oracle upgrade.

H4NLee:
close #574 

Implemented `MsgApproveOracleUpgrade` transaction with some refactoring.


x/oracle/keeper/oracle.go
```diff
@@ -13,24 +13,24 @@ func (k Keeper) RegisterOracle(ctx sdk.Context, msg *types.MsgRegisterOracle) er
...
+	if params.UniqueId != msg.UniqueId {
+		return sdkerrors.Wrapf(types.ErrRegisterOracle, types.ErrInvalidUniqueID.Error())
 	}
 
-	if oracle, err := k.GetOracle(ctx, oracleRegistration.OracleAddress); !errors.Is(types.ErrOracleNotFound, err) {
-		if oracle != nil {
-			return sdkerrors.Wrapf(types.ErrOracleRegistration, "already registered oracle. address(%s)", oracleRegistration.OracleAddress)
-		} else {
-			return sdkerrors.Wrapf(types.ErrOracleRegistration, err.Error())
-		}
```
H4NLee, line 29: Modified to return an error if the `OracleRegistration` has been stored already regardless of whether Oracle has been registered.
That is, re-requests to be shared the oracle private key again are not considered.
The reason for this is that exploits of re-requests are expected to be possible. 

For example, an oracle may be free to set a new commission rate overriding the commission max rate and commission max change rate initially set via a `RegisterOracle` re-request.
I can't think of a proper way to prevent this, so I'd like to limit the re-request feature.

Since there is no way for oracles to be re-shared the oracle private key, it is necessary to safely manage the shared oracle private key or the node private key that can decrypt the `EncryptedOraclePrivKey` stored on the chain.
If all keys are lost, oracle will have to request a new registration with a new oracle address.

x/oracle/keeper/oracle.go
```diff
@@ -40,20 +40,27 @@ func (k Keeper) RegisterOracle(ctx sdk.Context, msg *types.MsgRegisterOracle) er
...
 	oracleRegistration, err := k.GetOracleRegistration(ctx, msg.ApprovalSharingOracleKey.UniqueId, msg.ApprovalSharingOracleKey.TargetOracleAddress)
 	if err != nil {
 		return sdkerrors.Wrapf(types.ErrApproveOracleRegistration, err.Error())
 	}
 
+	// if EncryptedOraclePrivKey is already set, return error
+	if oracleRegistration.EncryptedOraclePrivKey != nil {
+		return sdkerrors.Wrapf(types.ErrApproveOracleRegistration, "already approved oracle registration. if you want to be shared oracle private key again, please register oracle again")
+	}
+
```
H4NLee, line 63: Instead of checking that if oracle is set by `k.HasOracle()` which was in `validateApproveOracleRegistration`, it returns error if the `oracleRegistration.EncryptedOraclePrivKey` is not nil.
That means someone approved the oracle registration, so that the `EncryptedOraclePrivKey` is not nil.

x/oracle/keeper/oracle.go
```diff
@@ -88,33 +95,28 @@ func (k Keeper) ApproveOracleRegistration(ctx sdk.Context, msg *types.MsgApprove
 
 }
 
-// validateApproveOracleRegistration checks signature
-func (k Keeper) validateApproveOracleRegistration(ctx sdk.Context, msg *types.MsgApproveOracleRegistration) error {
-
+// validateApprovalSharingOracleKey validate unique ID of ApprovalSharingOracleKey and its signature
+func (k Keeper) validateApprovalSharingOracleKey(ctx sdk.Context, approval *types.ApprovalSharingOracleKey, signature []byte) error {
```
H4NLee, line 99: This validation function is used in both `ApproveOracleRegistration` and `ApproveOracleUpgrade`.

x/oracle/keeper/oracle.go
```diff
@@ -12,25 +12,21 @@ import (
 func (k Keeper) RegisterOracle(ctx sdk.Context, msg *types.MsgRegisterOracle) error {
 	oracleRegistration := types.NewOracleRegistration(msg)
 
-	if err := oracleRegistration.ValidateBasic(); err != nil {
-		return err
-	}
-
```
H4NLee, line 18: As [commented in the previous PR](https://github.com/medibloc/panacea-core/pull/579#discussion_r1059219257), I deleted this validation.

x/oracle/keeper/oracle_test.go
```diff
@@ -133,10 +132,11 @@ func (suite *oracleTestSuite) TestRegisterOracleSuccess() {
 	suite.Require().Equal(suite.oracleCommissionMaxChangeRate, oracleFromKeeper.OracleCommissionMaxChangeRate)
 }
 
-func (suite *oracleTestSuite) TestRegisterOracleFailedValidateToMsgOracleRegistration() {
```
H4NLee, line 136: Accordingly [this](https://github.com/medibloc/panacea-core/pull/582/files#r1059222830), deleted validation test code.

x/oracle/keeper/oracle.go
audtlr24: understand. In the current situation, if `Oracle` loses the key, the only way is to change the address and request a new registration.

x/oracle/keeper/oracle.go
```diff
@@ -88,33 +91,28 @@ func (k Keeper) ApproveOracleRegistration(ctx sdk.Context, msg *types.MsgApprove
...
-	if err := k.VerifyOracleSignature(ctx, msg.ApprovalSharingOracleKey, msg.Signature); err != nil {
-		return err
-	}
-
-	if msg.ApprovalSharingOracleKey.EncryptedOraclePrivKey == nil {
-		return fmt.Errorf("encrypted oracle private key is nil")
+	// check if the approver oracle exists
+	existing, err := k.HasOracle(ctx, approval.ApproverOracleAddress)
+	if err != nil || !existing {
+		return fmt.Errorf("failed to check if the approver oracle exists or not. address(%s)", approval.ApproverOracleAddress)
```
audtlr24, line 106: How about splitting the error? It seems that the address may not be valid and the oracle may not be in the store.

audtlr24 (approved):
Thanks for your hard work. I have one comment.

x/oracle/keeper/oracle.go
H4NLee: > In the current situation, if Oracle loses the key, the only way is to change the address and request a new registration.

Exactly.

x/oracle/keeper/oracle.go
H4NLee: Got it. [1c69a6d](https://github.com/medibloc/panacea-core/pull/582/commits/1c69a6de0ebc97888b33d92718875dc523c84cac)

As you can see the commit, I changed the `HasOracle` function.
Because we should check if the oracle exists and is active as well. Thank you for pointing out.

x/oracle/keeper/upgrade.go
```diff
@@ -129,3 +129,47 @@ func (k Keeper) SetOracleUpgrade(ctx sdk.Context, upgrade *types.OracleUpgrade)
...
+	// emit event
+	ctx.EventManager().EmitEvent(
+		sdk.NewEvent(
+			types.EventTypeApproveOracleUpgrade,
+			sdk.NewAttribute(types.AttributeKeyOracleAddress, msg.GetApprovalSharingOracleKey().GetTargetOracleAddress()),
+			sdk.NewAttribute(types.AttributeKeyUniqueID, upgradeInfo.GetUniqueId()),
+		),
+	)
+
+	// TODO: add to queue(?) for update unique ID
```
gyuguen, line 172: As discussed offline, it seems that you can check if the encryptedOraclePrivKey of OracleUpgrade is not nil without putting it in the queue.

gyuguen (approved):
LGTM

x/oracle/keeper/upgrade.go
H4NLee: @audtlr24, we probably don't need to use queue to update unique ID for upgraded oracles.
As @gyuguen said, we can know that which oracles are upgraded through the `EncryptedOralePrivKey` in `OracleUpgrade`. It makes simpler I think.

audtlr24: Okay, I understand. I will reflect it in PR #584.

inchori (approved):
lgtm

youngjoon-lee (commented):
lgtm

PR #583, opened 2022-12-29
feat: add oracle upgrade related genesis

### LINKED ISSUE #581: Implement OracleUpgrade and OracleUpgradeInfo to genesis

audtlr24:
closes #581 .


H4NLee (approved):
lgtm

x/oracle/types/genesis.go
```diff
@@ -41,6 +43,18 @@ func (m *GenesisState) Validate() error {
 		}
 	}
 
+	for _, oracleUpgrade := range m.OracleUpgrades {
+		if err := oracleUpgrade.ValidateBasic(); err != nil {
+			return err
+		}
+	}
+
+	if m.OracleUpgradeInfo != nil {
```
inchori, line 52: Does the `OracleUpgradeInfo` in `GenesisState` should be always not nil?

inchori (approved):
Gooood!
Just one question.

x/oracle/types/genesis.go
inchori: Ah.. I understand it. Please forgot this comment.

gyuguen: If we follow the specs in this [link](https://github.com/medibloc/panacea-core/pull/578#discussion_r1059859889), I think we should check it like @inchori said.

gyuguen (approved):
LGTM

x/oracle/types/genesis.go
H4NLee: right!

H4NLee: The `OracleUpgradeInfo` can be nil if there is no upgrade.
It seems no problem.

gyuguen: Aha, I was confused about how far the upgrade action could go.
If the upgrade command cannot be performed as long as the upgrade action does not occur, now is OK.
However, even if there is no upgrade operation, if the upgrade command is available (i.e. OracleShareKey is lost), we must set uniqueID in params and upgradeInfo when launching DEP.

As @H4NLee said before, if oracle can't use this address when the key is lost, no problem.

PR #584, opened 2022-12-30
feat: improve apply oracle using OracleUpgradeQueue

audtlr24:
Implement `OracleUpgradeQueue` to have a list of `Oracles` that have been upgraded.
In `ApplyUpgrade`, as it iterate through the queue, it replace the uniqueID of the oracle store.


x/oracle/types/keys.go
```diff
@@ -45,6 +46,18 @@ func GetOracleUpgradeKey(uniqueID string, address sdk.AccAddress) []byte {
 	return append(OracleUpgradeKey, CombineKeys([]byte(uniqueID), address)...)
 }
 
+func GetOracleUpgradeQueueKey(uniqueID string, addr sdk.AccAddress) []byte {
+	return append(OracleUpgradeQueueKey, CombineKeys([]byte(uniqueID), addr)...)
+}
```
H4NLee, line 51: Since we can have only one `OracleUpgradeInfo`, how about using the key with combination of `OracleUpgradeQueueKey` and `addr` for oracle upgrade queue?

x/oracle/types/keys.go
```diff
@@ -45,6 +46,18 @@ func GetOracleUpgradeKey(uniqueID string, address sdk.AccAddress) []byte {
...
+
+func GetOracleUpgradesKey(uniqueID string) []byte {
+	return append(OracleUpgradeQueueKey, []byte(uniqueID)...)
+}
```
H4NLee, line 55: The name of ₩GetOracleUpgradesKey₩ seems to be the key to get `OracleUpgrade`. I suggest to rename it such as `GetOracleUpgradeQueueKeyByUniqueID`.
Or, you can use the `types.OracleUpgradeQueueKey` itself for iterator if you do not use the unique ID for this queue.

x/oracle/keeper/upgrade.go
```diff
@@ -53,7 +53,22 @@ func (k Keeper) ApplyUpgrade(ctx sdk.Context, info *types.OracleUpgradeInfo) err
 	}
 	k.SetParams(ctx, params)
 
-	//TODO: update `Oracles` that already upgraded
+	iterator := k.GetOracleUpgradeQueueIterator(ctx, info.UniqueId)
+	defer iterator.Close()
+
+	for ; iterator.Valid(); iterator.Next() {
+		uniqueID, accAddr := types.SplitOracleUpgradeQueueKey(iterator.Key())
```
H4NLee, line 60: It seems that we don't have to split the key.
Just use the value for address and use unique ID from upgrade info.

H4NLee (commented):
And we need `InitGenesis` and `ExportGenesis` for this queue :)

x/oracle/types/keys.go
audtlr24: That's a good point. Since the uniqueID of all elements int the queue will be the same , it seems better to remove it.

x/oracle/types/keys.go
audtlr24: Thanks, I use `types.OracleUpgradeQueueKey` itself for iterator.

H4NLee:
I think we should clear(remove all) the queue in `SetUpgradeInfo` which is called when the proposal for upgrade is passed.

Let's assume that if a proposal for upgrade is passed, some oracles are added to the queue by upgrading accordingly. However, some critical bugs are found before the oracle upgrade target height. Thus, a new patch for fix is applied, and a new oracle upgrade proposal is passed. For this upgrade, the previous queue will have to be cleared.

audtlr24:
> I think we should clear(remove all) the queue in `SetUpgradeInfo` which is called when the proposal for upgrade is passed.
> 
> Let's assume that if a proposal for upgrade is passed, some oracles are added to the queue by upgrading accordingly. However, some critical bugs are found before the oracle upgrade target height. Thus, a new patch for fix is applied, and a new oracle upgrade proposal is passed. For this upgrade, the previous queue will have to be cleared.

I understand your comment. Then, do we go with the strategy of overwriting the existing `UpgradeInfo` and removing oracles from queue that performed `OracleUpgradeTx` when a new oracle upgrade proposal passes?

H4NLee:
> I understand your comment. Then, do we go with the strategy of overwriting the existing UpgradeInfo and removing oracles from queue that performed OracleUpgradeTx when a new oracle upgrade proposal passes?

Yes I think so.

gyuguen:
I think we don't need a queue.
We can determine wheter or not a change has been made to the encryptedOraclePrivKey field in the `OracleUpgrade` store

audtlr24:
I open a PR #586 that use checking `EncryptedOraclePrivKey` instead of using queue. So, I closed this PR.

PR #586, opened 2023-01-03
feat: add set oracles at ApplyUpgrade

audtlr24:
As discussed in https://github.com/medibloc/panacea-core/pull/582#discussion_r1060229772, instead of applying the queue as in PR #584, change it to check the `EncryptedOraclePrivKey` in the `OracleUpgrade` store.
Existing PR #584 will be closed.


x/oracle/keeper/upgrade.go
```diff
@@ -39,21 +39,35 @@ func (k Keeper) GetOracleUpgradeInfo(ctx sdk.Context) (*types.OracleUpgradeInfo,
...
-	//TODO: update `Oracles` that already upgraded
+	iterator := k.GetOracleUpgradeIterator(ctx, info.UniqueId)
+	defer iterator.Close()
+
+	for ; iterator.Valid(); iterator.Next() {
+		uniqueID, address := types.SplitOracleUpgradeKey(iterator.Key())
+		oracleUpgrade, err := k.GetOracleUpgrade(ctx, uniqueID, address.String())
+		if err != nil {
+			return err
+		}
```
gyuguen, line 59: I think you can change like below:
We don't have to called GetOracle(...) function.
```suggestion
		bz := iterator.Value()
		oracleUpgrade := &types.OracleUpgrade{}

		if err := k.cdc.UnmarshalLengthPrefixed(bz, oracleUpgrade); err != nil {
			return err
		}
```

gyuguen (approved):
Thank you.
I left a comment.

x/oracle/keeper/upgrade.go
audtlr24: I fixed it in [0e5ad56](https://github.com/medibloc/panacea-core/pull/586/commits/0e5ad56b39f93bb576e7249e4d4d0eb114687c53). Please check again, thanks.

H4NLee (approved):
lgtm.

inchori (approved):
LGTM

PR #567, opened 2022-12-22
docs: add DEP docs

youngjoon-lee:

PR #599, opened 2023-01-05
docs: add confidential oracle docs

inchori:
Based on https://github.com/medibloc/panacea-core/pull/567, added `confidential-oracle`.


.gitbook/3-protocol-devs/1-dep-specs/5-confidential-oracle.md
```diff
@@ -0,0 +1,50 @@
...
+When running an oracle with the genuine binary, the data encryption and decryption are performed by oracle private key which is sealed by unique ID located in Intel SGX.
+So if the binary is changed or corrupted, the data encryption and decryption would not work.
+
+So how to we set a correct unique ID and how to know it is not malicious binary? The answer is that the unique ID will be determined by Panacea governance.
+We can know the correct unique ID from Panacea as Single Source of Truth(SSOT), and know that what correct genuine binary is.
+Because of the reason, the oracle operator who wants to steal data could not use malicious binary and it could be prevented from potential data leakage.
+
+
+## Implementation of Confidential Oracle
+
```
inchori, line 41: I think this part is need to be complemented.

.gitbook/3-protocol-devs/1-dep-specs/5-confidential-oracle.md
```diff
@@ -0,0 +1,50 @@
...
+In DEP, anyone can operate oracle nodes but must use confidential computing.
+If the operator runs the confidential oracle, the only oracle software can decrypt a data and validate it. It means that the oracle operator can't decrypt a data and it will be not exposed to the operator.
+Also, when running the confidential oracle, the only genuine binary built in Intel SGX should be used so that it can be prevented from the node who uses malicious binary. It means that all oracles run the same binary.
+Using a genuine binary can preserve a privacy of data.
+
+### Additional Effort
+
+Using a confidential oracle could reinforce a security, but we developed an additional effort for protecting a data leak.
+When running an oracle with the genuine binary, the data encryption and decryption are performed by oracle private key which is sealed by unique ID located in Intel SGX.
+So if the binary is changed or corrupted, the data encryption and decryption would not work.
```
inchori, line 33: I'm not sure the word `corrupted` is right.

youngjoon-lee: `tampered` would be good.

.gitbook/3-protocol-devs/1-dep-specs/5-confidential-oracle.md
```diff
@@ -0,0 +1,50 @@
+# Confidential Oracle
```
youngjoon-lee, line 1: Your documentation is basically good, but I think you can organize contents much well. I saw many duplicated contents across many sections. I'd suggest the entire plot cleanly like below. You don't have to follow the plot I suggest, but I hope you can see the way how to design the storyline more cleanly.

[Plot Suggestion]

> In DEP, data must be verified as explain in the Data Validation specification.
> -> This cannot be done on chain since data is sensitive. That's why we need the off-chain oracle.
> -> Verifying data means that the oracle should be able to see the contents of data. It also means that oracle operators (human) can see the content of sensitive data. 
> -> _(Here, you can put some examples of malicious behavior. For example, if the oracle software stores something important to the disk without sealing it, it can be read by anyone who can access the disk. Or, if the oracle operator can modify the source code of the oracle to print logs of data contents or to send contents to somewhere else, it would cause serious privacy problems.)_
> -> To prevent it, we have designed the DEP oracle with confidential computing.
> -> Confidential computing guarantees that only selected code (binary) can access the data within the secure enclave. Using it, we can protect sensitive data from being leaked by malicious oracle operators while verifying data correctly in decentralized environments.


Then, you can describe the implementation strategy of confidential oracle. As you've done, you can say that we've used SGX and EGo. 

After that, you can describe how we designed the oracle specifically using SGX and EGo. You've already covered it in the `Additional Efforts` section, introducing unique ID. It is good but you can describe more details.
First, you can explain what an oracle key is, and what if the operator modify the oracle source code, in conjunction with the oracle key.
Then, you can introduce what `MsgRegisterOracle` is, by explaining what the remote report and oracle key mean.
Also, you can introduce how the oracle registration message is handled by other oracles. You can explain what the existing oracles validate.

inchori: Okay, I think that the plot you suggested seems to be good. I'll reorganize the docs and re-request soon.

inchori:
I reflected quite a lot of this document as @youngjoon-lee commented. If there have some points to be fixed or be complemented, please comment again.

.gitbook/3-protocol-devs/1-dep-specs/5-confidential-oracle.md
```diff
@@ -0,0 +1,57 @@
...
+It means that the all oracle operators must use the only genuine binary built by selected code. The only genuine binary
+can unseal the data in the SGX secure enclave.
+
+If the oracle wants to register, it must use remote report composed of the promised security version and unique ID.
+So the one of the oracle approves the registration it the new oracle's remote report is valid.
+If you want a register the oracle, you can verify remote report using below CLI before submitting a registration.
+
+```bash
+ego run oracled verify-report [report-file-path]
+```
```
youngjoon-lee, line 51: I don't think it's good to put this CLI guide here.

.gitbook/3-protocol-devs/1-dep-specs/5-confidential-oracle.md
```diff
@@ -0,0 +1,57 @@
...
+It also means that the oracle operators(human) can see the content of sensitive data. For example, if the oracle
+software stores the important sensitive data to the disk or storage without sealing it,
+it can be read by anyone who can access the disk or storage. Plus, if the malicious oracle operator modified the source
+code of the oracle to print logs of sensitive data or to leak a contents somewhere else,
+it would cause the serious privacy problems.
+To prevent it, we have designed the DEP oracle with confidential computing. Confidential computing guarantees that only
+selected code (binary) can access data in the secure enclave. Also it could protect a sensitive data from being
+leaked by malicious operators while verifying a data correctly in decentralized system.
+
+## Confidential Oracle
```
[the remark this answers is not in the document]
youngjoon-lee: You don't need to put everything into one or two paragraphs, as the examples I wrote above. You can define subsub-chapters if you want.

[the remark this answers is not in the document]
youngjoon-lee: Also you don't need to use all contents that I wrote :) You can just put some links to the oracle guide docs for details if you want.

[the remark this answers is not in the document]
inchori: I got it. Thanks. I think that **_Oracle Key_** and **_Oracle Key Handshake_** is especially need so I'll make sub-chapters of them.

.gitbook/3-protocol-devs/1-dep-specs/5-confidential-oracle.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Synopsis
+
+This document explains that why the oracles must use confidential computing.
+
+## Motivation
+
+In DEP, the data must be verified as explained in the [data validation documentation](./4-data-validation.md).
+Since the data is sensitive, we designed the off-chain oracle. Verifying data means that the content of data could be
+shown to the oracle.
```
H4NLee, line 21: In my opinion, off-chain oracles are not necessary because the data is sensitive.
The reason why we need confidential oracle is that even oracle has no right to see the data.

```suggestion
Verifying data means that the content of data could be shown to the oracle. However, no one would want their data to be exposed to the oracle for data verification. There may also be malicious oracles to exploit the data unfairly.
```

.gitbook/3-protocol-devs/1-dep-specs/5-confidential-oracle.md
```diff
@@ -0,0 +1,57 @@
...
+It also means that the oracle operators(human) can see the content of sensitive data. For example, if the oracle
+software stores the important sensitive data to the disk or storage without sealing it,
+it can be read by anyone who can access the disk or storage. Plus, if the malicious oracle operator modified the source
+code of the oracle to print logs of sensitive data or to leak a contents somewhere else,
+it would cause the serious privacy problems.
```
H4NLee, line 26: IMO, it seems too detailed in motivation.
I think just need to write why confidential computing is used here.
We can move this details into the main contents.
```suggestion
```

.gitbook/3-protocol-devs/1-dep-specs/5-confidential-oracle.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Confidential Oracle
+
+For running the DEP oracle with confidential computing, we choose an Intel SGX which offers hardware-based memory
+encryption for data security
+using [Microsoft Azure Confidential Computing VM](https://learn.microsoft.com/en-us/azure/confidential-computing/overview).
+Also, we choose an [EGo](https://www.edgeless.systems/products/ego/) SDK which enables to develop confidential apps
+written in Go language.
+
+The data encryption and decryption are performed by the oracle private key. The oracle private key is sealed by unique
```
H4NLee, line 39: I felt gaps between this two sentence. IMO, we need to introduce what the oracle private key is, first.

.gitbook/3-protocol-devs/1-dep-specs/5-confidential-oracle.md
H4NLee: agreed.

.gitbook/3-protocol-devs/1-dep-specs/5-confidential-oracle.md
inchori: I applied your comments in [fea8f3c](https://github.com/medibloc/panacea-core/pull/599/commits/fea8f3cc6e5755ab04ccf1999e19f75fddc4a4dc).

.gitbook/3-protocol-devs/1-dep-specs/5-confidential-oracle.md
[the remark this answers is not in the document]
inchori: [fea8f3c](https://github.com/medibloc/panacea-core/pull/599/commits/fea8f3cc6e5755ab04ccf1999e19f75fddc4a4dc). And I just use the **_Sealing Secrets and States_** part.

.gitbook/3-protocol-devs/1-dep-specs/5-confidential-oracle.md
```diff
@@ -0,0 +1,68 @@
...
+Verifying data means that the content of data could be shown to the oracle. However, no one would want their data to be
+exposed to the oracle for data verification. There may also be malicious oracles to exploit the data unfairly.
+
+## Confidential Oracle
+
+The oracle operators(human) can see the content of sensitive data. For example, if the oracle software stores the
+important sensitive data to the disk or storage without sealing it, it can be read by anyone who can access the disk or
+storage. To prevent it, we have designed the DEP oracle with confidential computing. Confidential computing guarantees
+that only selected code (binary) can access data in the secure enclave. Also it could protect a sensitive data from
+being leaked by malicious operators while verifying a data correctly in decentralized system.
```
H4NLee, line 29: I think that it would be better to move the part of from `To prevent it, ...` to the end inside motivation.
It seems to me that there is only a problem statement in motivation.

youngjoon-lee (approved):
Good. Danke, even though I've left so many suggestions for the plot.
This doc is one of the most difficult to write, and it's so hard to write everything at once now. We'll improve this doc gradually. So, this initial version is good enough.

audtlr24 (approved):
Thanks for your hard work. lgtm!

H4NLee (approved):
Thank you for your hard work.

gyuguen (approved):
Thank you for your hard work.
This article covered a very difficult topic.

.gitbook/3-protocol-devs/1-dep-specs/5-confidential-oracle.md
inchori: OK, I got it. But if I moved the part to inside motivation. I think some more content need to be added in **_Confidential Oracle_** section.

PR #601, opened 2023-01-06
docs: incentives for protocol devs

H4NLee:
For now, our incentives (or reward distribution) is quite simple, so I just write down simply.
If you think some more contents are needed in this doc, please leave comments.

FYI, I intentionally did not write down the contents related to the oracle commission rate(plus max rate and max change rate). Because it seems more related to oracle.

### DIFF
--- a/.gitbook/3-protocol-devs/1-dep-specs/6-incentives.md
+++ b/.gitbook/3-protocol-devs/1-dep-specs/6-incentives.md
@@ -13,7 +13,9 @@
 
 ## Synopsis
 
-When data providers [submit consent to provide their data](./3-data-provider-consent.md), the budget deposited by the data consumer is distributed to the provider and the oracle that verified the data.
+When data deal is created, a budget for data is deposited from data consumer account. 
+And the deposit is distributed to the data providers and the oracles that verified the data.
+Data consumers also can [deactivate the deal](./2-data-deal.md#Deactivate-Data-Deal) they created and refund the remaining budget at any time if the deal is in status `DEAL_STATUS_ACTIVE`.
 
 ### Motivation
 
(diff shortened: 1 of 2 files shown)


audtlr24 (approved):
lgtm. For the consumer's veiw, how about writing down the details of depositing funds when making a deal and refunding them when deactivating? 

I'm not sure if this would be better put into `2-data-deal.md`.

H4NLee:
> lgtm. For the consumer's veiw, how about writing down the details of depositing funds when making a deal and refunding them when deactivating?
I'm not sure if this would be better put into 2-data-deal.md.

It is written in `2-data-deal.md` briefly, but I agree that it would be great to add as you said.
It can provide the whole token flow of DEP. I'll add it.

.gitbook/3-protocol-devs/1-dep-specs/6-incentives.md
```diff
@@ -0,0 +1,66 @@
...
+## Technical Specification
+
+### Reward Distribution
+
+When [creating a data deal](./2-data-deal.md#create-data-deal), data consumers specify the quantity and the budget for the data.
+From this, the price per data can be calculated as:
+
+```
+price_per_data = deposit / max_num_data
+```
```
inchori, line 35: I think that it will be good for explaining how does the remainder calculated.

inchori (approved):
lgtm

gyuguen (approved):
LGTM

.gitbook/3-protocol-devs/1-dep-specs/6-incentives.md
H4NLee: Actually, we don't calculate remaining budget.
When consumer deactivate the deal, just balance goes back to consumer (by using `bankkeeper.GetBalance`).

youngjoon-lee (approved):
Good enough!

audtlr24 (approved):
lgtm

inchori (approved):
lgtm

PR #604, opened 2023-01-10
docs: v2.0.x: update multisig doc

youngjoon-lee:
Since we updated the cosmos sdk dependency, the CLI usage for multisig has been changed. Please see details from diff.


H4NLee (approved):

gyuguen (approved):
LGTM

PR #607, opened 2023-01-10
docs: add oracle installation

audtlr24:
`How about building from source?` part was written simply. If there is anything to add, please comment.

### DIFF
--- a/.gitbook/5-oracles/1-operate-oracle-nodes/1-oracle-installation.md
+++ b/.gitbook/5-oracles/1-operate-oracle-nodes/1-oracle-installation.md
@@ -1,0 +1,34 @@
+# Installation
+
+## Hardware Requirement
+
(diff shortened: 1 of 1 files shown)


.gitbook/5-oracles/1-operate-oracle-nodes/1-oracle-installation.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+### How about building from source?
+
+You can build from source by referring to the following [installation-from-source](https://github.com/medibloc/panacea-oracle/blob/main/docs/installation-src.md).
+
+However, we highly recommend installing using docker. 
+This is because the uniqueID in `EGo` is sensitively changes depending on the Go dependency or local debug environment.
+
+Therefore, it is recommended that you check the uniqueID with following CLI when you want to use the binary you built yourself.
```
H4NLee, line 36: It would be great to add comment that `scripts/enclave-prod.json` is used in docker. Because if it is signed using different `enclave.json`, it will result in different uniqueID

H4NLee (approved):
LGTM

gyuguen (approved):
LGTM

inchori (approved):
LGTM

youngjoon-lee (approved):
lgtm

PR #608, opened 2023-01-10
feat: add Oracle update at ApproveOracleUpgrade

### LINKED ISSUE #605: Improve upgrade oracle tx processing after upgrade target height
In Panacea, `upgrade-oracle` tx is processed after reaching the upgrade height, but this is not reflected in the `Oracle` store. Because the part that set to the `Oracle` store only works in `BeginBlocker`.

Therefore, we need to modify the function so that this part can be applied. For example, in `ApproveOracleUpgrade`, we can check the `uniqueID` of `Params` and call `SetOracle`.

audtlr24:
closes #605 .

Add `Oracle` store updates when an oracle upgrade request comes after upgrade height.


H4NLee (approved):
LGTM

x/oracle/keeper/upgrade.go
```diff
@@ -208,6 +208,18 @@ func (k Keeper) ApproveOracleUpgrade(ctx sdk.Context, msg *types.MsgApproveOracl
 		return sdkerrors.Wrapf(types.ErrApproveOracleUpgrade, err.Error())
 	}
 
+	// set oracle for request after upgrade height
+	if ctx.BlockHeight() >= upgradeInfo.Height {
```
H4NLee, line 212: FYI)
It means that target height of upgrade has passed. Thus, update the unique ID immediately.

youngjoon-lee: Well then.. Is this PR for covering a case when the OracleUpgrade proposal is passed without any ApproveOracleUpgrade tx because other existing oracles were busy, for example.

youngjoon-lee: Hm. I think I'm not understanding this issue correctly. Please elaborate the point of this issue well.

H4NLee: Yes. This is for the case that an oracle try to upgrade after target height.

This is out of scope of this PR but.. based on your example, oracle should be able to retry to request their upgrade. Because all existing oracles can be busy.
Plus, retry for oracle registration should also be allowed if the `EncryptedOraclePrivKey` is nil.

youngjoon-lee: Ah. I got it. Now, I understood that this PR is not related to the oracle-upgrade-proposal, and is related only with the oracle-upgrade & approve-oracle-upgrade transactions. Then, this PR makes sense.

It would be better to update the comment in the source code above like:
```
approve the oracle-upgrade request immediately if the upgrade height has been already passed
(if the oracle-upgrade request wasn't processed by the BeginBlocker at the upgrade height, because there was no approval at that time.)
```

youngjoon-lee (approved):
lgtm. pls see the suggestion https://github.com/medibloc/panacea-core/pull/608#discussion_r1066514202

PR #619, opened 2023-01-12
docs: oracle upgrade for oracle operators

H4NLee:


youngjoon-lee (approved):
lgtm

audtlr24 (approved):
lgtm

.gitbook/5-oracles/1-operate-oracle-nodes/7-oracle-upgrade.md
```diff
@@ -0,0 +1,112 @@
...
+panacead q oracle oracle-upgrade-info
+```
+
+### Upgrade Oracle Node
+
+{% hint style="info" %}
+You can upgrade oracle any time after an `oracle-upgrade` proposal has passed (even before the upgrade target height is reached).
+However, since the new version of oracle can be active after the target height, the current version of oracle must be run before the target height.
+
+You do not have to stop the old version of oracle when upgrading to a new version of oracle.
```
inchori, line 45: Does we need to guide change the `listen-addr`?

inchori (approved):
LGTM Thanks!

.gitbook/5-oracles/1-operate-oracle-nodes/7-oracle-upgrade.md
H4NLee: I think we don't have to. If the same address is used, oracle daemon returns error and exit.
And I expect that they shut down the old version of oracle and start the new one after upgrade.

PR #626, opened 2023-02-02
feat: change did module to use areis framework

### LINKED ISSUE #625: Change x/did module to use aries framework
## Background
The current `x/did` module implements proto by referring to the [w3c spec](https://www.w3.org/TR/did-core/).
However, in the current state, it is difficult to continuously apply updates to the spec, so we change to use [aries-framework-go](https://github.com/hyperledger/aries-framework-go), an open source that is being developed in accordance with the w3c spec.

## Implementation
- change `x/did` proto files that can store `aries-framework-go` did document.
- change related cli and keeper functions using `aries-framework-go` internal functions.

audtlr24:
closes #625.

- change the did `document` to be stored in byte array format. Documents stored in the form of byte array can be internally converted into [Doc](https://github.com/hyperledger/aries-framework-go/blob/b4e5124106cb4f71e9fb659918209f65535a2fbb/pkg/doc/did/doc.go#L270) instance defined in `aries-framework-go` through the `ParseDocument()` function.
- add `document_data_type` field to specify which version of `aries-framework-go` the `document` refers to.
- remove most of the existing validation functions, since the [ParseDocument](https://github.com/hyperledger/aries-framework-go/blob/b4e5124106cb4f71e9fb659918209f65535a2fbb/pkg/doc/did/doc.go#L478) `aries-framework-go` validates documents


x/did/keeper/msg_server_did.go
```diff
@@ -80,23 +79,29 @@ func (m msgServer) DeactivateDID(goCtx context.Context, msg *types.MsgDeactivate
...
 		return 0, sdkerrors.Wrapf(types.ErrVerificationMethodKeyTypeNotImplemented, "VerificationMethod: %v", verificationMethod.Type)
 	}
-	pubKeySecp256k1, err := secp256k1util.PubKeyFromBase58(verificationMethod.PublicKeyBase58)
-	if err != nil {
-		return 0, sdkerrors.Wrapf(types.ErrInvalidSecp256k1PublicKey, "PublicKey: %v", verificationMethod.PublicKeyBase58)
-	}
-	newSeq, ok := types.Verify(sig, signData, seq, pubKeySecp256k1)
+
+	var key secp256k1.PubKey = make([]byte, secp256k1.PubKeySize)
+	copy(key[:], document.VerificationMethod[0].Value)
```
youngjoon-lee, line 102: It's interesting that `...Value` can be directly casted to the `secp256k1.PubKey` even though the type of `...Value` is `[]byte`. Could you point me where you found a related example or doc? Or, did you get to know about this by your own testings?

youngjoon-lee: Ah, it seems that this code worked because you wrote tests like below: https://github.com/medibloc/panacea-core/pull/626/files#diff-e47c9f5f5ffaa1e4cbc9dacaa82f13d0abbc999d329f8c72c1657885cd61cd97R124, which uses a `[]byte` representation of `secp256k1.PubKey` when creating a `VerificationMethod`.
I'll also check whether aries-framework-go provides more canonical way or not.

x/did/keeper/private_test.go
```diff
@@ -30,27 +32,60 @@ func TestVerifyDIDOwnership_SigVerificationFailed(t *testing.T) {
...
+		VerificationMethod: verificationMethod,
+		Authentication:     authentication,
+		Created:            &createdTime,
 	}
-	doc := types.NewDIDDocument(did, types.WithVerificationMethods(verificationMethods), types.WithAuthentications(authentications))
+	docBz, _ := doc.JSONBytes()
+
+	document := &types.DIDDocument{
+		Document:         docBz,
+		DocumentDataType: "aries-framework-go@v0.1.8",
```
youngjoon-lee, line 84: We can define this string as a const somewhere.
Also, it'd be better to use `github.com/hyperledger/aries-framework-go/pkg/doc/did.Doc@v0.1.8`. This is something like:
```go
fmt.Sprintf(
  "%s.%s@%s",
  reflect.TypeOf(doc).PkgPath(),
  reflect.TypeOf(doc).Name(),
  version,
)
```
I'm still not sure where to put the version string, but let's put it at last.

x/did/keeper/msg_server_did.go
gyuguen, line 102: I think you can change simply like below :)
Or is there a reason you are copying a byte array?
```suggestion
	key := secp256k1.PubKey(document.VerificationMethod[0].Value)
```

gyuguen (commented):
Thank you!
I left several comments.

x/did/types/did.go
```diff
@@ -121,221 +28,36 @@ func NewDID(pubKey []byte) string {
 	return fmt.Sprintf("did:%s:%s", DIDMethod, idStr)
 }
 
-func ParseDID(str string) (string, error) {
-	did := str
-	if !ValidateDID(did) {
-		return "", sdkerrors.Wrapf(ErrInvalidDID, "did: %v", str)
+func ValidateDID(did string) (string, error) {
```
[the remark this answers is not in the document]
audtlr24: Thanks, I fixed it to only return error

x/did/types/did.go
```diff
@@ -121,221 +28,36 @@ func NewDID(pubKey []byte) string {
...
+	_, err := ariesdid.Parse(did)
+	if err != nil {
+		return "", sdkerrors.Wrapf(ErrInvalidDID, "did: %v", did)
 	}
 	return did, nil
```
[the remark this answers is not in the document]
audtlr24: Thanks! I change error message to include error and related did.

x/did/types/did.go
```diff
@@ -121,221 +28,36 @@ func NewDID(pubKey []byte) string {
...
-// VerificationMethodByID finds a VerificationMethod by ID.
-// If the corresponding VerificationMethod doesn't exist, it returns a false.
-func (doc DIDDocument) VerificationMethodByID(id string) (VerificationMethod, bool) {
-	for _, verificationMethod := range doc.VerificationMethods {
-		if verificationMethod.Id == id {
-			return *verificationMethod, true
-		}
-	}
-	return VerificationMethod{}, false
+	return doc.DocumentDataType == ""
```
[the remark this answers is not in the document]
audtlr24: You are right. There may be changes on the `deactivate` part, but changed it to return true if either field is not present.

x/did/keeper/msg_server_did.go
audtlr24: With the `document proof` in mind, it seems that the `verification method` value should be changed to `btcec public key`. Please review again after this part is applied.

audtlr24: As I said https://github.com/medibloc/panacea-core/pull/626#discussion_r1098193126, this part will be change. Please review this part again later.

H4NLee (commented):
good. Are you working in progress for proof in did document?

x/did/genesis_test.go
```diff
@@ -61,17 +62,16 @@ func (suite *genesisTestSuite) newDIDDocumentWithSeq(did string) (types.DIDDocum
...
-		&es256VerificationMethod,
-		&blsVerificationMethod,
-	}
-	verificationRelationship := types.NewVerificationRelationship(verificationMethods[0].Id)
-	authentications := []types.VerificationRelationship{
-		verificationRelationship,
-	}
-	doc := types.NewDIDDocument(did, types.WithVerificationMethods(verificationMethods), types.WithAuthentications(authentications))
+
+	authentication := types.NewVerification(es256VerificationMethod, ariesdid.Authentication)
```
inchori, line 66: Does we need to change the `types/genesis.go` generated by `v3/genesis.proto`?
It means that change a `DefaultGenesis` and `Validate`.

inchori (commented):
Thanks for your effort.
I left a comment for genesis function.

x/did/keeper/msg_server_did.go
gyuguen: OK.
If you change this, please ping to me

proto/panacea/did/v3/did.proto
```diff
@@ -0,0 +1,26 @@
...
+option java_multiple_files = true;
+option go_package = "github.com/medibloc/panacea-core/x/did/types";
+
+import "gogoproto/gogo.proto";
+
+// DIDDocument defines a W3C DID Document with document data type to specify the version of document spec.
+message DIDDocument {
+  bytes document = 1;
+  string document_data_type = 2;
+}
```
youngjoon-lee, line 13:
```suggestion
message DIDDocument {
  bytes data = 1;
  string type = 2;
}
```
Let's go with simple names, as https://github.com/libp2p/go-libp2p/blob/5741b6c9bbcc1185bdf94d816dca966b37ce61ff/core/crypto/pb/crypto.proto#L14

x/did/keeper/private_test.go
youngjoon-lee: @audtlr24 
Let's go with `github.com/hyperledger/aries-framework-go@v0.1.8/pkg/doc/did.Doc`. This is the official notation in Go.

proto/panacea/did/v3/did.proto
youngjoon-lee:
```suggestion
message DIDDocument {
  bytes document = 1;
  string document_type = 2;
}
```
Sorry. Let's go like this. This would be clearer since you're adding a `deactivated` field as https://github.com/medibloc/panacea-core/pull/633/files#diff-e50f8f7e53eb90e2acaa0285df776b2eb2b13f0dd5a28ab17a0c021224f509fcR13

audtlr24:
I'm closing this PR in its current state. Since that aries is following a previous w3c spec and aries is not based on protobuf, it seems like a good idea to stick with the existing v2 x/did and not apply it right away.

I'll reopen it later if we need the various features of aries in a VC/VP-based datadeal model.

x/did/keeper/msg_server_did.go
audtlr24: In #633, the keytype was changed to a `btcec` type.

audtlr24: In https://github.com/medibloc/panacea-core/pull/633, the keytype was changed to a btcec type.

x/did/genesis_test.go
audtlr24: This is already reflected. The newly defined `DIDDocument` is stored in the `GenesisState` and is implemented to validate it.

proto/panacea/did/v3/did.proto
audtlr24: I will reflect this when we reopen this PR in the future.

PR #632, opened 2023-02-10
feat: add presentation definition to deal

### LINKED ISSUE #631: Add `presentation_definition` to data deal
With [`presentation_definition`](https://identity.foundation/presentation-exchange/#presentation-definition), consumers can specify data that meets specific types and/or conditions. The `presentation_definition` will be added to `deal` as below:

```proto
message Deal {
  // ...
  bytes presentation_definition = 10;
}
```

H4NLee:
close #631


youngjoon-lee (commented):
Basically, everything looks good. But, as we discussed in Slack, we need to consider using either data_schema or presentation_definition, instead of mixing them.

gyuguen (approved):
LGTM

Additionally, as discussed offline, I think we should add validation to only use either data_schema or pd.

audtlr24 (approved):
LGTM. As we discussed in other channels, I think the logic to handle `data_schema` and `pd` will be in a subsequent PR.

x/datadeal/types/message_deal.go
```diff
@@ -16,8 +16,13 @@ func (m *MsgCreateDeal) Type() string {
...
 func (m *MsgCreateDeal) ValidateBasic() error {
-	if len(m.DataSchema) == 0 {
-		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "no data schema")
+	if m.HasDataSchema() == m.HasPresentationDefinition() {
+		return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "one of data schema and presentation definition can be provided")
+	}
+	if m.HasPresentationDefinition() {
+		if err := ValidatePD(m.PresentationDefinition); err != nil {
+			return sdkerrors.Wrapf(sdkerrors.ErrInvalidRequest, "invalid presentation definition")
+		}
```
H4NLee, line 25: 이는 데이터 스키마와 PD 중 하나가 제공되었는지 확인하기 위한 것입니다.

proto/panacea/datadeal/v2/deal.proto
```diff
@@ -17,6 +17,7 @@ message Deal {
   string consumer_address = 7;
   repeated AgreementTerm agreement_terms = 8;
   DealStatus status = 9;
+  bytes presentation_definition = 10;
```
youngjoon-lee, line 20: Let's move this one to right below the `data_schema = 3` for readability. For now, we don't need to care that much of backward compatibility since we're not on mainnet yet.

youngjoon-lee (approved):
lgtm

proto/panacea/datadeal/v2/deal.proto
H4NLee: I see. :+1:

x/datadeal/types/message_deal_test.go
```diff
@@ -46,7 +46,7 @@ func TestMsgCreateDealValidateBasicEmptyValue(t *testing.T) {
 
 	err := msg.ValidateBasic()
 	require.ErrorIs(t, err, sdkerrors.ErrInvalidRequest)
```
gyuguen, line 48: It would be nice to add a test where both data_schema and pd exist.

gyuguen (approved):
Thank you.
I left one comment :)

x/datadeal/types/message_deal_test.go
H4NLee: Ok. I'll cover it as well.

H4NLee: done in [6554699](https://github.com/medibloc/panacea-core/pull/632/commits/65546994e9cbe071c707dcf94fb98a332bd36ae8)

PR #633, opened 2023-02-14
feat: did module spec change through application of did document proof

audtlr24:
## Background

In `aries-framework-go`, [`DocumentSigner.Sign`](https://github.com/hyperledger/aries-framework-go/blob/c509c1eb30d843e889a4b183bcd493ef0b22109d/pkg/doc/signature/signer/signer.go#L67) and [`VerifyProof`](https://github.com/hyperledger/aries-framework-go/blob/c509c1eb30d843e889a4b183bcd493ef0b22109d/pkg/doc/did/doc.go#L1211) methods are provided to verify the signature of the did document.
- `DocumentSigner.Sign` add `Proof` at did document using private key corresponding to public key of verificationMethod in did document.
- Information such as which verificationMethod to use and which encryption method to use can be defined through [`Signer.Context`](https://github.com/hyperledger/aries-framework-go/blob/c509c1eb30d843e889a4b183bcd493ef0b22109d/pkg/doc/signature/signer/signer.go#L48).
- `VerifyProof` verifies that the did document's proof field was created using the specified verificationMethod in the document.

Also, in the `Proof` of the document, there are `Domain` and `Nonce` that can replace the existing sequence to prevent replay-attack.
- We can specify this value in `Signer.Context` before document sign, and the proof value will be generated with different value depending on these.

## Implementation
Remove the `signature` from the existing tx and replace it with the `proof` inside the document
- Since DID Ownership cannot be properly verified using only `VerifyProof`, additional verification is implemented in `VerifyDIDOwnership`.
- Since `aries-framework-go` uses `ecdsa` key, implement conversion parts with `secp256k1` key used by panacea.

Remove the `sequence` part and change to use the `Domain` inside the document proof
- `Domain` is used because it is judged that the string `Domain` is more readable than the byte array `Nonce`.

Add a `Deactivated` field at `DIDDocument` and simply replace the related tx.
- In `Deactivation`, since there is no document update, it is simply changed to verify only the account signature.


proto/panacea/did/v3/tx.proto
```diff
@@ -47,9 +44,7 @@ message MsgUpdateDIDResponse {}
 // MsgDeactivateDID defines the Msg/DeactivateDID request type.
 message MsgDeactivateDID {
   string did = 1;
-  string verification_method_id = 2;
-  bytes signature = 3;
-  string from_address = 4;
+  string from_address = 2;
```
youngjoon-lee, line 47: > - Add a Deactivated field at DIDDocument and simply replace the related tx.
>
>   - In Deactivation, since there is no document update, it is simply changed to verify only the account signature.

Here, we still have to have a DID authentication (= ownership verification). Without it, anyone can deactivate a certain DID that is not actually owned by him.

You can probably define `MsgDeactivateDID` exactly same as `MsgUpdateDID`, so that you can utilize a `proof` in the `MsgDeactivateDID.document`. If so, a user who tries to deactivate a DID must get his DID document from Panacea first, put a proof into the document, and put the document into the `MsgDeactivateDID`.

audtlr24: There is something I misunderstood. As you said, anyone can deactivate it. What do you think is the way to store the panacea account address in `DIDDocument` and compare it when `DeactivateDID`? I understand the method you suggested, but I'm wondering which method is better.

youngjoon-lee: We basically try to not couple a DID with a certain chain account, since someone can lost his chain account key while not losing his DID key. That’s why we don’t store a mapping between DID and chain account in Panacea :)

audtlr24: Thanks, I understand. I will implement it in the way you suggested.

youngjoon-lee: Also, DID has a special concept so called ‘controller’. For example, you can be one of controllers of your kid’s DID, if your DID is written as a controller in your kid’s DID document. Then, you can update/deactivate your kid’s DID document on behalf of your kid. You don’t target this feature yet, but our basic mechanism should be flexible to cover these cases in the future.

x/did/types/document_sign.go
```diff
@@ -0,0 +1,152 @@
...
+		DocumentURL: url,
+		Document:    doc,
+	}, nil
+}
+
+func NewSignerContext(vmID string, signatureType string, sequence uint64) *signer.Context {
+	return &signer.Context{
+		SignatureType: signatureType,
+		Creator:       vmID,
+		Domain:        strconv.FormatUint(sequence, 10),
```
youngjoon-lee, line 110:
```suggestion
		Challenge:        strconv.FormatUint(sequence, 10),
```
It'd be clearer to use `Challenge` instead of `Domain` or `Nonce`: https://www.w3.org/TR/vc-data-model/#example-a-simple-example-of-a-verifiable-presentation.
The `Domain` is for the verifier (receiver) to specify the scope in which the `Challenge` value can have meaning. For example, the verifier can enforce the presenter to use a `Domain: did:panacea:abc...` and `Challenge: 10 (= sequence)`.
But now, I don't think we need to use `Domain`. It'd be enough to use `Challenge` for now. 

The `Nonce` is slightly different. In general, the `Challenge` is set by verifier (not presenter), as standardized as [Challenge-Response authentication](https://en.wikipedia.org/wiki/Challenge%E2%80%93response_authentication). In many cases, the presenter must set a response of the challenge to the `Challenge` field in the proof. But, in our case, the challenge response is the same as the challenge itself (it means that the presenter doesn't have to do any computation on the challenge).
On the other hand, the `Nonce` is usually chosen by presenter, for example, random values or time-based values. But, in practice, the uses of `Challenge` and `Nonce` are not clearly distinguished :)

It seems that it's preferred to use `Challenge` over `Nonce` these days: https://github.com/w3c/vc-data-integrity/issues/34, even though this link is not about the DID document proof.

proto/panacea/did/v3/query.proto
```diff
@@ -11,19 +11,19 @@ option go_package = "github.com/medibloc/panacea-core/x/did/types";
...
 message QueryDIDRequest {
 	// NOTE: Using base64 due to the URI path cannot contain colons.
 	string did_base64 = 1;
 }
 
-// QueryDIDResponse is the response type for the Query/DIDDocumentWithSeq RPC method.
+// QueryDIDResponse is the response type for the Query/DIDDocument RPC method.
 message QueryDIDResponse {
-	DIDDocumentWithSeq did_document_with_seq = 1;
+	DIDDocument did_document= 1;
```
youngjoon-lee, line 28:
```suggestion
	DIDDocument did_document = 1;
```

x/did/types/document_sign.go
youngjoon-lee, line 110: It'd be clearer to use `Nonce` over `Domain`. You can convert uint64 into []byte: https://go.dev/play/p/BN_4jwLXilk. 

The `Domain` is usually for the verifier to specify the scope in which the `Nonce` has a meaning.

x/did/keeper/did.go
```diff
@@ -6,24 +6,24 @@ import (
...
 }
 
-func (k Keeper) GetDIDDocument(ctx sdk.Context, did string) types.DIDDocumentWithSeq {
+func (k Keeper) GetDIDDocument(ctx sdk.Context, did string) *types.DIDDocument {
 	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.DIDKeyPrefix)
 	key := []byte(did)
 	bz := store.Get(key)
 	if bz == nil {
-		return types.DIDDocumentWithSeq{}
+		return &types.DIDDocument{}
```
youngjoon-lee, line 21: You can return `nil` since you decided to return a pointer.

x/did/types/document_sign.go
audtlr24: As we've discussed in other channels, `Nonce` don't have any effect on the generation of proof values, so they can't be used as a defense against replay-attacks. I will look at the role of `Domain` and `Nonce` in detail, but for now, let's leave it at that.

audtlr24:
I'm closing this PR in its current state. Since that `aries` is following a previous w3c spec and `aries` is not based on `protobuf`, it seems like a good idea to stick with the existing v2 `x/did` and not apply it right away. 

I'll reopen it later if we need the various features of `aries` in a VC/VP-based `datadeal` model.

PR #635, opened 2023-02-16
feat: v2.0.x: upgrade ibc-go from v2 to v4.3.0 (with cosmos-sdk v0.45.12 and informalsystems/tendermint v0.34.24)

youngjoon-lee:
Cherry-picked from 8293dbba08975cdf99ed869c42606f52a8c50332 and 1aec6cb471d41d758e85b124cfb1ddd47da95869, but [added an upgrade handler](https://github.com/medibloc/panacea-core/pull/635/commits/017b91d693e583ec262fd35f02728a513afcd2c0) to bump the consensus version of the ibc-transfer module to 2 as described in the [ibc-go v3->v4 migration guide](https://ibc.cosmos.network/main/migrations/v3-to-v4.html#migration-to-fix-support-for-base-denoms-with-slashes).
(This change wasn't made explicitly in the `main` branch because it already has an [upgrade handler](https://github.com/medibloc/panacea-core/blob/8293dbba08975cdf99ed869c42606f52a8c50332/app/app.go#L920) for the new oracle and datadeal modules. ~~It may be great to add the same comment as [this](https://github.com/medibloc/panacea-core/pull/635/commits/017b91d693e583ec262fd35f02728a513afcd2c0) to the upgrade handler in the `main`~~ <- DONE https://github.com/medibloc/panacea-core/pull/636)

This change will probably be released as v2.0.6 before releasing DEP officially (that will be potentially released as v3.x.x).


H4NLee (approved):
thanks

gyuguen (approved):
Thank you!!

PR #640, opened 2023-02-20
feat: remove cosmwasm

audtlr24:
## Background
We decided it would be better to remove cosmwasm from the chain. There are three reasons for this
1. Hardship in removing CosmWasm once somebody starts using it; currently, nobody is using CosmWasm on Panacea yet
2. Possibility that anybody can deploy a contract that does not fit the purpose of Panacea
3. Replacable through Cosmos SDK v0.46.x that now includes x/nft module later


audtlr24:
I haven't tested the upgrade handler yet. I will open the pr after testing is complete.

audtlr24:
Upgrading directly from v2.0.5 had issues with the VersionMap part, so I tested the upgrade in branch `v2.0.5/add-set-version-map`.

gyuguen:
I think you can remove to wasm config in Dockerfile

https://github.com/medibloc/panacea-core/blob/main/Dockerfile

* Line 8~9
* Line 20~21
* Line 28~29

gyuguen:
And you can remove this part.
https://github.com/medibloc/panacea-core/blob/ft/na/remove-cosmwasm/.gitbook/overview/tech-stack.md?plain=1#L45
https://github.com/medibloc/panacea-core/blob/ft/na/remove-cosmwasm/README.md?plain=1#L15

gyuguen (approved):
LGTM

I tested this locally and didn't have any issues.
And I left a few comment :)

.gitbook/overview/tech-stack.md
```diff
@@ -41,15 +41,6 @@ Also, data holders can reveal only data that they really want to share using the
...
 
-
-## Smart Contract
-
-Panacea provides the Smart Contract framework powered by CosmWasm.
-Any developers can implement their own smart contracts using the Rust programming language and deploy them on Panacea.
-It means that any fancy ideas such as DeFi and NFT can be implemented on Panacea,
-and they can be used for boosting the liquidity of health data flow of medical systems built on Panacea.
-
-
```
youngjoon-lee, line 52: I updated a diagram as well: 15ad4233e4f30f15a697dbb24de1d731f76614d2

app/app.go
```diff
@@ -848,6 +763,9 @@ func (app *App) registerUpgradeHandlers() error {
 	})
 
 	app.UpgradeKeeper.SetUpgradeHandler("v2.0.6", func(ctx sdk.Context, plan upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
+
+		app.IBCKeeper.ConnectionKeeper.SetParams(ctx, ibcconnectiontypes.DefaultParams())
+
```
youngjoon-lee, line 768: Do we need to have this for the v2.0.6 upgrade handler?

audtlr24: Ah, we don't need it. I fix it, thanks!

.gitbook/overview/tech-stack.md
audtlr24: Thanks!

app/app.go
youngjoon-lee: thanks!

youngjoon-lee (approved):
lgtm,man

inchori (approved):
LGTM!

H4NLee (approved):
LGTM

PR #98, opened 2021-04-19
x/did: Support service / controller / various verificationRelationship in DID Document

youngjoon-lee:
Closes #48 

- Now, DID Document can contain (according to [the semi-finalized new DID spec](https://www.w3.org/TR/did-core/))
  - `controller`
  - `service`
  - `assertionMethod`: essential for VC/VP verification using the [vc-service](https://github.com/medibloc/vc-service)
  - `keyAgreement`
  - `capabilityInvocation`
  - `capabilityDelegation`

PR #646, opened 2023-03-03
feat: change to a consumer storage service model instead of ipfs

audtlr24:
closes #98.
- change `cid` to `data_endpoint` in `unsigned_certificate`
- add `consumer_service_endpoint` to `deal`


proto/panacea/datadeal/v2/consent.proto
```diff
@@ -19,7 +19,7 @@ message Certificate {
 
 // UnsignedCertificate defines a certificate issued by an oracle as a result of data validation.
 message UnsignedCertificate {
-  string cid = 1;
+  string data_endpoint = 1;
```
audtlr24, line 22: I'm confused as to whether this field is necessary. If we enforce a specification that all `consumer_service` store with the same key type(e.g. dataHash), `consumer` or `provider` can infer `data_endpoint` from the `deal`.

H4NLee: IMO, we don't need `data_endpoint`. It can be obtained from `deal_id` and `data_hash` in consumer storage.

gyuguen: I agree. We don't need `data_endpoint`.

gyuguen (commented):
I don't think the consumer's endpoint is needed for certification.
As @H4NLee  said, that information should only come from the deal.

proto/panacea/datadeal/v2/consent.proto
audtlr24: Thanks for your feedback. I remove it in [26f9f11](https://github.com/medibloc/panacea-core/pull/646/commits/26f9f1119f024c6fa65c73020a335ea2fd911bd6).

gyuguen (approved):
LGTM

youngjoon-lee (approved):
lgtm

H4NLee (approved):

PR #654, opened 2023-03-20
feat: add a `panacead dep hash-data` command

### LINKED ISSUE #652: Add a `dep` command and `datahash` command
## Background
The `encrypt-data` and `decrypt-data` command are only used in DEP(Data Exchange Protocol). But for now, the commands are located in root `panacead` command. Also, the users are now need to get a `datahash` of provided data by using `sha256sum` manually. It could cause a providing same data which is not following canonical JSON format.

## Implementation
Add a `dep` command and placed the `encrypt-data` and `decrypt-data` command under `dep` command. Add a `panacead dep hash-data ...` command for provider to get a unique datahash of provided data. 
As a result, the commands will be:
- `panacead dep encrypt-data ...`
- `panacead dep decrypt-data ...`
- `panacead dep hash-data ...`

inchori:
Closes #652

### DIFF
--- a/cmd/panacead/cmd/dep/hash_data.go
+++ b/cmd/panacead/cmd/dep/hash_data.go
@@ -14,8 +14,9 @@ func HashDataCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "hash-data [data-file-path]",
 		Short: "Hash a data by sha256 algorithm",
-		Long:  ``,
-		Args:  cobra.ExactArgs(1),
+		Long: `
+This command can hash a data by sha256 algorithm. The data will be encoded to canonical JSON format and printed as a hash value.`,
+		Args: cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			origData, err := os.ReadFile(args[0])
 			if err != nil {
--- a/cmd/panacead/cmd/dep/hash_data.go
+++ b/cmd/panacead/cmd/dep/hash_data.go
@@ -1,0 +1,40 @@
+package dep
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"os"
+
+	"github.com/cyberphone/json-canonicalization/go/src/webpki.org/jsoncanonicalizer"
+	"github.com/spf13/cobra"
+)
+
+func HashDataCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "hash-data [data-file-path]",
+		Short: "Hash a data by sha256 algorithm",
+		Long:  ``,
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			origData, err := os.ReadFile(args[0])
+			if err != nil {
+				return err
+			}
+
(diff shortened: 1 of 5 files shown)


cmd/panacead/cmd/dep/hash_data.go
```diff
@@ -0,0 +1,41 @@
...
+			if err != nil {
+				return err
+			}
+
+			jsonDataHashBz := sha256.Sum256(jsonData)
+			_, err = fmt.Fprintln(cmd.OutOrStdout(), hex.EncodeToString(jsonDataHashBz[:]))
+			if err != nil {
+				return err
+			}
+			return nil
```
gyuguen, line 36: Is there a reason you wrote this in `io.Writer` instead of just doing `fmt.Println(hex.EncodeToString(jsonDataHashBz[:]))`?

gyuguen (approved):
LGTM

I left a comment.

cmd/panacead/cmd/dep/hash_data.go
inchori: Because I want to use `StdOut` as other commands(`encrypt-data`, `decrypt-data`) do. So I opened a #561 last time, too.

cmd/panacead/cmd/dep/hash_data.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+func HashDataCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "hash-data [data-file-path]",
+		Short: "Hash a data by sha256 algorithm",
+		Long: `
+This command can hash a data by sha256 algorithm. The data will be encoded to canonical JSON format and printed as a hash value.`,
+		Args: cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
```
youngjoon-lee, line 20:
```suggestion
func HashDataCmd() *cobra.Command {
	cmd := &cobra.Command{
		Use:   "hash-data [json-file-path]",
		Short: "Hash a JSON data by sha256 algorithm",
		Long: `
This command can hash a JSON data by sha256 algorithm. The data will be encoded to canonical JSON format and printed as a hash value.`,
		Args: cobra.ExactArgs(1),
		RunE: func(cmd *cobra.Command, args []string) error {
```
It would be better to mention that this command targets only JSON data since it does JSON canonical encoding implicitly.

cmd/panacead/cmd/dep/hash_data.go
youngjoon-lee: In fact, the `fmt.Println()` writes to stdout. So, what you implemented is the same as `fmt.Println()` as @gyuguen said, unless you specify any output file parameter when you run the command (implemented by cobra).

For now, it's okay to use either of them.

cmd/panacead/cmd/dep/hash_data.go
inchori: Fixed it in [4f76219](https://github.com/medibloc/panacea-core/pull/654/commits/4f76219ab2f7c25d26b4c0be50ea73a77be22974).

youngjoon-lee (approved):
lgtm

PR #657, opened 2023-03-31
docs: Update Docker image version from 'latest' to 'main'

kyungilcho:
Update the Docker image version in the documentation to reflect the actual available version. The 'latest' tag was not available, causing confusion for users. The 'main' tag should be used instead.


gyuguen (approved):
LGTM
The current version is correct to use.

audtlr24 (approved):
lgtm

youngjoon-lee:
Let's keep using `latest`. The `main` tag shouldn't be used by external users because it's only for internal CIs. Instead, we need to create a git tag from our latest stable commit and publish a docker image with `latest` tag.

gyuguen:
I think our documentation should use the latest docker image version. I'll close this PR.

Issue #704, opened 2024-07-07
Nodes/Relayers are down


arcturiano:
It is currently not possible to view balance in Keplr wallet, nor is it possible to deposit or withdraw to and from Osmosis nor TFM.

gyuguen:
Our public nodes and relayers are all up and running. The IBC between OSMOSIS and MEDIBLOC is working fine, as shown below.

https://www.mintscan.io/medibloc/relayers/channel-1/osmosis/channel-82?sector=status
https://www.mintscan.io/medibloc/relayers/channel-1/osmosis/channel-82?sector=operators
https://www.mintscan.io/medibloc/tx/AA20CD290F69FC35B9FA6586693DD0ADDD5C4E0085A829369990BEBC482CA97E?height=16848601

arcturiano:
Thank you, yet for some reason we still cannot deposit into Osmosis.  When clicking on deposit, balance does not load and in TFM it says "Erros -Nodes are down".  

Trying to send  IBC to Osmosis manually with Keplr, but transactions fail.

We changed RPCs in Keplr and balance is now loading there, but still unable to send to Osmosis for some reason.  

Any ideas how this transaction was sent? https://www.mintscan.io/medibloc/tx/AA20CD290F69FC35B9FA6586693DD0ADDD5C4E0085A829369990BEBC482CA97E?height=16848601

gyuguen:
I haven't tried it in Keplr, but I did it on the site below and the tokens moved without a problem: https://app.osmosis.zone/portfolio

arcturiano:
May I know what wallet you are using?

For Keplr is not possile to deposito to Osmosis via https://app.osmosis.zone/portfolio guessing it is a Keplr problem, as it is also not connecting to TFM nor can an IBC transaction be sent. Tried as a normal IBC and it did notwork,neither via a manual IBC transaction. This used to work about three weeks ago.  

Now Osmosis does not get balance from Keplr and no way to deposit.

drjavedp:
Same issue for me as well. Kepler does not show the balance. Mintscan shows the wallet balance and transactions correctly.

Also wallet.gopanacea.org displays a message "unable to connect to node".  Balance is displayed correctly. Even after setting a very high fee send transactions fail with a message "something wrong in the transaction". Receive transactions working fine on wallet.gopanacea.org as well.

Please correct these errors at the earliest.

drjavedp:
[...]

PR #711, opened 2026-07-02
feat: add ICS-29 fee middleware and packet-forward-middleware to the IBC transfer stack

renezander030:
## Why
Panacea's most visible user-facing pain is IBC transfer reliability — holders have reported deposits to Osmosis stalling and "nodes are down" errors (#704). A large part of that class of problem is that **relayers have no on-chain incentive** to service Panacea's channels, and the chain has **no multi-hop routing**. Both are now standard hardening on active Cosmos app-chains, and neither was wired in.

This PR adds the two pieces of middleware that address it, on the current ibc-go v7 line (no major SDK upgrade required).

## What
- **ICS-29 fee middleware** (`29-fee`) — lets users escrow relayer fees on the transfer channel, giving relayers a reason to deliver Panacea packets.
- **Packet-forward-middleware** (PFM v7.2.0) — enables multi-hop IBC routing through Panacea.

Transfer stack is wired base-outward as **transfer → packet-forward → fee**, with the transfer keeper's `ICS4Wrapper` set to the fee keeper so outbound sends flow transfer → fee → channel.

### Changes
- `app/keepers/keepers.go` — new `IBCFeeKeeper` + `PacketForwardKeeper`; fee keeper created before transfer; transfer `ICS4Wrapper` = fee keeper; wrapped transfer stack built and set on the IBC router.
- `app/keepers/keys.go` — `feeibc` + `packetfowardmiddleware` store keys.
- `app/app.go` — module basics, module-manager entries, `maccPerms` (both `nil`), begin/end/init-genesis ordering, PFM param subspace, upgrade registration.
- `app/upgrades/v2_3_0/` — upgrade handler adding the two new stores; `RunMigrations` initialises both modules with default genesis (fee % = 0, no in-flight packets).

### Dependency bumps (required by PFM v7.2.0; patch-level)
- `ibc-go/v7` v7.5.2 → **v7.7.0**
- `cosmos-sdk` v0.47.12 → **v0.47.13**
- adds `ibc-apps/middleware/packet-forward-middleware/v7` v7.2.0

## Upgrade / migration
Introduces on-chain upgrade **`v2.3.0`**. Adding the two modules is state-machine-breaking, so it must ship via a coordinated governance upgrade at a set height — the `StoreUpgrades{Added: [feeibc, packetfowardmiddleware]}` handler is included. Default fee percentage is 0, so existing transfer behaviour is unchanged until fees are opted into.

## Verification
- `go build ./...` — passes
- `go vet ./app/...` — clean
- `panacead` binary builds

Not yet exercised end-to-end: the repo has no interchaintest/e2e suite, so a live two-chain relayer run of a fee-incentivized transfer and a 3-chain forward is a recommended follow-up before a mainnet upgrade.

## Scope note
Deliberately **not** bundled with an SDK v0.50 / ibc-go v8 major upgrade or x/did work — each is independently consensus-breaking and belongs in its own release.


cl9200:
Closing this because it does not address a current Panacea requirement.

Osmosis does not support ICS-29, so the fee middleware cannot be used on the Panacea–Osmosis channel.
PFM is also unnecessary when Osmosis is the intermediate hop, and we have no use case requiring
Panacea itself to act as an intermediate chain.

We don’t have a concrete use case for either middleware at this time.

renezander030:
Thanks for the clear rationale, that makes sense. Without ICS-29 support on the Osmosis side there is indeed no path for the fee middleware on that channel today.

